PR #2, opened 2015-12-05
Markdown cleanups

teran-mckinney:
Still not completely consistent, but should be a start. Much easier to read with the monospace formatting around the commands.


kauffj:
Ty @teran-mckinney!

PR #31, opened 2016-05-07
Configure linux build on travis-ci.org

jobevers:
Settled on using travis-ci over circleci because travis will allow us to build both linux and osx, which is a feature circleci doesn't support.

This pull request sets up the changes needed to get the build to work for linux.  Somebody with access to the lbry org/repo will need to add the repo to travis-ci (or give me access so that I can do it).

Note that no tests are run; this is really the minimal amount that could be done to get this working.
Also note that the built .deb file isn't copied anywhere.

### DIFF
--- a/packaging/ubuntu/ubuntu_package_setup.sh
+++ b/packaging/ubuntu/ubuntu_package_setup.sh
@@ -8,7 +8,7 @@
 set -euo pipefail
 
 BRANCH=${1:-master}
-WEB_UI_BRANCH=${2:-}
+WEB_UI_BRANCH=${2:-master}
 
 BUILD_DIR="lbry-build-$(date +%Y%m%d-%H%M%S)"
 mkdir "$BUILD_DIR"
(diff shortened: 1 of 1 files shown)


kauffj:
@lyoshenka should be the one to merge this. Also, what does @jackrobison need to do to get the OSX build working in Travis?

lyoshenka:
@jobevers this looks good for the most part, but I'd like to retain the ability to build the project just by running the script. is there a good way to know if we're in circle or not (maybe it sets some env variables)? that way we can do circle-specific things and non-circle things as necessary.

packaging/ubuntu/ubuntu_package_setup.sh
```diff
@@ -6,33 +6,44 @@
...
-
-# check out LBRY
-git clone https://github.com/lbryio/lbry.git --branch "$BRANCH"
+$SUDO pip install make-deb
 
 # build packages
+#
+# dpkg-buildpackage outputs its results into '..' so
+# we need to move lbry into the build directory 
+mv $SOURCE_DIR lbry
```
lyoshenka: it looks like this line is moving the current directory into lbry, not lbry into the current directory. is that what you intended? it seems to be the opposite of what the comment says

packaging/ubuntu/ubuntu_package_setup.sh
```diff
@@ -6,33 +6,44 @@
...
+SUDO=''
+if (( $EUID != 0 )); then
+    SUDO='sudo'
+fi
 
 BRANCH=${1:-master}
 WEB_UI_BRANCH=${2:-}
 
-BUILD_DIR="lbry-build-$(date +%Y%m%d-%H%M%S)"
+BUILD_DIR="$HOME/lbry-build-$(date +%Y%m%d-%H%M%S)"
```
lyoshenka: does circleci require this? it would be nicer if everything could happen in the current dir

jobevers:
Finally got the OSX app building on travis too.  Will clean this up later in the week.  I'll split up OSX and linux because OSX has some extra complications with codesigning that would be better tested in a branch instead of a pull request. (For security reseasons travis-ci doesn't enable secure environment variables on pull requests)

The encrypted certificates and identify files that are in this pull request are mine (self-signed). The actual certs will need to be added to the repo - @jackrobison.

PR #47, opened 2016-06-23
Fix OSX build issue with openssl

jobevers:
Use python from python.org.

- py2app won't  included the system default in the app bundle
- brew's version of python links against brew's version of openssl - which doesn't get included in the bundle so it fails on a machine that doesn't have that version of openssl


jobevers:
Tested the app bundle on a clean vm (without brewed openssl) and it appears to work.

PR #69, opened 2016-07-14
Revert to 2 params as isinstance cannot support 3+

kimihiro64:


lbrynet/lbrynet_daemon/LBRYDaemonCLI.py
```diff
@@ -35,7 +35,7 @@ def main():
     if len(args) > 1:
         if isinstance(args[1], dict):
             params = args[1]
-        elif isinstance(args[1], str, unicode):
+        elif isinstance(args[1], str) or isinstance(args[1], unicode):
```
jobevers: `elif isinstance(args[1], basestring):`

jobevers: can also use a tuple for the classes: https://docs.python.org/2/library/functions.html#isinstance

kimihiro64:
There we go.

jobevers:
Travis has had difficulties on osx, merging despite the build failure there.

PR #72, opened 2016-07-15
Refactor jsonrpc get

jobevers:
Fixes two bugs:
 - `os.dirname` should be `os.path.dirname`
 - the `download_directory` parameter was being ignored in the output

Refactorings:
Move parameter handling into its own function and
better use of the `.get()` function for dictionaries.

Early return on the failed checks is more readable.

The lambda function in the callback was long and
hard to read so moved it out.


jobevers:
Closing as https://github.com/lbryio/lbry/pull/73 also includes these commits

PR #73, opened 2016-07-15
Refactor _download_name

jobevers:
Nested functions are the devil, especially ones that
use variables from the outer scope. Refactoring
_download_name to use a helper class helps make
the scoping more explicit and will undoubtably
prevent bugs in the future.

I think this makes _download_name drastically more readable.

Also cleaned up some duplicated code and
made download_directory respect the passed in parameter
instead of being the default.

Fixes two bugs:
 * os.dirname should be os.path.dirname
 * the download_directory parameter was being ignored in the output

Refactorings:
Move parameter handling into its own function and
better use of the .get() function for dictionaries.

Early return on the failed checks is more readable.

The lambda function in the callback was long and
hard to read so moved it out.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -298,7 +298,7 @@ ignored-classes=twisted.internet,RequestMessage
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E1101 when accessed. Python regular
 # expressions are accepted.
-generated-members=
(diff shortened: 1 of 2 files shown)


jobevers:
As this has been open for five days without comment - I'm going to assume that nobody has a problem with it.

PR #75, opened 2016-07-16
use asctime in logging format

jobevers:
Instead of manually adding datetime.now() use
asctime formatter to do it for us.


jobevers:
closing this in favor of https://github.com/lbryio/lbry/pull/94.  I tried to rebase and ran into conflicts and it was easier to make a new branch.

PR #76, opened 2016-07-16
Rename metadata field content-type to content_type, as the - characte…

kimihiro64:
…r makes the property inaccessible when parsed by reactJS.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -48,5 +48,5 @@ DEFAULT_CACHE_TIME = 3600
 DEFAULT_UI_BRANCH = "master"
 
 SOURCE_TYPES = ['lbry_sd_hash', 'url', 'btih']
-BASE_METADATA_FIELDS = ['title', 'description', 'author', 'language', 'license', 'content-type']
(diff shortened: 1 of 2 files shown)


jobevers:
First - being picky - please limit the length of the first line of the commit message.  This would read better as:
```Rename metadata field content-type to content_type.

The dash in content-type makes the property inaccessible when
parsed by reactJs.  Switching to underscore.
```

Second - changes to the meta-data parsing *have to be* backwards compatible. This change will fail here: https://github.com/lbryio/lbry/blob/ed0f9d4fb5b403290cf75833c5063fc2d628a4d6/lbrynet/core/LBRYWallet.py#L329 when the meta-data read from the blockchain is using `content-type` instead of `content_type`

On the JS side, can you handle this by using `obj['content-type']` instead of `obj.content-type`?

kimihiro64:
I figured since we only have a small handful of files at the moment, it would be easy to manually update them to use content_type. The ['content-type'] does work, although I'd prefer we change it at some point. At the very least, use _ instead of - for any new properties?

PR #78, opened 2016-07-17
emergency fix: force lbrycrd to be the default wallet

jobevers:
Ignore save settings for wallet type and use the default wallet (lbrycrd)
instead.  lbryum can be used but only if specified on the command line.


lyoshenka:
looks good to me. ive been refactoring that file a bit and agree with the comments you made

lyoshenka:
didnt test it though

RdeWilde:
But would that mean wallet(file)s switch too? That would become a small disaster.

jobevers:
@rdewilde - Its not great but one solution is to manually send coins from the lbryum wallet to the lbrycrd wallet.

PR #79, opened 2016-07-17
script to migrate private keys to lbrycrd

jobevers:
Exports the private keys out of the lbryum wallet, reformats them and imports them into lbrycrd.

Requires the lbrycrdd is already running and that lbrycrd-cli is on the path.


jobevers:
Tested this script - it worked for at least one person one slack.

PR #85, opened 2016-07-18
Update protobuf library

jobevers:
Updating to protobuf-v3.0.0b3 because it
fixes a bug here https://github.com/google/protobuf/blob/e841bac4fcf47f809e089a70d5f84ac37b3883df/python/google/protobuf/internal/python_message.py#L59
where six.moves is not always available.
In particular, the system python on OSX has an old version of six
which is without six.moves.


jobevers:
This ones trivial enough - merging it in.

PR #92, opened 2016-07-20
Add support for file uploads (for Publish page)

alexliebowitz:
This adds a new resource and shouldn't affect anything else

### DIFF
--- a/lbrynet/lbrynet_daemon/LBRYDaemonServer.py
+++ b/lbrynet/lbrynet_daemon/LBRYDaemonServer.py
@@ -4,6 +4,10 @@ import shutil
 import json
 import sys
 import mimetypes
+import mimetools
+import tempfile
+import time
+import cgi
 
 from datetime import datetime
 from appdirs import user_data_dir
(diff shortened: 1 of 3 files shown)


lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -29,6 +33,184 @@
 log.addHandler(handler)
 log.setLevel(logging.INFO)
 
+class LBRYDaemonRequest(server.Request):
```
jobevers: This class is large enough that it should live in its own module.

Huge modules are harder to read.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -29,6 +33,184 @@
...
+    #           second write completes.
+    memorylimit = 1024*1024*100
+
+    # enable/disable debug logging
+    do_log = True
+
+    # re-defined only for debug/logging purposes
+    def gotLength(self, length):
+        if self.do_log:
+            print '%f Headers received, Content-Length: %d' % (time.time(), length)
```
jobevers: print statements shouldn't be used for logging.  Use the built in logging functionality.  `do_log` can also be removed in favor of that.

this should be: `log.debug('Headers received, Content-Length: %s', length)`

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -29,6 +33,184 @@
...
+        server.Request.gotLength(self, length)
```
jobevers: prefer to use `super` if possible.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -29,6 +33,184 @@
...
+                        return
+                    raise
+
+            self.content.seek(0, 0)
+
+        self.process()
+
+    # re-definition of cgi.parse_multipart that uses a single temporary file to store
+    # data rather than storing 2 to 3 copies in various lists.
+    def parse_multipart(self, fp, pdict):
```
jobevers: I think this can be a function outside of the class.  I'd suggest pulling it out.

This is also a good candidate for some unit tests.  Add a new file in the `tests` directory that mirrors whatever new module this ends up (for example, maybe its `/tests/lbrynet/lbrynet_daemon/test_LBRYDaemonRequests.py`.  Inherit a new test case from `twisted.trial.unittest`.  There is an example in `/tests/lbrynet/core/server/test_BlobRequestHandler.py`

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -29,6 +33,184 @@
...
+        if self.method == b"POST" and ctype:
+            mfd = b'multipart/form-data'
+            key, pdict = cgi.parse_header(ctype)
+            if key == b'application/x-www-form-urlencoded':
+                args.update(parse_qs(self.content.read(), 1))
+            elif key == mfd:
+                try:
+                    self.content.seek(0,0)
+                    args.update(self.parse_multipart(self.content, pdict))
+                    #args.update(cgi.parse_multipart(self.content, pdict))
```
jobevers: This is crappy on twisted's part.  It would be nice to be able to set this instead of having to copy the entire function.  Might be worth creating an issue with twisted.

Would this work:
```
def requestReceived(self, command, path, version):
    old_parse_multipart = cgi.parse_multipart
    cgi.parse_multipart = parse_multipart
    super(LBRYDaemonRequest, self).requestRecieved(command, path, version)
    cgi.parse_multippart = old_parse_multipart
```

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -179,12 +361,35 @@ def render_GET(self, request):
     #     call.addErrback(lambda err: log.info("Error: " + str(err)))
     #     call.cancel()
```
jobevers: being picky, but leaving two lines between classes and other top-level modules helps make the code more readable.

https://google.github.io/styleguide/pyguide.html#Blank_Lines

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -179,12 +361,35 @@ def render_GET(self, request):
...
+    """
+
+    def __init__(self, api):
+        self._api = api
+
+    def render_POST(self, request):
+        origfilename = request.args['file_filename'][0]
+        uploaded_file = request.args['file'][0]  # Temp file created by request
+
+        # Move to a new temporary dir and restore the original file name
```
jobevers: comments should explain more why the code is doing something instead of what.

If you need a comment to say what the code is doing, then it needs to be made more readable.  In this case, I'd suggest a new function, maybe `move_to_temp_dir_and_restore`

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -179,12 +361,35 @@ def render_GET(self, request):
...
+        newdirpath = tempfile.mkdtemp()
+        newpath = os.path.join(newdirpath, origfilename)
+        shutil.move(uploaded_file.name, newpath)
+        self._api.uploaded_temp_files.append(newpath)
+
+        return json.dumps(newpath)
```
jobevers: I don't really know... but this might be more useful to clients if it was an object, maybe `json.dumps({'uploaded_filepath': newpath})

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -179,12 +361,35 @@ def render_GET(self, request):
...
+
 
 class LBRYDaemonServer(object):
     def _setup_server(self, wallet):
         self.root = LBRYindex(os.path.join(os.path.join(data_dir, "lbry-ui"), "active"))
         self._api = LBRYDaemon(self.root, wallet_type=wallet)
         self.root.putChild("view", HostedLBRYFile(self._api))
+        self.root.putChild("upload", LBRYFileUpload(self._api))
```
jobevers: I know `view` is already here, but having the api defined in multiple places makes it hard to look up.  Not worth changing now, but a point to consider for future refactorings.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: Good catch. It's in PEP 8, so I do follow this (the only thing I knowingly bend in PEP 8 in is line lengths)

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: Thought about this and decided to go with just a string. I've seen REST APIs that return just strings before, and it makes sense when you're returning a single value, not a data structure that could have more elements later.

Also, this isn't part of the main JSON-RPC API; it's a custom resource used by the web UI, like /view (this one happens to return some JSON, which I guess technically makes it a separate REST API). It's a one-off hack to deal with the fact that we can't collect the original file path from the Web UI.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: The thing is that it's not an endpoint in the main JSON-RPC API, it's a POST resource for a form in the web UI. We wouldn't want to mix that functionality in with the JSON-RPC code, ditto for "view."

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: This class is 99% copy and pasted from [that blog post](http://sammitch.ca/2013/07/handling-large-requests-in-twisted/
). Mostly I just deleted code branches we didn't need. I didn't want to spend a ton of time refactoring someone else's code; I'd rather have done a library import but 1) it's not on PyPI and 2) I had some changes that would require monkey patching if I didn't edit the code

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: I lean toward the "[packages as units of reuse](http://stackoverflow.com/a/107836)" model, and this is unlikely to get reused anywhere else. It includes a slightly ugly hack that only applies to this exact situation (i.e. not deleting the temp files). You're right, though, the file is just too big. I'll move it out.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: In this case, the "what" and the "why" are kind of the same--the file starts in a tempfile with a randomized name and we need to put it in a new spot with the correct name.

Also, `move_to_temp_dir_and_restore_file_name`  is the shortest function name I can think of that would convey the needed info on its own, and that's pretty long winded. I guess I'll just modify LBRYDaemonRequest so it directly puts them in the right place (makes the code even less reusable since we're creating temp dirs as well as temp files, but whatever).

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
alexliebowitz: Twisted knows about this and they're not changing anything because it will be moot once they fix a different bug (see that blog post).

Modifying built in modules and then changing them back is kinda gnarly... even more of a hack than copy and pasting code IMO. It also creates a potential race condition if we ever use cgi.parse_multipart in another thread (unlikely to come up in practice but it's not strictly guaranteed safe like in JS).

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
jobevers: An approach to get what you want using the logging module would be to make a logger for this class `log = logging.getLogger('lbrynet.lbrynet_daemon.LBRYDaemonServer.LBRYDaemonRequests')` Then  at runtime, a caller can turn off the log statements using something like `logging.getLogger('lbrynet.lbrynet_daemon.LBRYDaemonServer.LBRYDaemonRequests').setLevel('WARNING')`

PR #94, opened 2016-07-20
use asctime in logging format

jobevers:
Instead of manually adding datetime.now() use
asctime formatter to do it for us.


jackrobison:
👍

PR #97, opened 2016-07-21
lbryum fixes

jackrobison:
-fix _do_send_many
-fix situation where lbryum won’t connect when it starts out already up to date

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.3.8
+current_version = 0.3.9
 commit = True
 tag = True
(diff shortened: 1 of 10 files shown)


jobevers:
As someone who is perhaps a little too anal-retentive about a clean commit history  - this could have been just one, maybe two commits.

Also, There are three merge commits in here - I'd suggest looking into rebasing instead of merging.

kauffj:
@jobevers you and @lyoshenka are going to be best friends!

Shouldn't this PR be getting reviewed by one of you before merge anyway? Developers should only merge their own PRs if they are extraordinarily simple or extraordinarily urgent.

PR #98, opened 2016-07-21
Remove datetime.now() from log statements

jobevers:
Some of these snuck back in.


jobevers:
This is trivial enough, and the other related changes have already been working.  merging.

PR #101, opened 2016-07-25
Properly Compare Version Strings

jobevers:


lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -53,8 +54,6 @@
 from lbrynet.lbryfile.LBRYFileMetadataManager import DBLBRYFileMetadataManager, TempLBRYFileMetadataManager
 # from lbryum import LOG_PATH as lbryum_log
 
-log = logging.getLogger(__name__)
```
lyoshenka: @jobevers are you sure this doesnt belong? is `log` magically defined everywhere now?

jobevers: no magic - this line was redundant: https://github.com/lbryio/lbry/pull/101/files#diff-4510ef2fe0fe3e8b0ff6c011d8cc31abR68

jackrobison:
👍

PR #102, opened 2016-07-25
Fix: Log configuration should not happen when a module is imported

jobevers:
Instead, move the responsibility to the main program. Also,
each module had the same, redundant setup.


lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -56,6 +56,7 @@
 log = logging.getLogger(__name__)
 
 
+# TODO: this code snippet is everywhere. Make it go away
```
lyoshenka: :+1: i thought the same thing a few days ago

lyoshenka: :+1: i thought the same thing a few days ago

lyoshenka:
untested ACK

jackrobison:
👍

PR #104, opened 2016-07-26
Remove print line

jobevers:
This line is responsible for flooding the console with messages like
<_io.BytesIO object at 0x7f9e286e52f0>


jobevers:
Trivial.  Merging this.

PR #107, opened 2016-07-27
Metadata versioning, publishing things, and USD denominated key fees

jackrobison:


kauffj:
Prepare for tsk tsking from @jobevers and/or @lyoshenka for mixing features in one branch :D

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+import json
+import time
+
+from copy import deepcopy
+from googlefinance import getQuotes
+from lbrynet.conf import CURRENCIES
+import logging
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
jobevers: remove the setLevel line

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+
+SOURCE_TYPES = ['lbry_sd_hash', 'url', 'btih']
+BASE_METADATA_FIELDS = ['title', 'description', 'author', 'language', 'license', 'content-type', 'sources']
+OPTIONAL_METADATA_FIELDS = ['thumbnail', 'preview', 'fee', 'contact', 'pubkey']
+
+#v0.0.1 metadata
+METADATA_REVISIONS = {'0.0.1': {'required': BASE_METADATA_FIELDS, 'optional': OPTIONAL_METADATA_FIELDS}}
+#v0.0.2 metadata additions
+METADATA_REVISIONS['0.0.2'] = {'required': ['nsfw', 'ver'], 'optional': ['license_url']}
+CURRENT_METADATA_VERSION = '0.0.2'
```
jobevers: Add a version number inside the meta-data itself.  Without it, we'll end up trying to back-out what version we're working with and that will become a pain.

In the very near future we'll need to document what it means to change a version (one option: http://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/)

We should also consider using standardized json-schema validators: https://pypi.python.org/pypi/jsonschema

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+#v0.0.1 fee
+FEE_REVISIONS = {'0.0.1': {'required': ['amount', 'address'], 'optional': []}}
+CURRENT_FEE_REVISION = '0.0.1'
+
+
+class LBRYFeeFormat(dict):
```
jobevers: Please write some tests for this.  My brain can't handle tracing through the logic of something with 3 nested for loops and some if statements.

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, fee_dict):
+        dict.__init__(self)
+        self.fee_version = None
+        f = deepcopy(fee_dict)
+        assert len(fee_dict) == 1
+        for currency in fee_dict:
```
jobevers: Whenever I see a nesting, I try to make its body a function:

```
for currency in fee_dict:
    self.processCurrency(f, currency)
```

This helps keeps functions small.  Which makes them more understandable and lose prone to error.

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+            assert currency in CURRENCIES, "Unsupported currency: %s" % str(currency)
```
jobevers: the `%s` operator should automatically convert `currency` to a string

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+                    self[currency].update({k: f[currency].pop(k)})
+                for k in FEE_REVISIONS[version]['optional']:
+                    if k in fee_dict[currency]:
+                        self[currency].update({k: f[currency].pop(k)})
+                if not len(f):
+                    self.fee_version = version
+                    break
+            assert f[currency] == {}, "Unknown fee keys: %s" % json.dumps(f.keys())
+
+        self.amount = self[self.currency_symbol]['amount'] if isinstance(self[self.currency_symbol]['amount'], float) else float(self[self.currency_symbol]['amount'])
```
jobevers: This line is too long.  `self.amount = self.calculateAmount()`

I think the 80 character limit suggested by pep8 is too restrictive, but if you start going over 100 - think about doing it differently.

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+        self.address = self[self.currency_symbol]['address']
+
+
+class LBRYFee(LBRYFeeFormat):
```
jobevers: Add tests for this too

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, fee_dict, rate_dict):
+        LBRYFeeFormat.__init__(self, fee_dict)
+        rates = deepcopy(rate_dict)
+
+        assert 'BTCLBC' in rates and 'USDBTC' in rates
+        for fx in rate_dict:
+            assert int(time.time()) - int(rates[fx]['ts']) < 3600, "%s quote is out of date" % fx
```
jobevers: `assert self.is_recent_rate(`rates[fx]`), "%s quote is out of date" % fx`

lbrynet/core/LBRYMetadata.py
```diff
@@ -0,0 +1,135 @@
...
+        # log.error("%f * %f = %f" % (float(btc), self._BTCLBC['spot'], float(btc) * self._BTCLBC['spot'] / (1.0 - BITTREX_FEE)))
+        return float(btc) * self._BTCLBC['spot'] / (1.0 - BITTREX_FEE)
+
+    def _lbc_to_btc(self, lbc):
+        # log.error("lbc to btc: " + str(lbc))
+        # log.error("%f / %f = %f" % (self._BTCLBC['spot'], float(lbc), self._BTCLBC['spot'] / float(lbc)))
+        return self._BTCLBC['spot'] / float(lbc)
+
+
+class Metadata(dict):
```
jobevers: Tests this too 😄

lbrynet/core/LBRYWallet.py
```diff
@@ -19,15 +15,24 @@
...
+from lbrynet.interfaces import IRequestCreator, IQueryHandlerFactory, IQueryHandler, ILBRYWallet
+from lbrynet.core.client.ClientRequest import ClientRequest
+from lbrynet.core.Error import UnknownNameError, InvalidStreamInfoError, RequestCanceledError
+from lbrynet.core.Error import InsufficientFundsError
+from lbrynet.core.sqlite_helpers import rerun_if_locked
+from lbrynet.conf import SOURCE_TYPES
+from lbrynet.core.LBRYMetadata import Metadata
 
 log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
jobevers: remove this line

lbrynet/core/LBRYWallet.py
```diff
@@ -78,13 +83,59 @@ def __init__(self, db_dir):
         self._batch_count = 20
         self._first_run = self._FIRST_RUN_UNKNOWN
 
+        self._USDBTC = None
+        self._BTCLBC = None
+        self._exchange_rate_updater = task.LoopingCall(self._update_exchange_rates)
+
+    def _usd_to_btc(self):
+        if self._USDBTC is not None:
+            if int(time.time()) - int(self._USDBTC['ts']) < 600:
```
jobevers: `if self.is_recent_quote(self._USDBTC):`

lbrynet/core/LBRYWallet.py
```diff
@@ -315,24 +370,18 @@ def get_claim_for_name(claims):
...
-            for field in value_dict:
-                if field in OPTIONAL_METADATA_FIELDS:
-                    r_dict[field] = value_dict[field]
-
+            m = Metadata(value_dict)
         if 'txid' in result:
-            d = self._save_name_metadata(name, r_dict['sources']['lbry_sd_hash'], str(result['txid']))
-            d.addCallback(lambda _: r_dict)
+            d = self._save_name_metadata(name, str(result['txid']), m['sources']['lbry_sd_hash'])
+            d.addCallback(lambda _: log.info("lbry://%s complies with %s" % (name, m.metaversion)))
```
jobevers: log will do the string interpolation for you `log.info("lbry://%s complies with %s", name, m.metaversion)`

lbrynet/core/LBRYWallet.py
```diff
@@ -78,13 +83,59 @@ def __init__(self, db_dir):
         self._batch_count = 20
         self._first_run = self._FIRST_RUN_UNKNOWN
 
+        self._USDBTC = None
```
jobevers: I'm not convinced mixing the currency conversion logic inside of the wallet is the right path.  The wallet acts as an abstraction over lbrycrd/lbryum and is responsible for interacting with the bitcoin network.

In SOLID design (https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)) classes should have a single responsibility.  

I would suggest making a new class, maybe `CurrentExchangeRates`.  I would also make another class, `ExchangeRate`.  The later would be an abstraction for the the functions calls that you have (USD->BTC, BTC->LBC).  The former would act as a container for the `ExchangeRates` and be responsible for ensuring the freshness of the quotes.

lbrynet/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 log = logging.getLogger(__name__)
 logging.getLogger(__name__).addHandler(logging.NullHandler())
-log.setLevel(logging.ERROR)
+log.setLevel(logging.INFO)
```
jobevers: This is a note to me to delete this section - as with my other logging pull, log configuration is better handled by the caller, not at import time.

lbrynet/core/LBRYWallet.py
```diff
@@ -341,28 +390,15 @@ def _get_stream_info_from_value(self, result, name):
...
-        d = self._send_name_claim(name, json.dumps(value), bid)
+    def claim_name(self, name, bid, m):
+
+        metadata = Metadata(m)
+
+        d = self._send_name_claim(name, json.dumps(metadata), bid)
 
         def _save_metadata(txid):
-            d = self._save_name_metadata(name, value['sources']['lbry_sd_hash'], txid)
+            log.info("Saving metadata for claim %s" % txid)
```
jobevers: pylint can warn about this, I'll turn it on. log will do the interpolation for you.

lbrynet/core/LBRYWallet.py
```diff
@@ -520,19 +558,13 @@ def _open_db(self):
                                 "    txid text, " +
                                 "    sd_hash text)")
 
-    def _save_name_metadata(self, name, sd_hash, txid):
-        d = self.db.runQuery("select * from name_metadata where txid=?", (txid,))
+    def _save_name_metadata(self, name, txid, sd_hash):
+        d = self.db.runQuery("select * from name_metadata where name=? and txid=? and sd_hash=?", (name, txid, sd_hash))
```
jobevers: Sorry, I think I missed something.  Why is this necessary?

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -146,8 +146,8 @@ def __init__(self, root, wallet_type=None):
         self.connected_to_internet = True
         self.connection_problem = None
         self.query_handlers = {}
-        self.git_lbrynet_version = None
-        self.git_lbryum_version = None
+        self.pip_lbrynet_version = None
+        self.pip_lbryum_version = None
```
jobevers: can we stick with grabbing the version from `git` and undo all the changes in this file related to pkg_resources?

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -242,6 +243,8 @@ def __init__(self, root, wallet_type=None):
             self.session_settings['last_version'] = self.default_settings['last_version']
             self.first_run_after_update = True
             log.info("First run after update")
+            log.info("lbrynet %s --> %s" % (self.session_settings['last_version']['lbrynet'], self.default_settings['last_version']['lbrynet']))
```
jobevers: break this into multiple lines.  I generally prefer putting each argument on its own line:

```
log.info(
    "lbrynet %s --> %s",
    self.session_settings['last_version']['lbrynet'], 
    self.default_settings['last_version']['lbrynet']
)
```

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -603,6 +598,24 @@ def _check_connection_problems(self):
...
+            d = defer.succeed(self.pending_claims.pop(name))
+            d.addCallback(lambda _: self._get_lbry_file("name", name, return_json=False))
+            d.addCallback(lambda l: _start_file(l) if l.stopped else "LBRY file was already running")
+
+        for name in self.pending_claims:
+            log.info("Checking if new claim for lbry://%s is confirmed" % name)
+            d = self._resolve_name(name, force_refresh=True)
+            d.addCallback(lambda _: self._get_lbry_file_by_uri(name))
+            d.addCallbacks(lambda lbry_file: _start_file(name) if self.pending_claims[name] == lbry_file['txid'] and not isinstance(lbry_file['metadata'], str) else self._add_to_pending_claims(name, self.pending_claims.pop(name)),
+                           self._add_to_pending_claims(name, self.pending_claims.pop(name)))
```
jobevers: 😢

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -347,6 +350,7 @@ def __init__(self, root, wallet_type=None):
         self.internet_connection_checker = LoopingCall(self._check_network_connection)
         self.version_checker = LoopingCall(self._check_remote_versions)
         self.connection_problem_checker = LoopingCall(self._check_connection_problems)
+        self.pending_claim_checker = LoopingCall(self._check_pending_claims)
```
jobevers: I'd strongly consider pulling PendingClaims into its own class.

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1667,12 +1680,17 @@ def jsonrpc_resolve_name(self, p):
             metadata from name claim
         """
 
-        if 'name' in p.keys():
+        if 'force' in p:
```
jobevers: force = p.get('force', False)

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1889,21 +1907,44 @@ def jsonrpc_publish(self, p):
...
+                    d.addCallback(lambda _: self.session.wallet.get_new_address())
+                    d.addCallback(lambda addr: _set_address(addr, c))
 
         pub = Publisher(self.session, self.lbry_file_manager, self.session.wallet)
-
-        d = pub.start(name, file_path, bid, metadata, fee)
-        d.addCallbacks(lambda msg: self._render_response(msg, OK_CODE),
-                       lambda err: self._render_response(err.getTraceback(), BAD_REQUEST))
+        d.addCallback(lambda _: self._get_lbry_file_by_uri(name))
+        d.addCallbacks(lambda l: None if not l else _delete_data(l), lambda _: None)
```
jobevers: Try to avoid having any logic inside the lambda functions in these callbacks.  Tracing through twisted callbacks is hard enough.  Only use the lambda function to change the order or arguments, etc.

lbrynet/lbrynet_daemon/LBRYDaemonServer.py
```diff
@@ -67,6 +67,7 @@ def requestReceived(self, command, path, version):
         from twisted.web.http import parse_qs
         if self.do_log:
             print '%f Request Received' % time.time()
+            print self.content
```
jobevers: remove this.  Its garbage.  I removed it in an earlier commit - was there not a conflict?

lbrynet/lbrynet_daemon/LBRYDownloader.py
```diff
@@ -45,8 +47,7 @@ def __init__(self, sd_identifier, session, wallet, lbry_file_manager, max_key_fe
         self.wallet = wallet
```
jobevers: pass in the CurrentExchangeRate object here, along with the wallet

lbrynet/lbrynet_daemon/LBRYDownloader.py
```diff
@@ -85,27 +86,23 @@ def check_status(self):
...
-                self.key_fee_address = None
-        else:
-            self.key_fee = None
-            self.key_fee_address = None
-        if self.key_fee > self.max_key_fee:
-            log.info("Key fee %f above limit of %f didn't download lbry://%s" % (self.key_fee, self.max_key_fee, self.resolved_name))
-            return defer.fail(None)
-        else:
-            pass
+            self.fee = LBRYFee(self.stream_info['fee'], {'USDBTC': self.wallet._USDBTC, 'BTCLBC': self.wallet._BTCLBC})
```
jobevers: Following my suggestions from earlier, this would change to:
`self.fee = LBRYFee(self.stream_info['fee'], self.exchange_rates.get_rate_dict())`

jobevers:
I'm done with this pass.  I'd really like to see tests. There are some examples in `tests/lbrynet`.  For the metadata - It should be pretty easy to throw together some metadata input dictionaries, process them and and assert they match.

I'd also like to see the exchange rate logic stay out of the wallet and made a suggestion about how to go about that.

jobevers:
I threw together some example tests: https://github.com/lbryio/lbry/commit/6a1efedc1fb4ddf89079e6b2c7185448feeb1100

lbrynet/core/LBRYMetadata.py
jackrobison: I was doing str for in the case that someone manually posts a malformed fee, so if the assert errors the error message itself can't error

lbrynet/core/LBRYWallet.py
jackrobison: TIL, I thought it had to be in a tuple when there were multiple parameters

lbrynet/core/LBRYWallet.py
jackrobison: You can have multiple claims in the same tx. None of the keys are assuredly unique, but the set of all three is.

lbrynet/core/LBRYWallet.py
jackrobison: Done

lbrynet/core/LBRYMetadata.py
jackrobison: Done

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1165,6 +1166,12 @@ def _update_claim_cache(self):
         return defer.succeed(True)
 
     def _resolve_name(self, name, force_refresh=False):
+        try:
+            verify_name_characters(name)
+        except:
```
jobevers: avoid the bare exception

lbrynet/core/LBRYMetadata.py
```diff
@@ -14,6 +14,7 @@
 
 # Metadata version
 SOURCE_TYPES = ['lbry_sd_hash', 'url', 'btih']
+NAME_ALLOWED_CHARSET = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0987654321-'
```
jobevers: I'd allow underscores too

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -33,7 +33,7 @@
 from lbrynet.core.server.BlobAvailabilityHandler import BlobAvailabilityHandlerFactory
 from lbrynet.core.server.BlobRequestHandler import BlobRequestHandlerFactory
 from lbrynet.core.server.ServerProtocol import ServerProtocolFactory
-from lbrynet.core.Error import UnknownNameError, InsufficientFundsError
+from lbrynet.core.Error import UnknownNameError, InsufficientFundsError, InvalidNameError
```
jobevers: Not your fault... but a pet peeve of mine is redundancy in naming like this. `UnknownName`, `InsuffiecientFunds`, `InvalidName` all stand-alone.  There is no need to also append `Error` to them.

Don't fix.  I'm just venting

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1897,43 +1902,38 @@ def _clean(n):
...
-
-            return consolidated_results
+        def get_est_costs(results):
+            def _get_costs(search_result):
+                log.info("**" + search_result['name'])
+                d = self._get_est_cost(search_result['name'])
+                d.addCallback(lambda p: _save_cost(search_result, p))
+                return d
+
+            def _save_cost(value, cost):
```
jobevers: this should be renamed `_log_cost` or so

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1897,43 +1902,38 @@ def _clean(n):
...
+            log.info("Estimating costs")
+            dl = defer.DeferredList([_get_costs(r) for r in results], consumeErrors=True)
+            return dl
 
         log.info('Search nametrie: ' + search)
 
         d = self._search(search)
         d.addCallback(lambda claims: claims[:self.max_search_results])
-        d.addCallback(resolve_claims)
+        d.addCallback(get_est_costs)
```
jobevers: Using nested functions in the callbacks is a tolerable pattern when they're tiny.  If a nested function has other nested functions, its becomes difficult to read and impossible to re-use.  In the future, consider making separate classes or un-nested functions.  Those can be tested and re-used.

Also, the scoping rules on nested functions become difficult to follow which leads to bugs.

jobevers:
reviewed the search stuff, some fairly minor comments.

jobevers:
I will echo what @kauffj mentioned.  Ideally, a pull request will be a stand-alone feature.  This could have been:

1. change metadata to allow for versioning
1. add currency support
1. pending claims
1. get_claim_info 
1. search

Additionally, when possible, bug fixes are nice to apply on their own as that lets them get merged in quickly while development on the feature can continue.

jobevers:
Before merging - the pip version check stuff is still in place, which needs to be reverted back to checking github.

PR #109, opened 2016-08-02
add download_descriptor

jackrobison:


lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1072,6 +1072,24 @@ def _setup_lbry_file_opener(self):
...
+        def cb(result):
+            if not r.called:
+                r.callback(result)
+
+        def eb():
+            if not r.called:
+                r.errback(Exception("sd timeout"))
+
+        r = defer.Deferred(None)
+        reactor.callLater(3, eb)
```
jobevers: why 3?

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -1904,34 +1922,20 @@ def jsonrpc_search_nametrie(self, p):
...
-                    if thumb is None:
-                        tr['thumbnail'] = "img/Free-speech-flag.svg"
-                    tr['name'] = i[1][0]['name']
-                    tr['cost_est'] = i[1][1]
-                    t.append(tr)
+                td = {k: i['value'][k] for k in i['value']}
+                td['cost_est'] = float(i['cost'])
+                td['thumbnail'] = i['value'].get('thumbnail', "img/Free-speech-flag.svg")
+                td['name'] = i['name']
+                t.append(td)
```
jobevers: are these changes related to download descriptors?

jackrobison: Yes - lighthouse uses download_descriptor to cache sd blobs, which it then uses to calculate the stream costs. Previously, the daemon would calculate costs after getting the search results, now they're included in those results.

The (not very) _clean function was just to map the search return to the keys the ui uses. I think Alex has the UI querying lighthouse directly, and not through the daemon. In any case, when the UI handles the return keys 'cost', 'available', 'value', and 'name' then the _clean function can be deleted.

PR #110, opened 2016-08-03
Lighthouse client

jackrobison:
-add LighthouseClient as wrapper for search server

-move cost calculation to lighthouse, speed up search

-upnp redirects

-add get_peers_for_hash


lbrynet/lbrynet_daemon/Lighthouse.py
```diff
@@ -0,0 +1,28 @@
...
+from txjsonrpc.web.jsonrpc import Proxy
+from twisted.internet import defer
+from lbrynet.conf import  SEARCH_SERVERS
+
+log = logging.getLogger(__name__)
+
+
+class LighthouseClient(object):
+    def __init__(self):
+        self.servers = SEARCH_SERVERS
```
jobevers: allow callers to specify the servers:
```
def __init__(self, servers=None):
    self.servers = servers or SEARCH_SERVERS
```

jobevers:
For commit 3af8b7bd0ce518cead7d4aa56300bdaf223f66ac, can you explain why you're appending to the upnp entries?  I have a feeling in a few months when that code is revisited, one of us will appreciate the note.

jobevers:
same for request.notifyFinish - that looks like a bug fix, but some details on what bug it fixed would be appreciated.

PR #111, opened 2016-08-05
claim updates

jackrobison:
-update over existing claim in publish
-add claim_id to lbry files
-include supports in claim total
-add is_mine field to get_claim_info return

### DIFF
--- a/lbrynet/core/LBRYMetadata.py
+++ b/lbrynet/core/LBRYMetadata.py
@@ -90,10 +90,14 @@ class LBRYFeeValidator(dict):
 
 
 class Metadata(dict):
-    def __init__(self, metadata, is_hex=False):
+    @classmethod
+    def load_from_hex(cls, metadata):
+        return cls(json.loads(metadata.decode('hex')))
+
+    def __init__(self, metadata):
         dict.__init__(self)
         self.meta_version = None
-        metadata_to_load = deepcopy(metadata if not is_hex else json.loads(metadata.decode("hex")))
+        metadata_to_load = deepcopy(metadata)
 
         self._verify_sources(metadata_to_load)
         self._verify_metadata(metadata_to_load)
--- a/lbrynet/core/LBRYMetadata.py
+++ b/lbrynet/core/LBRYMetadata.py
@@ -133,3 +133,6 @@ class Metadata(dict):
 
     def serialize(self):
         return json.dumps(self).encode("hex")
+
+    def as_json(self):
+        return json.dumps(self)
--- a/lbrynet/core/LBRYMetadata.py
+++ b/lbrynet/core/LBRYMetadata.py
@@ -90,10 +90,10 @@ class LBRYFeeValidator(dict):
 
 
 class Metadata(dict):
-    def __init__(self, metadata):
+    def __init__(self, metadata, is_hex=False):
         dict.__init__(self)
         self.meta_version = None
-        metadata_to_load = deepcopy(metadata)
+        metadata_to_load = deepcopy(metadata if not is_hex else json.loads(metadata.decode("hex")))
 
         self._verify_sources(metadata_to_load)
         self._verify_metadata(metadata_to_load)
@@ -126,3 +126,6 @@ class Metadata(dict):
                     assert self.meta_version == self['ver'], "version mismatch"
                 break
         assert metadata == {}, "Unknown metadata keys: %s" % json.dumps(metadata.keys())
+
+    def serialize(self):
+        return json.dumps(self).encode("hex")
--- a/lbrynet/core/LBRYWallet.py
+++ b/lbrynet/core/LBRYWallet.py
@@ -595,7 +595,7 @@ class LBRYWallet(object):
 
     def _get_claim_metadata_for_sd_hash(self, sd_hash):
         d = self.db.runQuery("select name, txid from name_metadata where sd_hash=?", (sd_hash,))
-        d.addCallback(lambda r: r[0] if len(r) else None)
+        d.addCallback(lambda r: r[0] if r else None)
         return d
 
     def _update_claimid(self, claim_id, name, txid):
--- a/lbrynet/core/LBRYWallet.py
+++ b/lbrynet/core/LBRYWallet.py
@@ -1221,11 +1221,12 @@ class LBRYumWallet(LBRYWallet):
 
     def _send_name_claim_update(self, name, claim_id, txid, value, amount):
         def send_claim_update(address):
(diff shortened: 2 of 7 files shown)


lbrynet/core/LBRYMetadata.py
```diff
@@ -90,10 +90,10 @@ def _load_revision(self, version, f):
 
 
 class Metadata(dict):
-    def __init__(self, metadata):
+    def __init__(self, metadata, is_hex=False):
         dict.__init__(self)
         self.meta_version = None
-        metadata_to_load = deepcopy(metadata)
+        metadata_to_load = deepcopy(metadata if not is_hex else json.loads(metadata.decode("hex")))
```
jobevers: I'd move the logic for this into a class function.  Maybe:

```
@classmethod
def load_from_hex(cls, metadata):
    return cls(json.loads(metadata.decode("hex")))
```

In general `Boolean arguments loudly declare that the function does more than one thing. They are confusing and should be eliminated.`

lbrynet/core/LBRYMetadata.py
```diff
@@ -126,3 +126,6 @@ def _verify_metadata(self, metadata):
                     assert self.meta_version == self['ver'], "version mismatch"
                 break
         assert metadata == {}, "Unknown metadata keys: %s" % json.dumps(metadata.keys())
+
+    def serialize(self):
+        return json.dumps(self).encode("hex")
```
jobevers: why hexlify this?

lbrynet/core/LBRYWallet.py
```diff
@@ -324,6 +324,10 @@ def _check_result_fields(r):
             for k in ['value', 'txid', 'n', 'height', 'amount']:
                 assert k in r, "getvalueforname response missing field %s" % k
 
+        def _log_success(claim_id):
+            log.info("lbry://%s complies with %s, claimid: %s" % (name, metadata.meta_version, claim_id))
```
jobevers:
```
log.info("lbry://%s complies with %s, claimid: %s", name, metadata.meta_version, claim_id)
```

lbrynet/core/LBRYWallet.py
```diff
@@ -355,34 +366,68 @@ def _build_response(m, result):
...
+            d.addCallback(lambda claim: _filter_my_claims(claim) if claim is not False else False)
+        return d
 
-        d = self._send_name_claim(name, json.dumps(metadata), bid)
+    def claim_name(self, name, bid, m):
+        def _make_update(old_claim_info):
+            txid = old_claim_info['txid']
+            log.info("Updating from claim tx %s" % txid)
+            r = old_claim_info['value'] if isinstance(old_claim_info['value'], dict) else {}
+            for k in metadata:
```
jobevers: Avoid pulling in variables from the outer-scope.

lbrynet/core/LBRYWallet.py
```diff
@@ -355,34 +366,68 @@ def _build_response(m, result):
...
+                r[k] = metadata[k]
```
jobevers: Lines 416-418 could be made into their own function

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -2429,6 +2423,29 @@ def jsonrpc_get_peers_for_hash(self, p):
...
+        def _x(r):
+            log.info(str(r))
+            return r
+
+        def _get_metadata_and_txid(old_claim_info):
+            txid = old_claim_info['txid']
+            r = old_claim_info['value'] if isinstance(old_claim_info['value'], dict) else {}
+            for k in metadata:
+                r[k] = metadata[k]
+            return txid, json.dumps(Metadata(r))
```
jobevers: this is identical to the code early.  DRY.

lbrynet/core/LBRYWallet.py
```diff
@@ -355,34 +366,68 @@ def _build_response(m, result):
...
+                d.addCallback(lambda claims: next(c['claimId'] for c in claims if c['name'] == name))
+                d.addCallback(lambda cid: self._update_claimid(cid, name, txid))
+                return d
 
+        d = self._get_claimid_for_tx(name, txid)
+        d.addCallback(_get_id_for_return)
+        return d
 
-    def claim_name(self, name, bid, m):
+    def get_claim_info(self, name, force_good_metadata=True, is_mine=False):
```
jobevers: same complaint about boolean arguments here.  Functions should be small and *do one thing*. 

The requiring that the claim be your own can moved outside of this function and happen by the caller:

```
d = get_claim_info(...)
d.addCallback(ensure_my_claim)
```

For `force_good_metadata`, I need to think more about a concrete suggestion.  There probably needs to be a larger refactoring that allows `get_claim_info` to be extensible. (https://en.wikipedia.org/wiki/Open/closed_principle)

lbrynet/core/LBRYWallet.py
```diff
@@ -534,21 +575,38 @@ def handle_checks(results):
     def _open_db(self):
         self.db = adbapi.ConnectionPool('sqlite3', os.path.join(self.db_dir, "blockchainname.db"),
                                         check_same_thread=False)
-        return self.db.runQuery("create table if not exists name_metadata (" +
+
+        def create_tables(transaction):
```
jobevers: If we start adding more tables or doing more serious SQL work - we should look into sqlalchemy and alembic.  Otherwise we'll unnecessarily be reinventing the wheel for a lot of the basic db and data layer management.

lbrynet/lbrynet_daemon/LBRYDaemon.py
kauffj: My first tattoo is going to say DRY

lbrynet/core/LBRYMetadata.py
jackrobison: it isn't directly needed, if was while i was figuring out what formats go into _mktx in lbryum for claim id and metadata

lbrynet/core/LBRYWallet.py
```diff
@@ -1259,10 +1267,10 @@ def _send_support(d, a, n, c):
         return d
 
     def _broadcast_transaction(self, raw_tx):
-        log.info("Broadcast: %s" % str(raw_tx))
         cmd = known_commands['broadcast']
         func = getattr(self.cmd_runner, cmd.name)
         d = threads.deferToThread(func, raw_tx)
+        d.addCallback(lambda r: r if len(r) == 64 else Exception("Transaction rejected"))
```
jobevers: Does twisted automatically raise Exceptions, or how does this work?

PR #120, opened 2016-08-11
replace callLater with deferLater

jobevers:
We need the output of `_wait_on_lbry_file`, so `deferedLater` is more appropriate then `callLater`


jackrobison:
LGTM

PR #121, opened 2016-08-12
Save settings

jobevers:
This fixes: https://app.asana.com/0/158602294500137/165137057233590

### DIFF
--- a/lbrynet/lbrynet_console/LBRYSettings.py
+++ b/lbrynet/lbrynet_console/LBRYSettings.py
@@ -1,15 +1,25 @@
 import binascii
+import functools
 import json
-import unqlite
 import logging
 import os
+
 from twisted.internet import threads, defer
+import unqlite
 
 
 log = logging.getLogger(__name__)
 
 
+def run_in_thread(fn):
+    @functools.wraps(fn)
+    def wrapped(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


lbrynet/lbrynet_console/LBRYSettings.py
```diff
@@ -1,15 +1,25 @@
 import binascii
+import functools
 import json
-import unqlite
 import logging
 import os
+
 from twisted.internet import threads, defer
+import unqlite
```
jackrobison: Can we bump unqlite in requirements.txt up to the current version?

jobevers:
I looked into that, unqlite requires cython and I wasn't sure if it was worth adding the extra dependency and complexity in the build process.

But - seeing as using 0.5.3 makes the windows build easier, I've added https://app.asana.com/0/136290697597644/168474034092992

lbrynet/lbrynet_console/LBRYSettings.py
jobevers: I looked into that, but unqlite 0.5.3 requires cython and I wasn't sure it was worth the extra complexity in the build process.

But, since it makes the windows build easier, I've added: https://app.asana.com/0/136290697597644/168474034092992

PR #130, opened 2016-08-18
add os x install script to replace out of date and non existent one in lbry-setup

jackrobison:


lyoshenka:
I recommend you paste you script into http://www.shellcheck.net/ so it can check issues for you.

I have no way of testing this since I don't have OSX.

PR #131, opened 2016-08-18
change content-type to content_type, refactor metadata validation

jackrobison:


kauffj:
I think this can go up but I'd also like to set a time to discuss the way that this could be done better. @jackrobison connect with me on Slack and let's discus in real time.

PR #134, opened 2016-08-21
Add setting to show/hide NSFW content

sonatagreen:
For use with https://github.com/lbryio/lbry-web-ui/pull/36.


alexliebowitz:
We implemented this on the front end, but ended up deciding to go with local storage, so it doesn't need a daemon component. Thanks!

PR #135, opened 2016-08-22
LBRYcrdWallet update claim fix

jackrobison:
-fix log line that could raise an exception
-json encode value sent to lbrycrd-cli updateclaim


lyoshenka:
did not test, looks good to me

PR #137, opened 2016-08-26
More reliable stream producer

alexliebowitz:
Currently only supports direct streaming (i.e. no range requests).


jobevers:
Is this still WIP?

PR #146, opened 2016-08-30
Use new exclude_claimtrietx option in lbryum getbalance command, fix maturity use

kaykurokawa:
This uses the exclude_claimtrietx option in lbryum pull request https://github.com/lbryio/lbryum/pull/30

Thus the balance should now be displayed as the available amount of credits, not used for any claim trie transactions (claim,support,update). 

This also fixes an incorrect use of the field 'unmatured' which would have resulted in some strange behaviors if you held some new coins (probably someone had to patch this in during the early stage of testing where they were testing with unmatured coins ).  

This patches to show the fully confirmed balance only, but if we want to show the unconfirmed and insecure balance (with the confirmed balance) it should be result['confirmed'] + result['unconfirmed'] + result['unmatured'] , we could potentially  add another function to show the unconfirmed and insecure balance. 

I've not yet actually tested this patch myself.

### DIFF
--- a/lbrynet/core/LBRYWallet.py
+++ b/lbrynet/core/LBRYWallet.py
@@ -1200,8 +1200,10 @@ class LBRYumWallet(LBRYWallet):
     def get_balance(self):
         cmd = known_commands['getbalance']
         func = getattr(self.cmd_runner, cmd.name)
-        d = threads.deferToThread(func)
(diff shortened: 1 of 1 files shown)


kaykurokawa:
Was able to test this with lbryum and balance gets updated correctly when you make a claim ,

From a UX viewpoint, there is no way to know how much you spent in claims however.

PR #153, opened 2016-09-05
changed dependencies for ArchLinux

hpierce1102:
[Issue 151 - Installing on Arch Linux](https://github.com/lbryio/lbry/issues/151) 

Arch Linux was choking the unqlite dependency for whatever reason. I bumped that version number from 0.2.0 to 0.4.0 and subsequently updated setup tools to the newest version.

I'd like to throw in a disclaimer that I'm not a python dev, and have no idea what the implications of these changes are - but with these changes I managed to successfully install on Arch Linux where I couldn't originally.

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -1,18 +1,13 @@
 #!/usr/bin/env python
-"""Bootstrap setuptools installation
 
-To use setuptools in your package's setup.py, include this
-file in the same directory and add this to the top of your setup.py::
-
-    from ez_setup import use_setuptools
-    use_setuptools()
-
-To require a specific version of setuptools, set a download
-mirror, or use an alternate download directory, simply supply
-the appropriate options to ``use_setuptools()``.
-
-This file can also be run as a script to install or upgrade setuptools.
 """
+Setuptools bootstrapping installer.
+
+Maintained at https://github.com/pypa/setuptools/tree/bootstrap.
+
+Run this script to install or upgrade setuptools.
+"""
+
 import os
 import shutil
 import sys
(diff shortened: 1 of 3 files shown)


kauffj:
Awesome! There may be some obscure reason we're on unqlite 0.2.0. @jobevers or @jackrobison can you comment?

kauffj:
And if you're in our Slack, please DM me at kauffj so I know who you are. Great contribution!

abueide:
@hpierce1102 doing this seems to have broken my python 3, anything I try to run with python 3 spits out "ImportError: The 'packaging.version' package is required; normally this is bundled with this package so if you get this warning, consult the packager of your distribution." and i don't know how to fix it.

hpierce1102:
@abueide I can't confirm this, my python3 environment appears to be unaffected. Can you provide any more details? 

That error message suggests to me that you're importing a specific package that references a missing dependency (which it claims to include).  If you're *certain* that this is because of running setup.py, I might guess that something happened to python3's import paths. Though again, I can't confirm this, my import paths for python3 are more or less what I'd expect them to be.

@kauffj, I'm not currently a Slack user - but if this PR ends up getting any more complicated (as it appears to be) I'd look into joining.

abueide:
@hpierce1102 I fixed the problem, and re-ran setup.py seems to have worked fine. I think the error was probably caused by something else. I think it might have been another command I ran using setup.py (python setup.py build bdist_egg), which I found in RUNNING.md

kauffj:
@jimmykiselak are you running LBRY on Arch?

jobevers:
I'll fire up an instance of Arch and sort this out next week.

jobevers:
https://github.com/lbryio/lbry/pull/183

Issue #138, opened 2016-08-26
wallet should not try to send txouts with an amount of 0


jimmykiselak:
https://github.com/lbryio/lbry/blob/master/lbrynet/core/LBRYWallet.py#L283

lbrycrdd will soon reject txouts with an amount less than 1. This block of code should only run if the amount is greater than 0 to avoid ever inadvertently creating an invalid txout.

jackrobison:
https://github.com/lbryio/lbry/pull/142/files

PR #154, opened 2016-09-08
Allow 0 amount

kimihiro64:
We seem to already have 0 amounts in the claims, this assertion is breaking lighthouse


jackrobison:
https://github.com/lbryio/lbry/issues/138

PR #163, opened 2016-09-15
fix windows cli

jackrobison:
Make lbrynet-cli easier to use on windows by converting keyword args to json. Windows doesn't want to decode arguments given as json from powershell, this provides an alternative.

cli now accepts commands like:

`lbrynet-cli resolve_name name=one`

as well as json:

`lbrynet-cli resolve_name '{"name": "one"}'`

### DIFF
--- a/lbrynet/lbrynet_daemon/LBRYDaemonCLI.py
+++ b/lbrynet/lbrynet_daemon/LBRYDaemonCLI.py
@@ -1,5 +1,6 @@
 import sys
 import json
+import argparse
 
 from lbrynet.conf import API_CONNECTION_STRING
 from jsonrpc.proxy import JSONRPCProxy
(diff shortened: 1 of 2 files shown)


lbrynet/lbrynet_daemon/LBRYDaemonCLI.py
```diff
@@ -21,8 +34,28 @@ def main():
...
-    args = sys.argv[1:]
-    meth = args[0]
+    parser = argparse.ArgumentParser()
+    parser.add_argument('method', nargs=1, type=str)
+    parser.add_argument('params', nargs=argparse.REMAINDER, default=None)
+    args = parser.parse_args()
+    meth = args.method[0]
+    params = {}
+    if args.params:
+        if len(args.params) > 1:
```
jobevers: This code would be nicer as its own function - which could/should then be unit-tested

PR #165, opened 2016-09-16
add get_my_claim, remove is_mine from get_claim_info response

jackrobison:
This changes the return for `get_claim_info` to not include `is_mine` as a field, and instead adds a `get_my_claim` function to look for a users claims.

### DIFF
--- a/lbrynet/core/LBRYWallet.py
+++ b/lbrynet/core/LBRYWallet.py
@@ -368,6 +368,19 @@ class LBRYWallet(object):
         d.addCallback(_get_id_for_return)
         return d
 
+    def get_my_claim(self, name):
+        def _get_claim_for_return(claim):
+            if not claim:
+                return False
+            d = self.get_claim(name, claim['claim_id'])
+            d.addCallback(lambda c: self._format_claim_for_return(name, c, claim['txid']))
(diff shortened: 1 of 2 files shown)


lbrynet/core/LBRYWallet.py
```diff
@@ -368,6 +368,19 @@ def _get_id_for_return(claim_id):
...
+    def get_my_claim(self, name):
+        def _get_claim_for_return(claim):
+            if not claim:
+                return False
+            d = self.get_claim(name, claim['claim_id'])
+            d.addCallback(lambda c: self._format_claim_for_return(name, c, claim['txid']))
+            return d
+
+        d = self.get_name_claims()
+        d.addCallback(lambda claims: next((c for c in claims if c['name'] == name and not c['is spent']), False))
```
jobevers: this is doing too much on one line, I'd make a separate function: `d.addCallback(find_first_unspent, name)`
and, of course, unit test it.

Should it be `'is_spent'` instead of `'is spent'`

jackrobison: `is spent` is what comes out of lbryum and lbrycrdd for `getnameclaims`

PR #171, opened 2016-09-20
api sessions

jackrobison:


lbrynet/lbrynet_daemon/auth/util.py
```diff
@@ -0,0 +1,82 @@
...
+        return base58.b58decode(self.token())
+
+    def get_hmac(self, message):
+        decoded_key = self._raw_key()
+        signature = hmac.new(decoded_key, message, hashlib.sha256)
+        return base58.b58encode(signature.digest())
+
+    def compare_hmac(self, message, token):
+        decoded_token = base58.b58decode(token)
+        target = base58.b58decode(self.get_hmac(message))
```
lyoshenka: why are you b58decode-ing both the token and the target? i think if you're gonna compare them, comparing them encoded is the same as comparing them decoded? i'm not totally sure though

lbrynet/lbrynet_daemon/LBRYDaemon.py
```diff
@@ -40,6 +40,7 @@
 from lbrynet.lbrynet_daemon.LBRYPublisher import Publisher
 from lbrynet.lbrynet_daemon.LBRYExchangeRateManager import ExchangeRateManager
 from lbrynet.lbrynet_daemon.Lighthouse import LighthouseClient
+from lbrynet.lbrynet_daemon.auth.server import LBRYJSONRPCServer, auth_required, authorizer
```
lyoshenka: lets not use LBRY in names. all this stuff is LBRY stuff. its part of the lbrynet module. there's no reason to repeat that all over the place.

if you're in a refactoring mood, it would be great to remove it basically everywhere (e.g. LBRYPublisher, LBRYExchangeRateManager, etc)

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _check_headers(self, request):
+        origin = request.getHeader("Origin")
+        referer = request.getHeader("Referer")
+
+        if origin not in [None, 'http://localhost:5279']:
+            log.warning("Attempted api call from %s", origin)
+            raise InvalidHeaderError
+
+        if referer is not None and not referer.startswith('http://localhost:5279/'):
```
lyoshenka: This may prevent someone from opening an lbry:// link from another site. You should test this

lbrynet/lbrynet_daemon/LBRYDaemonCLI.py
```diff
@@ -36,7 +35,7 @@ def get_params_from_kwargs(params):
 
 
 def main():
-    api = JSONRPCProxy.from_url(API_CONNECTION_STRING)
+    api = LBRYAPIClient()
 
     try:
         s = api.is_running()
```
lyoshenka: what is `s`? please use descriptive names

lbrynet/lbrynet_daemon/LBRYDaemonControl.py
```diff
@@ -113,8 +116,14 @@ def start():
         if args.launchui:
             d.addCallback(lambda _: webbrowser.open(UI_ADDRESS))
 
-        lbrynet_server = server.Site(lbry.root)
+        checker = PasswordChecker()
+        realm = HttpPasswordRealm(lbry.root)
+        p = portal.Portal(realm, [checker, ])
```
lyoshenka: descriptive var name here too please

lbrynet/lbrynet_daemon/auth/auth.py
```diff
@@ -0,0 +1,50 @@
...
+from lbrynet.lbrynet_daemon.auth.util import load_api_keys, APIKey, API_KEY_NAME, save_api_keys
+from lbrynet.lbrynet_daemon.LBRYDaemon import log_dir as DATA_DIR
+
+log = logging.getLogger(__name__)
+
+
+# initialize api key if none exist
+if not os.path.isfile(os.path.join(DATA_DIR, ".api_keys")):
+    keys = {}
+    api_key = APIKey.new()
```
lyoshenka: cant you pass `name` as a param here instead of calling rename on the next line?

lbrynet/lbrynet_daemon/auth/client.py
```diff
@@ -0,0 +1,168 @@
...
+                                            'User-Agent': USER_AGENT,
+                                            'Authorization': self.__auth_header,
+                                            'Content-type': 'application/json'},)
+            r = req.prepare()
+            http_response = self.__conn.send(r)
+            cookies = http_response.cookies
+            self.__cookies = cookies
+            # print "Logged in"
+
+            uid = cookies.get('TWISTED_SESSION')
```
lyoshenka: `TWISTED_SESSION` should probably be a const

lbrynet/lbrynet_daemon/auth/client.py
```diff
@@ -0,0 +1,168 @@
...
+
+            self.__api_token = api_key.token()
+            self.__api_key_obj = api_key
+        else:
+            self.__auth_header = auth
+            self.__conn = connection
+            self.__cookies = cookies
+            self.__url = url
+
+            if cookies.get("secret", False):
```
lyoshenka: cookie name should be a constant, not hardcoded. also, the name should probably start with "lbry_", so its easier to debug in the browser later

lbrynet/lbrynet_daemon/auth/util.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, key, name=None):
+        self.key = key if isinstance(key, str) else key['token']
+        self.name = name if name else hashlib.sha256(self.key).hexdigest()
+        self.expiration = None if isinstance(key, str) else key.get('expiration', None)
+        self.update({self.name: {'token': self.key, 'expiration': self.expiration}})
+
+    @classmethod
+    def new(cls, expiration=None, seed=None, name=None):
+        key_val = generate_key(seed)
+        key = {'token': key_val, 'expiration': expiration}
```
lyoshenka: this is confusing. is `key` the key, or is `key_val`. it seems like `key_val` is the actual key, and `key` is a combination of the key and the expiration. maybe rename `key`?

in fact, why do you pass the key and expiration to `APIKey` as a dict, and then pass the name separately. why not pass all three as plain variables. what's the point of the dict?

lyoshenka: in fact, why does this extend `dict` at all? why not make a class with 3 properties (name, token, expiration)?

lbrynet/lbrynet_daemon/auth/util.py
```diff
@@ -0,0 +1,82 @@
...
+        old = self.keys()[0]
+        t = self.pop(old)
+        self.update({name: t})
+
+
+def load_api_keys(path):
+    if not os.path.isfile(path):
+        raise Exception("Invalid api key path")
+
+    f = open(path, "r")
```
jobevers: Use `with`:
```
with open(path, "r") as f:
    data = yaml.load(f.read())
```

lbrynet/lbrynet_daemon/auth/util.py
```diff
@@ -0,0 +1,82 @@
...
+    f.close()
+
+    keys = {key: APIKey(data[key], name=key)[key] for key in data}
+
+    return keys
+
+
+def save_api_keys(keys, path):
+    data = yaml.safe_dump(dict(keys))
+    f = open(path, "w")
```
jobevers: use `with` here too

lbrynet/lbrynet_daemon/auth/util.py
jackrobison: cleaned this up

jobevers:
I think we're still a ways from deciding how to handle this with regards to the UX.  If you removed `@AuthJSONRPCServer.auth_required` decorator from each method would it be functionally the same as it is now without authentication?

If so, maybe we should do that and get this merged in.

jobevers:
Assigning this to Jack so that nagging nancy leaves me alone.

jobevers:
Closing as this has all been moved into the other pull request "refactor-settings".  Please re-open if that is not the case.

PR #174, opened 2016-09-26
Remove LBRY Prefix

jobevers:
A good tedious task to pass a gloomy Sunday afternoon.


jimmykiselak:
Some of these are fine (LBRYSession -> Session, LBRYWallet -> Wallet) but I
would caution against changing LBRYFile to File. An LBRYFile is just one
particular method of breaking a regular plaintext file into chunks suitable
for being distributed via lbrynet and then recombining them. There are
other ways it can be done (for example, not encrypting the chunks, not
putting the encryption key directly in the stream descriptor, etc) and I
think it would be a good idea to design the system with that in mind.
LBRYFile is not a good name but I think File is a step in the wrong
direction.

jobevers:
I switched `File` to `EncryptedFile`

jobevers:
Probably better to get this in sooner rather then later as it touches 46 files and will be a pain to handle merge conflicts.

@jackrobison: thoughts?

PR #178, opened 2016-09-30
Track Data Uploaded and other stuff

jobevers:
I was on an airplane for six hours so there is a lot here and its not all related; its probably easier to review this commit by commit - I tried to break each one into logical sections.


jobevers:
A problem with working on the plane - the lack of internet means its hard to test. This doesn't work yet.
As an aside - getting end-to-end/functional tests working on osx would be great.

jobevers:
I tried, without success, to trigger the Blob Bytes Uploaded event - which I think highlights the continued need to have a better testing environment.  Fortunately, I think its covered under the under the functional tests that Jack fixed in the blob pricing pull request.

Once that is merged in, I'll rebase and re-run the functional tests to make sure this is working.

jobevers:
The main deliverables on this are the addition of two events.  One that tracks the amount of data uploaded and another that reports the total size of the blobs available to be uploaded.

https://app.asana.com/0/158829550589337/187745449149470
https://app.asana.com/0/158829550589337/187745449149472

lbrynet/analytics/events.py
```diff
@@ -23,30 +23,38 @@ def __init__(self, context, lbry_id, session_id):
...
+        return self._event('Heartbeat')
 
     def download_started(self, name, stream_info=None):
+        properties = {
+            'name': name,
+            'stream_info': get_sd_hash(stream_info)
+        }
+        return self._event('Download Started', properties)
+
+    def metric_observed(self, metric_name, value):
```
lyoshenka: How do you feel about enumerating all the possible metric names. idk what the pythonic way of doing this is. In php, I used class constants. In ruby, this would probably be a symbol. The idea is that this is not a string, its a unique identifier that represents the same thing everywhere in the code. pycharm should be able to tell me every place where this thing appears.

In general, I hate using strings for anything other that input/output.

lbrynet/analytics/track.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Track(object):
+    """Track and summarize observations of metrics."""
+    def __init__(self):
+        self.data = collections.defaultdict(list)
+
+    def add_observation(self, metric, value):
+        self.data[metric].append(value)
+
+    def summarize(self, metric, op=sum):
```
lyoshenka: i would call this something like `summarize_and_reset` so it's clear it has unexpected side effects. or i'd make reset a separate method

lbrynet/core/utils.py
```diff
@@ -1,16 +1,35 @@
 import base64
+import datetime
 import distutils.version
-import random
-import os
 import json
+import os
+import random
 import yaml
 import datetime
```
lyoshenka: have you been able to figure out how to get pycharm to alphabetically sort imports? they have an option for it, but it doesnt seem to work

lbrynet/core/utils.py
```diff
@@ -28,10 +51,7 @@ def is_valid_blobhash(blobhash):
     """
     if len(blobhash) != blobhash_length:
         return False
-    for l in blobhash:
-        if l not in "0123456789abcdef":
-            return False
-    return True
+    return all(is_valid_hashcharacter(l) for l in blobhash)
```
lyoshenka: :+1:

lyoshenka: Though if you're going that far, why not go all the way:

    return len(blobhash) == blobhash_length and all(is_valid_hashcharacter(l) for l in blobhash)

lbrynet/lbrynet_console/Console.py
```diff
@@ -366,11 +369,13 @@ def _setup_query_handlers(self):
         ]
 
         def get_blob_request_handler_factory(rate):
-            self.blob_request_payment_rate_manager = PaymentRateManager(self.session.base_payment_rate_manager,
-                                                                        rate)
-            handlers.append(BlobRequestHandlerFactory(self.session.blob_manager,
-                                                      self.session.wallet,
-                                                      self.blob_request_payment_rate_manager))
+            self.blob_request_payment_rate_manager = PaymentRateManager(
+                self.session.base_payment_rate_manager, rate
```
lyoshenka: another pycharm question: how do you get it to do alignment correctly? i feel like its really stupid about autoformatting long lines

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -353,12 +426,15 @@ def __init__(self, root, wallet_type=None):
...
 
-        self.internet_connection_checker = LoopingCall(self._check_network_connection)
-        self.version_checker = LoopingCall(self._check_remote_versions)
-        self.connection_problem_checker = LoopingCall(self._check_connection_problems)
-        self.pending_claim_checker = LoopingCall(self._check_pending_claims)
-        self.send_heartbeat = LoopingCall(self._send_heartbeat)
-        # self.lbrynet_connection_checker = LoopingCall(self._check_lbrynet_connection)
+        self.looping_call_manager = LoopingCallManager()
+        looping_calls = [
+            ('internet_connection_checker', CheckInternetConnection(self)),
```
lyoshenka: these strings should also not be strings. again, not sure how to handle this in python

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2838,3 +2771,123 @@ def is_in_cache(self):
...
+        self.return_json = return_json
+
+    def retrieve_file(self):
+        d = self.search_for_file()
+        if self.return_json:
+            d.addCallback(self._get_json)
+        return d
+
+    def search_for_file(self):
+        if self.search_by == "name":
```
lyoshenka: more evil strings. after reading all this code, i think they should be constants

lbrynet/analytics/events.py
jobevers: I am using `analytics.constants` for this

lbrynet/core/utils.py
jobevers: I use emacs but, also no idea how to sort there either.

lbrynet/lbrynet_console/Console.py
jobevers: The alignment comes from my own OCD issues

PR #182, opened 2016-10-07
NoCacheStaticFile

jackrobison:


lbrynet/lbrynet_daemon/Resources.py
```diff
@@ -0,0 +1,181 @@
...
+class NoCacheStaticFile(static.File):
+    def render_GET(self, request):
+        """
+        Begin sending the contents of this L{File} (or a subset of the
+        contents, based on the 'range' header) to the given request.
+        """
+        self.restat(False)
+        request.setHeader('cache-control', 'no-cache, no-store, must-revalidate')
+        request.setHeader('expires', '0')
+
```
jobevers: Can this be changed to?

```
def render_GET(self, request):
    self.add_no_cache(request)
    return static.File.render_GET(self, request)
```

PR #183, opened 2016-10-10
upgrade unqlite to 0.4.0

jobevers:
unqlite 0.2.0 was causing problems on arch linux.  Upgrading to
the current version (0.6.0) causing difficulties in building
the package so 0.4.0 it is.

Eventually it would be nice to get rid of unqlite.


jobevers:
I have no idea what bug we're running into on the build.  I'd rather put the effort into getting rid of unqlite instead of tracking down this bug.

lyoshenka:
@jobevers please delete branch if this is no longer needed

PR #187, opened 2016-10-14
Rewrite of migration and validation system using JSON Schema

alexliebowitz:
This refactors the Validator class into a new class called StructuredDict. It does all validation using JSON Schema (so far no need for Python code, but if we need it, JSON Schema has good support for adding custom validation rules).

Migration now uses methods that operate directly on the dictionary instead of a DSL.


lbrynet/metadata/metadata_schemas.py
```diff
@@ -0,0 +1,269 @@
+VER_001 = {
+    '$schema': 'http://json-schema.org/draft-04/schema#',
+    'title': 'LBRY metadata schema 0.0.1',
+    'definitions': {
+        'fee_info': {
```
alexliebowitz: The fee_info sub-schema is duplicated from Fee.py. If we found a way to use JSON files instead of hard coding it in Python, we could use JSON Schema's `$ref` keyword to pull in the schema from the other file instead.

lbrynet/metadata/StructuredDict.py
```diff
@@ -0,0 +1,57 @@
...
+    def migrate(self, target_version=None):
+        if target_version:
+            assert self._versions.index(target_version) > self.versions.index(self.version), "Current version is above target version"
+
+        above_starting_version = False
+        for version, schema, migration in self._versions:
+            if not above_starting_version:
+                if version == self.version:
+                    above_starting_version = True
+                continue
```
jobevers: I think its worth abstracting this logic away.  Maybe something like:

```
for version, schema, migration in skip_lower_version(self._versions, self.version):
```

```
def skip_lower_versions(versions, target_version):
    for version, schema, migration in versions:
        if version >= target_version:
            yield version, schema, migration
```

lbrynet/metadata/Metadata.py
```diff
@@ -14,73 +14,28 @@ def verify_name_characters(name):
...
 
-    METADATA_MIGRATIONS = [
-        MIGRATE_MV001_TO_MV002,
-        MIGRATE_MV002_TO_MV003,
-    ]
 
-    current_version = CURRENT_METADATA_VERSION
-    versions = METADATA_REVISIONS
-    migrations = METADATA_MIGRATIONS
+    def _migrate_001_to_002(self):
```
jobevers: I think the migrations would be cleaner if they were functional:

```
def _migrate_001_to_002(data):
    result = copy.deepcopy(data)
    result['ver'] = '0.0.2'
    return result
```

lbrynet/metadata/Metadata.py
```diff
@@ -14,73 +14,28 @@ def verify_name_characters(name):
...
+        self['ver'] = '0.0.2'
 
-    def __init__(self, metadata, process_now=True):
-        Validator.__init__(self, metadata, process_now)
-        self.meta_version = self.get('ver', Metadata.MV001)
-        self._load_fee()
+    def _migrate_002_to_003(self):
+        if self['ver'] == '0.0.2':
```
jobevers: This should be an assert

lbrynet/metadata/StructuredDict.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, value, starting_version, migrate=True, target_version=None):
+        dict.__init__(self, value)
+
+        self.version = starting_version
+        self._schemas = dict([(version, schema) for (version, schema, _) in self._versions])
+
+        self.validate(starting_version)
+
+        if migrate:
+            self.migrate(target_version)
```
jobevers: I like the pattern of always having dumb constructors and moving logic into classmethods.

jobevers:
Tests are failing because jsonschema is missing.  You'll want to add it to the requirements.txt and setup.py files.

lbrynet/metadata/Metadata.py
```diff
@@ -14,73 +14,27 @@ def verify_name_characters(name):
...
 
-    METADATA_REVISIONS[MV003] = [
-        (Validator.REQUIRE, 'content_type', skip_validate),
-        (Validator.SKIP, 'content-type'),
-        (Validator.OPTIONAL, 'sig', skip_validate),
-        (Validator.IF_KEY, 'sig', (Validator.REQUIRE, 'pubkey', skip_validate), Validator.DO_NOTHING),
-        (Validator.IF_KEY, 'pubkey', (Validator.REQUIRE, 'sig', skip_validate), Validator.DO_NOTHING),
-    ]
+    def __init__(self, metadata, migrate=True, target_version=None):
+        self._versions = [
```
jobevers: this would be better as a class variable not an instance variable.

It does mean that `_migrate_001_to_002`, etc will need to be static methods that take a dictionary as an argument and modify it.

jobevers:
travis is being difficult. with their osx build.  I'm going to merge it in and trigger a build and publish on master. The tests pass on the linux build so I'm not worried.

Issue #190, opened 2016-10-17
publish fails in an un-obvious way when given an invalid file


jackrobison:
functionName: jsonrpc_publish
		asciTime: 2016-10-17 18:06:28,026
		fileName: Daemon.py
		levelNo: 20
		lineNo: 2008
		message: {u'file_path': None, u'bid': 50, u'name': u'football', u'metadata': {u'description': u'Enjoy!', u'license': u'Creative Commons Attribution 4.0 International', u'author': u'Cataclysmic Gaming', u'title': u'Epic Football Challenges!', u'language': u'en', u'license_url': u'https://creativecommons.org/licenses/by/4.0/legalcode', u'nsfw': False, u'thumbnail': u'http://www.thefa.com/~/media/images/thefaportal/pillars/governance/nike-ball-wembley-netherlan.ashx'}}
		loggerName: lbrynet.lbrynet_daemon.Daemon
		levelName: INFO

		functionName: start
		asciTime: 2016-10-17 18:06:28,295
		fileName: Publisher.py
		levelNo: 20
		lineNo: 50
		message: Starting publish for football
		loggerName: lbrynet.lbrynet_daemon.Publisher
		levelName: INFO

		functionName: _show_publish_error
		asciTime: 2016-10-17 18:06:28,296
		fileName: Publisher.py
		levelNo: 40
		lineNo: 157
		message: coercing to Unicode: need string or buffer, NoneType found
		loggerName: lbrynet.lbrynet_daemon.Publisher
		levelName: ERROR

		functionName: _ebRender
		asciTime: 2016-10-17 18:06:28,296
		fileName: Daemon.py
		levelNo: 40
		lineNo: 490
		message:  [Failure instance: Traceback (failure with no frames): <type 'exceptions.Exception'>: Publish failed ]
		loggerName: lbrynet.lbrynet_daemon.Daemon
		levelName: ERROR

PR #191, opened 2016-10-17
make publish fail when given an invalid file

jackrobison:
https://github.com/lbryio/lbry/issues/190

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -2010,11 +2010,16 @@ class Daemon(jsonrpc.JSONRPC):
             metadata = Metadata(p['metadata'])
             make_lbry_file = False
             sd_hash = metadata['sources']['lbry_sd_hash']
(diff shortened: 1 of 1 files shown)


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2010,11 +2010,16 @@ def _reflect_if_possible(sd_hash, txid):
...
+            log.info("Update publish for %s using existing stream", name)
         except ValidationError:
             make_lbry_file = True
             sd_hash = None
             metadata = p['metadata']
             file_path = p['file_path']
+            if not file_path:
+                return defer.fail(Exception("No file given to publish"))
+            if not os.path.isfile(file_path):
+                return defer.fail(Exception("Specified file for publish doesnt exist: %s" % str(file_path)))
```
jobevers: the `str` isn't necessary as `%s` already does that: https://docs.python.org/2/library/stdtypes.html#string-formatting-operations

PR #197, opened 2016-10-19
Rename DHT tests such that trial automatically runs them

jobevers:
There are still tests in `tests/dht` that I didn't rename/move because they aren't currently passing.


jobevers:
trivial enough.  merging

PR #198, opened 2016-10-20
fix problems in blobrequester

jackrobison:
-also move is_generous_host into a conf setting

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -3,7 +3,7 @@ Some network wide and also application specific parameters
 """
 import os
 
-
+is_generous_host = True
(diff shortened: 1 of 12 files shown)


jobevers:
Generally passing a boolean parameter is a code smell and having separate functions / classes is a better pattern.

```Boolean arguments loudly declare that the function does more than one thing. They are confusing and should be eliminated.```
http://www.informit.com/articles/article.aspx?p=1392524

This article http://martinfowler.com/bliki/FlagArgument.html also has some good commentary.

In the case of `is_generous`, it would be better to have a `Generous` strategy and a non-generous one.

Might be worth refactoring some of this when you have some down-time.

PR #200, opened 2016-10-20
Start to cleanup functional tests to improve readability

jobevers:
No functionality was changed, just made some simple refactorings.

- delete redundant Fake objects
- Move Streamify tests into their own module
- Refactor start_lbry_uploader to be a class instead of a function with a bunch of inner functions 
   - I like this pattern much much better as it makes it significnately more clear what state is being shared


jobevers:
Merging this as its test only refactoring.

PR #186, opened 2016-10-14
add nOut to blockchainname.db (WIP)

jackrobison:
-save nOut in claim dbs, instead of just using txid
-delete lbry files for abandoned claims, otherwise they get checked repeatedly

PR #202, opened 2016-10-20
work on jack's add-nout branch

kaykurokawa:
Further work on jack's branch add-nout to integrate nout : https://github.com/lbryio/lbry/pull/186 . 

This pull request also integrates unmerged new lbryum claim transaction commands https://github.com/lbryio/lbryum/pull/34 

This patch changes the daemon json rpc commands support_claim, abandon_claim, and publish . As such some changes to lbry-web-ui is probably required. 

This patch will also break lbrycrd support if merged, as it does not have txid + nout support for claim transactions. 

I was able to test the publishing successfully through the web ui and the abandon_claim / support_claim through the cli. But further testing is needed as this is a major change.

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -307,12 +307,12 @@ class Wallet(object):
         d.addCallback(lambda r: None if 'txid' not in r else r['txid'])
         return d
 
-    def get_stream_info_from_txid(self, name, txid):
+    def get_stream_info_from_txid(self, name, txid, nout):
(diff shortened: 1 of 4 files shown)


kaykurokawa:
During migration, n should be filled in as -1 and correct value should be percolated during run

jobevers:
Merging this as it is only going into the add-nout branch.

@kaykurokawa  - you should have write access to the lbry repo.  Let me know if you don't.

PR #204, opened 2016-10-24
Cleanup logging

jobevers:
- Make the logging across the OSX app, the windows app and the CLI consistent.
 - Remove the repeated code for setting up the log file
    - Moved logic for log file upload into its own class
 - change the logic for the `--verbose` flag:
   - if no arguments are specified after `--verbose`, set all `lbrynet` loggers to debug
   - if arguments are specified after `--verbose` only set those loggers to debug

The idea with the --verbose flag is that we will be able set the modules that we're working on to be at the debug level.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -3,9 +3,11 @@ import os
 import sys
 from appdirs import user_data_dir
 
+
 LINUX = 1
 DARWIN = 2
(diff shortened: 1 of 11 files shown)


jobevers:
Addresses https://app.asana.com/0/158829550589337/198385575656803

PR #205, opened 2016-10-24
Migration fixes for add-nout branch

kaykurokawa:
Fixing migration for add-nout branch. 

Fix migration to set nout to -1 instead of 0, have Wallet.py percolate correct nout in _save_name_metadata and _update_claimid

Disable migration in case there is no blockchainname.db file (for fresh installs) , otherwise migration will throw an exception and lbry will be stuck on startup.


lbrynet/db_migrator/migrate1to2.py
```diff
@@ -32,11 +36,12 @@ def migrate_blockchainname_db(db_dir):
     name_metadata = file_cursor.execute("select * from name_metadata").fetchall()
     claim_metadata = file_cursor.execute("select * from claim_ids").fetchall()
 
+    # fill n as -1, Wallet.py will be responsible for filling in correct n 
     for name, txid, sd_hash in name_metadata:
-        mem_cursor.execute("insert into name_metadata values (?, ?, ?, ?) ", (name, txid, 0, sd_hash))
+        mem_cursor.execute("insert into name_metadata values (?, ?, ?, ?) ", (name, txid, -1, sd_hash))
```
jobevers: can the `-1` be made a constant, perhaps `NOUT_UNSET`.  I dislike [magic numbers](http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad) almost as much as grin dislikes strings.

lbrynet/core/Wallet.py
```diff
@@ -638,10 +638,16 @@ def _clean_bad_records(self):
         d = self.db.runQuery("delete from name_metadata where length(txid) > 64 or txid is null")
         return d
 
+
     def _save_name_metadata(self, name, txid, nout, sd_hash):
         assert len(txid) == 64, "That's not a txid: %s" % str(txid)
         d = self.db.runQuery("delete from name_metadata where name=? and txid=? and n=? and sd_hash=?",
                              (name, txid, nout, sd_hash))
+        # delete legacy v1 database entry with missing nouts
```
jobevers: I like to avoid comments that explain what the code is doing; I believe comments should be reserved for explanations of why the code is as it is.

If a comment is necessary, consider making changes to improve the readability of the code.  In this case, I'd make a function call: `d.addCallback(lambda _: self.delete_legacy_entries(name, txid, sd_hash))`

kaykurokawa:
changing -1 to constant UNSET_NOUT declared in migrate1to2 as per jobever's comments

kaykurokawa:
Closing this and will open a new pull request after some rebasing

PR #211, opened 2016-10-26
Refactor settings

jackrobison:
-make settings configurable with environmental variables

-turn authenticated api on/off


jobevers:
@jackrobison  - I made a change to this requiring that the environment variables be prefixed/namespaced with `LBRY_` and that they are all uppercase.  This is pretty standard for environment variables

lbrynet/conf.py
```diff
@@ -226,3 +226,11 @@ def API_CONNECTION_STRING(self):
     @property
     def UI_ADDRESS(self):
         return "http://%s:%i" % (DEFAULT_SETTINGS.API_INTERFACE, self.api_port)
+
+
+# TODO: don't load the configuration automatically. The configuration
```
jobevers: @jackrobison - a note for future work.

jobevers:
Moved the settings into the `conf.py` file.  `lbrynet/__init__.py` is loaded in `setup.py` (https://github.com/lbryio/lbry/blob/master/setup.py#L6) so having imports and dependencies in there was causing problems in the build

jobevers:
Merging as I think its better to get this in and fix up the warts than let the pull request linger any longer.

lbrynet/lbrynet_daemon/DaemonControl.py
```diff
@@ -96,24 +126,34 @@ def start():
...
         print "To quit press ctrl-c or call 'stop' via the API"
 
     if test_internet_connection():
         lbry = DaemonServer()
 
-        d = lbry.start(branch=args.branch if args.branch else DEFAULT_UI_BRANCH,
-                       user_specified=args.ui,
-                       wallet=args.wallet,
-                       branch_specified=True if args.branch else False)
+        d = lbry.start()
```
jobevers: Moving from arguments to a global configuration object is an anti-pattern. I accept that there will be some global configuration because that is how things were before this pull request, but we shouldn't be adding more global state.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1158,7 +904,7 @@ def eb():
 
         return r
 
-    def _download_name(self, name, timeout=DEFAULT_TIMEOUT, download_directory=None,
+    def _download_name(self, name, timeout=lbrynet_settings.download_timeout, download_directory=None,
```
jobevers: default argument values should always be immutable.  You'll run into weird behavior otherwise.  This would be better as 
```
def _download_name(self, name, timeout=None, ...):
    if timeout is None:
        timeout = lbrynet_settings.download_timeout
```

PR #213, opened 2016-10-27
Log error when resolve name fails

jobevers:


jobevers:
Don't merge this until the refactor settings branch is in, I accidently based this off of that.

lbrynet/core/log_support.py
```diff
@@ -152,4 +152,5 @@ def failure(failure, log, msg, *args):
         args: values to substitute into `msg`
     """
     args += (failure.getErrorMessage(),)
-    log.error(msg, *args, exc_info=failure.getTracebackObject())
+    exc_info = (failure.type, failure.value, failure.getTracebackObject())
```
jobevers: This was only working previously by accident

PR #215, opened 2016-10-28
delete RUNNING.md

lyoshenka:
this file is way out of date. if we still need it, please rewrite it. otherwise, lets delete it.


jobevers:
agreed, not useful.

PR #216, opened 2016-10-28
Channels metadata (WIP)

alexliebowitz:
Adds a channel field to the stream metadata schema, and creates a new versioned schema for the metadata in channel claims.


jobevers:
first pass: seems fine

jackrobison:
I'm going to close this, the WIP changes over to using protobufs and will likely end up in its own repo

PR #226, opened 2016-10-31
Start listening before launching browser

jobevers:
Hopefully, by waiting to launch the browser until after listenTCP has happened, we should see less people complaining about localhost refusing to connect.


jobevers:
Hopefully addresses https://app.asana.com/0/198587529699250/204729611542523

jobevers:
Lgtmd by Jeremy in slack

PR #228, opened 2016-10-31
Read config file in windows and osx

jobevers:
Also moves the config loading to BEFORE the log configuration

### DIFF
--- a/lbrynet/lbrynet_daemon/DaemonControl.py
+++ b/lbrynet/lbrynet_daemon/DaemonControl.py
@@ -69,22 +69,12 @@ def start():
     args = parser.parse_args()
 
     utils.setup_certs_for_windows()
-    lbrynet_log = conf.get_log_filename()
-    log_support.configure_logging(lbrynet_log, args.logtoconsole, args.verbose)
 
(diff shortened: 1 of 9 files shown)


lbrynet/lbrynet_daemon/DaemonControl.py
```diff
@@ -119,13 +107,24 @@ def start():
...
 
+def update_settings_from_args(args):
+    to_pass = {}
+    if args.ui:
+        to_pass['local_ui_path'] = args.ui
+    if args.branch:
+        to_pass['ui_branch'] = args.branch
+    to_pass['use_auth_http'] = args.useauth
+    to_pass['wallet_type'] = args.wallet
+    log.debug('Settings overrides from cmd line: %s', to_pass)
```
jobevers: oops, logging hasn't been configured yet

jobevers:
Checking os.path.isdir creates a race-condition.  I'll do:
http://stackoverflow.com/a/14364249

jobevers:
Build on linux passed, travis seems slow on osx so I'm going to merge so another release can be cut.

PR #229, opened 2016-10-31
Fix updates in name_claim function 

kaykurokawa:
This patch removed is_mine and broke the name_claim function in Wallet.py in cases when we were updating : https://github.com/lbryio/lbry/pull/165/files because there was no longer a 'is_mine" field in claims.

We can now use get_my_claim to determine if we have made claim over a name, which is the correct behaviour over using get_claim_info . ( since we could have made a non winning claim)


jobevers:
LGTM

PR #230, opened 2016-11-01
Better error detection in Wallet for invalid streams

kaykurokawa:
_get_stream_info_from_value and _get_claim_info function in Wallet could fail with uninformative exceptions in the case where the value of a claim is not a json dictionary, or is not a json string. 

Have Metadata class detect when it is fed an non dictionary item as metadata and throw a TyepError exception.  

Catch  TypeError exceptions from Metadata and ValueError exceptions from failed json.loads('...') calls. 

Change InvalidStreamInfoError to contain the stream info.

### DIFF
--- a/lbrynet/core/Error.py
+++ b/lbrynet/core/Error.py
@@ -53,11 +53,12 @@ class InvalidStreamDescriptorError(Exception):
 
 
 class InvalidStreamInfoError(Exception):
-    def __init__(self, name):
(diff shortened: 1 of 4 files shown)


jobevers:
One final thing, just me being picky - have a space after your commas.

kaykurokawa:
Fixed as per comments.

PR #236, opened 2016-11-02
More robustly stop the reactor

jobevers:
Was getting a "cannot stop a reactor that isn't running error"
when playing around with error handling on startup.

Following the suggestion from http://stackoverflow.com/a/36738480
it seems to work better


jobevers:
small fix, merging

PR #238, opened 2016-11-02
fix missing conf file on first run

jackrobison:
This fixes a bug where the daemon tries to read the conf file, whether or not it exists.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -271,12 +271,15 @@ class Config(DefaultSettings):
 
 def update_settings_from_file(filename=None):
     filename = filename or settings.get_conf_filename()
-    try:
-        updates = load_settings(filename)
-        log.info("Loaded settings file: %s", updates)
-        settings.update(updates)
-    except OSError as ex:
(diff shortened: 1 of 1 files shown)


lbrynet/conf.py
```diff
@@ -271,12 +271,15 @@ def get_conf_filename(self):
...
-        log.info("Loaded settings file: %s", updates)
-        settings.update(updates)
-    except OSError as ex:
-        log.info('%s: Failed to update settings from %s', ex, filename)
+    if os.path.isfile(filename):
+        try:
+            updates = load_settings(filename)
+            log.info("Loaded settings file: %s", updates)
+            settings.update(updates)
+        except OSError as ex:
```
jobevers: generally its better to ["beg for forgiveness"](http://stackoverflow.com/a/12270384).  If `filename` is missing, it will throw an IOError.  (Which was my mistake, I should have caught IOError when I wrote this)

I'd suggest removing the os.path.isfile check and doing

```
try:
    ...
except (IOError, OSError) as ex:
    ...
```

jobevers:
LGTM. Merging.

Issue #237, opened 2016-11-02
On OSX, on shutdown, the goodbye notification shows up twice


jobevers:
The goodbye notification shows up twice before the application quits

PR #242, opened 2016-11-03
Shutdown issues

jobevers:
- Switch the osx app to use applicationShouldTerminate_ instead of replyToApplicationShouldTerminate_
- Add more logging to the shutdown process
- cleanup whitespace

Fixes https://github.com/lbryio/lbry/issues/237


jobevers:
LGTM'd by @alexliebowitz in slack

PR #249, opened 2016-11-05
fix --wallet argument bug, use .json conf file if not given a .yml file

jackrobison:


jobevers:
Sorry.  Coveralls shouldn't fail a build because of a decrease in coverage.  Not sure why its doing that.

jobevers:
Also, please assign each pull request in the future.

lbrynet/lbrynet_daemon/DaemonControl.py
```diff
@@ -115,7 +115,7 @@ def update_settings_from_args(args):
     if args.branch:
         to_pass['ui_branch'] = args.branch
     to_pass['use_auth_http'] = args.useauth
-    to_pass['wallet_type'] = args.wallet
+    to_pass['wallet'] = args.wallet
```
jobevers: worth considering, in the future, adding a check in `settings.update()` to require that the setting already exists, would catch errors like this.

jobevers: also, sorry, I made this mistake :/

jobevers:
LGTM

PR #252, opened 2016-11-08
Bundled UI

jobevers:


jobevers:
Will probably still have to fight with the CI tools for a while to get the S3 upload right!

packaging/travis/setup_osx.sh
```diff
@@ -7,4 +7,6 @@ wget https://www.python.org/ftp/python/2.7.11/python-2.7.11-macosx10.6.pkg
 sudo installer -pkg python-2.7.11-macosx10.6.pkg -target /
 pip install -U pip
 brew install gmp
-
+# GNU vs BSD sed handles in-place every so slightly different
+# so its worth installing the GNU version
+brew install gnu-sed --with-default-names
```
jobevers: This can be taken out. I wrote a python script instead of using sed

jobevers:
This changes the build process to include the UI code from the development branch in the package for all non-production builds. The UI manager will prefer the bundled UI and also not fetch updates.

The version is also changed,  with the commit sha appended. Otherwise it might be difficult to tell if we are running a production release or a QA one. 

Forgive all of the commits. I'll clean them up tomorrow.

jobevers:
Squashed in another branch, will re-open there.

PR #258, opened 2016-11-10
Fix deferred for loading user-specified UIs

alexliebowitz:
In UIManager, pass self._load_ui into deferred as function object instead of calling immediately


jobevers:
Thanks for catching this. Bugs like this are one reason why a good test suite is important.

PR #260, opened 2016-11-10
meaningful error messages, improve cli, remove unused search from daemon

jackrobison:


jobevers:
This needs to be broken apart into separate commits.  I'd suggest doing

```
$ git reset HEAD~1
```

Which will set you back one commit:
```
$ git status
On branch error-messages
Your branch is behind 'origin/error-messages' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   lbrynet/conf.py
	modified:   lbrynet/dht/protocol.py
	modified:   lbrynet/lbrynet_daemon/Daemon.py
	modified:   lbrynet/lbrynet_daemon/DaemonCLI.py
	deleted:    lbrynet/lbrynet_daemon/Lighthouse.py
	modified:   lbrynet/lbrynet_daemon/auth/server.py
```

From here, you can do `git add -p` and selectively add and commit each piece. There are probably 3-5 separate commits in here.

The commit that puts reactor.iterate() back in place should include a commit message explaining the error that happens without that line of code so that when we go to look at it in the future we'll have better context for why its there.

 After making all of the commits, do a `$ git push -f` to force the push and overwrite the original commit.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -125,36 +143,35 @@ def expire_session():
...
         if self._use_authentication:
             if function_name in self.authorized_functions:
                 try:
                     self._verify_token(session_id, parsed, token)
-                except InvalidAuthenticationToken:
-                    log.warning("API validation failed")
-                    request.setResponseCode(self.UNAUTHORIZED)
-                    request.finish()
+                except InvalidAuthenticationToken as err:
+                    log.error("API validation failed")
```
jobevers: Th InvalidAuthenticationToken error seems like something that would be expected to happen on occasion, so I think the log level here should be at the warning level.
Log statements at the error level should be reserved for things that went wrong and for things we need to fix in the code: https://github.com/lbryio/lbry/wiki/Log-Levels#error.  Logs are not for the user, they are for us.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -114,8 +130,10 @@ def expire_session():
         content = request.content.read()
         try:
             parsed = jsonrpclib.loads(content)
-        except ValueError:
-            return server.failure
+        except ValueError as err:
+            log.error("Unable to decode request json")
```
jobevers: An error here isn't the fault of the daemon so I think this should be a log.warning.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -19,6 +20,12 @@ def default_decimal(obj):
         return float(obj)
 
 
+class JSONRPCException(Exception):
+    def __init__(self, err, code):
+        self.faultCode = code
+        self.faultString = err.getTraceback()
```
jobevers: better would be to keep the err: `self.err = err` and then have
```
@property
def fault_string(self):
    return err.getTraceback()
```

Generally, the dumber a constructor is the better. They should almost always be simple assignments of the passed in args to variables.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -90,7 +97,16 @@ def __init__(self, use_authentication=settings.use_auth_http):
     def setup(self):
         return NotImplementedError()
 
+    def _render_error(self, request, failure, version=jsonrpclib.VERSION_1, response_code=FAILURE):
```
jobevers: nit: long line

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -90,7 +97,16 @@ def __init__(self, use_authentication=settings.use_auth_http):
...
+        fault = jsonrpclib.dumps(JSONRPCException(Failure(failure), response_code), version=version)
```
jobevers: nit: long line

Also, why is `Failure(failure)` necessary as compared to only `failure`

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -125,36 +143,35 @@ def expire_session():
...
             if function_name in self.authorized_functions:
                 try:
                     self._verify_token(session_id, parsed, token)
-                except InvalidAuthenticationToken:
-                    log.warning("API validation failed")
-                    request.setResponseCode(self.UNAUTHORIZED)
-                    request.finish()
+                except InvalidAuthenticationToken as err:
+                    log.error("API validation failed")
+                    self._render_error(request, err, version, response_code=AuthJSONRPCServer.UNAUTHORIZED)
```
jobevers: nit: long line

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -125,36 +143,35 @@ def expire_session():
...
                 self._update_session_secret(session_id)
                 reply_with_next_secret = True
 
         try:
             function = self._get_jsonrpc_method(function_name)
-        except Exception:
-            log.warning("Unknown method: %s", function_name)
-            return server.failure
+        except AttributeError as err:
+            log.error("Unknown method: %s", function_name)
```
jobevers: is an error here the fault of the caller?  if so, probably want to change this to a warning

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -125,36 +143,35 @@ def expire_session():
...
+
         # cancel the response if the connection is broken
-        notify_finish = request.notifyFinish()
         notify_finish.addErrback(self._response_failed, d)
-        d.addErrback(self._errback_render, id)
-        d.addCallback(self._callback_render, request, id, version, reply_with_next_secret)
-        d.addErrback(notify_finish.errback)
-
+        d.addCallback(self._callback_render, request, version, reply_with_next_secret)
+        d.addErrback(lambda err: self._render_error(request, err, version))
```
jobevers: A `log_support.failure` call would be good to add to the errback as well.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -209,7 +226,7 @@ def _check_function_path(self, function_path):
         return True
 
     def _get_jsonrpc_method(self, function_path):
-        assert self._check_function_path(function_path)
+        assert self._check_function_path(function_path), AttributeError(function_path)
```
jobevers: this feels like an abuse of an assert.  More readable seems:

```
if not self._check_function_path(function_path):
    raise AttributeError(function_path)
```

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -255,20 +272,8 @@ def _callback_render(self, result, request, id, version, auth_required=False):
...
-
-    def _errback_render(self, failure, id):
-        log.error("Request failed:")
-        log.error(failure)
-        log.error(failure.value)
-        log.error(id)
-        if isinstance(failure.value, jsonrpclib.Fault):
-            return failure.value
-        return server.failure
+        except Exception as err:
```
jobevers: add a `log.exception('some message')` here

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -242,9 +259,9 @@ def _run_subhandlers(self, request):
                 assert handler(request)
```
jobevers: this also seems like an abuse of assert; asserts should be used primarily to check pre and post conditions in a function.

A problem here is that the broad try/catch will catch the assertion error along with
all of the other errors that might happen, which we than re-raise as a SubhandlerError
and possibly lose the context of the original error.

I think this would be better to get ride of the try/catch and have something like:

```
if not handler(request):
    raise SubhandlerError('Failed to process request', request)
```

jobevers:
Regarding 
image

Its not a big deal, but generally I'm a fan of rebasing over merge commits for bringing in changes from master.  Its definitely cleaner and as long as nobody else has started working off of your feature branch its safe.

lbrynet/lbrynet_daemon/auth/server.py
jobevers: log.exception already includes the traceback so `err.message` isn't necessary.  Its better to instead add some more context, like: `log.exception('failed to render results: %s', result)`

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -100,7 +101,11 @@ def __init__(self, use_authentication=settings.use_auth_http):
     def setup(self):
         return NotImplementedError()
 
-    def _render_error(self, request, failure, version=jsonrpclib.VERSION_1, response_code=FAILURE):
+    def _render_error(self, failure, request, version=jsonrpclib.VERSION_1, response_code=FAILURE,
```
jobevers: http://www.informit.com/articles/article.aspx?p=1392524:
```
Boolean arguments loudly declare that the function does more than one thing. They are confusing and should be eliminated.
```

Its generally better to refactor functions with boolean flags into new functions.  For example

```
def _log_and_render_error(self, failure, request, msg, **kwargs):
    log_support.failure(failure, log, msg)
    self._render_error(failure, request, **kwargs)
```

would be cleaner as it leaves `_render_error` simple, more readable and only doing what its name implies it does.

lyoshenka:
@jackrobison @jobevers whats the next step here?

jobevers:
There is one outstanding comment to be addressed, after that it can be merged in

Issue #261, opened 2016-11-10
Can't download theinternetsownboy


jobevers:
Riley just published theinternetsownboy and I can see the metadata, but am not able to download the file.

It should be up on reflector, need to investigate why its not.

jobevers:
reflector issues have been fixed

PR #262, opened 2016-11-10
Remove repeated code for reflector

jobevers:
Wrote this little script while I was investigating this: https://github.com/lbryio/lbry/issues/261

It generates a random blob of data and sends it to reflector; if there is an error a message can be sent to a webhook in slack.

### DIFF
--- a/lbrynet/core/HashBlob.py
+++ b/lbrynet/core/HashBlob.py
@@ -341,7 +341,6 @@ class HashBlobCreator(object):
         else:
             self.blob_hash = self.hashsum.hexdigest()
         d = self._close()
-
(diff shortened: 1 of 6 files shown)


jobevers:
No hurry on this review.

jobevers:
Added a second check that attempts to download the blob it just uploaded.  Unfortunately, that check frequently fails.

jobevers:
I'm going to move the actual script to a different repo, but some of the cleanup and re-org is sticking around.  PTAL.

jobevers:
Merging as its only some very basic re-arranging.

PR #268, opened 2016-11-11
add ui bundle to RC too

jobevers:


jobevers:
No need for travis to check the second commit as its windows only.

PR #269, opened 2016-11-11
Normalize python version

jobevers:
setuptools will force a version to be 'normalized'
which means that it removes the dash.

This will probably cause some weirdness on the version
numbers in the windows build; might have to re-work
my patches on cx_freeze: https://bitbucket.org/jobevers/cx_freeze/commits/all


jobevers:
osx builds are so far behind its dumb, merging as windows and linux passed

PR #272, opened 2016-11-15
Give priority to arguments over settings in UIManager.setup()

alexliebowitz:
Currently, certain options passed in through the API, like `check_requirements: true`, will sometimes be overridden by settings. This PR fixes that.

### DIFF
--- a/lbrynet/lbrynet_daemon/UIManager.py
+++ b/lbrynet/lbrynet_daemon/UIManager.py
@@ -61,10 +61,9 @@ class UIManager(object):
                 self.loaded_requirements = None
 
     def setup(self, branch=None, check_requirements=None, user_specified=None):
-        local_ui_path = (user_specified if user_specified is not None
-                         else settings.local_ui_path)
-        self.branch = branch if branch is not None else settings.ui_branch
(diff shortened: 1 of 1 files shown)


lbrynet/lbrynet_daemon/UIManager.py
```diff
@@ -61,9 +61,12 @@ def __init__(self, root):
                 self.loaded_requirements = None
 
     def setup(self, branch=None, check_requirements=None, user_specified=None):
-        local_ui_path = settings.local_ui_path or user_specified
-        self.branch = settings.ui_branch or branch
-        self.check_requirements = settings.check_ui_requirements or check_requirements
+        local_ui_path = (user_specified if user_specified is not None
```
jobevers: I prefer the succinctness of:

`local_ui_path = user_specified or settings.local_ui_path`

And its a common enough pattern for dealing with optional arguments that it is also more readable.

alexliebowitz: Fair enough. Fixed. I retained the "is not None" style for user_specified because it's a boolean.

PR #273, opened 2016-11-16
set build type in travis

lyoshenka:
Travis built this successfully, but I don't think that I can test whether it worked or not without getting the finished .deb. Can you please merge this before the nightlys get built? I'll check this first thing tomorrow morning and revert it if anything's broken.


lyoshenka:
fixed

jobevers:
LGTM

jobevers:
Oh, right, damn:  GNU sed and BSD sed are different.  BSD sed doesn't like an empty `-i` flag.

lyoshenka:
:unamused:

PR #276, opened 2016-11-16
Better parse environment variables for configuration

jobevers:


setup.py
```diff
@@ -47,7 +47,8 @@
     'base58',
     'googlefinance',
     'requests_futures',
-    'PyYAML'
+    'PyYAML',
+    'envparse'
```
jackrobison: You've probably got to add this to the [windows list](https://github.com/lbryio/lbry/blob/838508f342d96bac0e3a3b7458c12b7bbac10963/setup.py#L244-L244) as well

lbrynet/conf.py
```diff
@@ -137,56 +75,118 @@ def update(self, other):
...
+    pointtrader_server=(str, 'http://127.0.0.1:2424'),
+    reflector_servers=(list, [("reflector.lbry.io", 5566)], server_port),
+    wallet=(str, LBRYUM_WALLET),
+    ui_branch=(str, "master"),
+    default_ui_branch=(str, 'master'),
+    data_dir=(str, default_data_dir),
+    lbryum_wallet_dir=(str, default_lbryum_dir),
+    use_auth_http=(bool, False),
+    sd_download_timeout=(int, 3),
+    # TODO: document what the 'address' field is used for
```
jackrobison: [It's because the max key fee gets converted into LBC by the exchange rate manager, and beforehand it gets passed through the Fee validator. ](https://github.com/lbryio/lbry/blob/76fb7697fff8dedc853312276d72f8c92ce33562/lbrynet/lbrynet_daemon/ExchangeRateManager.py#L161-L161) Perhaps the thing to do is refactor the exchange rate conversion in a later pull request to take a currency and an amount rather than a "real" fee dict.

setup.py
jobevers: thanks!

PR #278, opened 2016-11-17
fix uncaught analytics exception

jackrobison:


lbrynet/analytics/api.py
```diff
@@ -16,6 +16,8 @@ def log_response(fn):
     def _log(future):
         if future.cancelled():
             log.warning('Request was unexpectedly cancelled')
+        elif future.exception():
+            log.warning(future.exception_info())
```
jobevers: This would be slightly better as

```
exc, traceback = future.exception_info()
log.warning('Failed to send an analytics event', exc_info=(type(exc), exc, traceback))
```

jobevers:
one small suggestion, otherwise LGTM

PR #279, opened 2016-11-17
Force auto connect to be on for lbryum

kaykurokawa:
This will force auto connect to be on for lbryum when using lbry. 

Auto connect allows lbryum wallet to connect to a functioning server if the default is down, and reconnect if a connection dies. 

This should fix lbryum wallet startup being slow in some instances. 

Goes with merged lbryum patch https://github.com/lbryio/lbryum/pull/52

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -1125,6 +1125,7 @@ class LBRYumWallet(Wallet):
 
         def setup_network():
             self.config = make_config(self._config)
+            self.config.set_key('auto_connect',True)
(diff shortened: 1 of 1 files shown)


jobevers:
this should already be set: https://github.com/lbryio/lbry/blob/e03db79f79fdf4c20a960485b15cdf468af4dba8/lbrynet/lbrynet_daemon/Daemon.py#L801

kaykurokawa:
Ah, that has a spelling mistake, should be auto_connect instead of auto-connect. Will close and make new pull request that fixes that mistake.

PR #281, opened 2016-11-18
WIP: Lbry in a box

jobevers:


jobevers:
nagging nancy - I'm still working on this, but thanks for the reminder.

jobevers:
I think most of these are already in or are part of another PR

PR #285, opened 2016-11-22
Allow 0.0.0.0 for api interface

jobevers:
For a host to be able to access the daemon running inside a docker
container the damon needs to be listening to 0.0.0.0

### DIFF
--- a/lbrynet/lbrynet_daemon/auth/server.py
+++ b/lbrynet/lbrynet_daemon/auth/server.py
@@ -1,4 +1,6 @@
 import logging
+import urlparse
+
 from decimal import Decimal
 from zope.interface import implements
 from twisted.web import server, resource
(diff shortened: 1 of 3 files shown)


lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -189,16 +190,39 @@ def _render_message(self, request, message):
...
+
+    def _check_origin(self, request):
         origin = request.getHeader("Origin")
-        referer = request.getHeader("Referer")
-        if origin not in [None, settings.ORIGIN]:
-            log.warning("Attempted api call from %s", origin)
+        if not self._check_source_of_request(origin):
+            log.warning("Attempted api call from invalid origin: %s", origin)
             return False
-        if referer is not None and not referer.startswith(settings.REFERER):
```
jackrobison: The `startswith` was necessary to let the UI work, where the referrer isn't strictly equal to the interface.

jackrobison:
Also, I think we should default auth to being on if localhost isn't the interface.

jobevers:
I am not worried about setting auth to being on. I think that if somebody overrides the origin, they know what they are doing. You are right though, this has security concerns - this shouldn't be a setting that we advertise or document. This should be mostly used to facilitate testing.  I'll add a comment around the setting as a warning to anybody that looks over the code.

lbrynet/lbrynet_daemon/auth/server.py
jobevers: okay.  I'll take a closer look at some of the common referrers and fix this up.

BTW - this is a good example of where tests are useful to prevent a regression.

PR #286, opened 2016-11-22
Allow Access-Control-Allow-Origin to be configurable

jobevers:
Being able to make requests from a different port then 5279 is useful - for example it enables somebody to setup a different server for the UI, say one that supports hot-reloading, and still access the jsonrpc.


jobevers:
There are two pull-requests that this is already rebased off of, so those should go in before this one.

jobevers:
replaced by https://github.com/lbryio/lbry/pull/310

PR #287, opened 2016-11-24
get_lbry_session_info function to retrieve lbry id, blob, and file count information

alexliebowitz:
Reports the same shortened, base58 encoded version included in logs


jackrobison:
Why's this needed?

jackrobison:
@alexliebowitz I moved the id info from `version` (as well as some blob/file count info) to a `get_lbry_session_info` function.

PR #290, opened 2016-11-28
Converting some log.info to log.debug in Wallet

kaykurokawa:
Downgrading log.info's that may contain user's personal information to log.debug's in Wallet


jobevers:
LGTM

PR #298, opened 2016-11-30
Allow environment settings to be overridden

jobevers:
Stores an overridden setting in the __dict__ and references
that first before pulling out the value in the environment.
This allows, for example, setting the UI on the cmd line.

This should address part of https://app.asana.com/0/198587529699250/220821321387630

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -22,6 +22,13 @@ KB = 2**10
 MB = 2**20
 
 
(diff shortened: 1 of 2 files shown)


jobevers:
@alexliebowitz I think this is related to the bug on the settings page.

alexliebowitz:
Superseded by #303

PR #300, opened 2016-12-01
Missing the return of a defer.fail when claim fails

kaykurokawa:
Fixing the missing return of a defer.fail when claim fails


jobevers:
can you change that to a `raise Exception(msg)`, other than that, LGTM

kaykurokawa:
changed to raise Exception(msg) , rebased and merging

PR #306, opened 2016-12-02
WIP: Remove 'success' and 'reason' as outputs in claim commands

kaykurokawa:
Remove 'success' and 'reason' as outputs in Daemon claim commands, as we throw an exception instead for failed claims instead. 

Have support and abandon claim commands do the same.


jobevers:
I think these can be tested by creating a MockWallet class that inherits from `Wallet` and has `_save_name_metadata`, `_abandon_claim`, `_support_claim` methods.

kaykurokawa:
will do this under new file test/unit/core/unit_Wallet.py

PR #308, opened 2016-12-05
WIP: Fix wallet initialization 

kaykurokawa:
7fb42e1 changes the order of how the wallet is loaded. Before it would load the network, load the disk wallet, and then catch up on the blockchain (if any steps failed, the subsequent steps would not be performed) . If loading the network failed, the wallet would never be loaded (we can still query the wallet for balance, and transactions even if not connected to the network). 

Now we load the disk wallet first, load the network, and than catch up on the blockchain so that the wallet is always loaded regardless of if the network loaded. 

dafb8f1 fixes cmd_runner so that it would always get loaded upon call. This patch is also needed for lbryum password entry for encryption.

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -1142,6 +1142,7 @@ class LBRYumWallet(Wallet):
         self.catchup_progress = 0
         self.max_behind = 0
 
+    # load disk wallet , setup network, and then check to see if caught up 
     def _start(self):
 
         network_start_d = defer.Deferred()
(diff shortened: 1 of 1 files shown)


jobevers:
LGTM.  Since this is a fork, can you make sure that tests / linting pass.

kaykurokawa:
This will need to be redone since https://github.com/lbryio/lbry/pull/331 changes a lot of the wallet code

kaykurokawa:
Closing as replaced by new pull request

PR #312, opened 2016-12-05
local_ui_path needs to be a str

jobevers:


jackrobison:
Was there an error otherwise?

jobevers:
I wasn't able to override it without this change.

jackrobison:
OK, LGTM

PR #317, opened 2016-12-06
Remove lbrynet-console

jackrobison:
This removes lbrynet-console, which hasn't been used in a good while. The only file that does see use - Settings.py - was moved into lbrynet/lbrynet_daemon. It's probably worth refactoring Settings.py to use sqlite like the rest of the databases (it's a unqlite holdout).


jobevers:
LGTM

PR #323, opened 2016-12-10
Immediately announce completed blobs

jobevers:
It doesn't seem necessary to wait up to 60 seconds for a new blob to
be announced to the dht. Immediately announce it and schedule the next
announce time as usual.

This won't fix any errors in reflector, but still seems like a good improvement to put into place.

### DIFF
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -120,6 +120,7 @@ class DiskBlobManager(BlobManager):
         return self.blob_creator_type(self, self.blob_dir)
 
     def _make_new_blob(self, blob_hash, upload_allowed, length=None):
+        log.debug('Making a new blob for %s', blob_hash)
         blob = self.blob_type(self.blob_dir, blob_hash, upload_allowed, length)
         self.blobs[blob_hash] = blob
         d = self._completed_blobs([blob_hash])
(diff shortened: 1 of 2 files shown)


jackrobison:
LGTM

lbrynet/core/BlobManager.py
```diff
@@ -143,9 +144,11 @@ def set_length(length):
 
     def blob_completed(self, blob, next_announce_time=None):
         if next_announce_time is None:
-            next_announce_time = time.time()
-        return self._add_completed_blob(blob.blob_hash, blob.length,
-                                        time.time(), next_announce_time)
+            next_announce_time = time.time() + self.hash_reannounce_time
+        d = self._add_completed_blob(blob.blob_hash, blob.length,
+                                     time.time(), next_announce_time)
+        d.addCallback(lambda _: self.hash_announcer.immediate_announce([blob.blob_hash]))
```
jackrobison: This'll announce it twice: right after _add_completed_blob and with the next tick of `DHTHashAnnouncer.run_manage_loop` (after _add_completed_blob is done)

jobevers: I set next_announce_time to be the hour in a future so that avoids the run_manage_loop one.

jackrobison: Disregard the above, the way you've got it does seem preferable - that way you're not waiting for the next tick of the hash announcer.

jobevers:

PR #324, opened 2016-12-13
sort the help functions

kauffj:
This returns a sorted list of functions when the `help` API method is called.

(Mainly because 1) it should and 2) I wanted to test a full code change / test cycle)

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -1353,7 +1353,7 @@ class Daemon(AuthJSONRPCServer):
         """
 
         if not p:
-            return self._render_response(self.callable_methods.keys(), OK_CODE)
+            return self._render_response(sorted(self.callable_methods.keys()), OK_CODE)
         elif 'callable_during_start' in p.keys():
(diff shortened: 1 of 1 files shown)


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2717,4 +2717,4 @@ def run_reflector_factory(factory):
     d = reactor.resolve(reflector_address)
     d.addCallback(lambda ip: reactor.connectTCP(ip, reflector_port, factory))
     d.addCallback(lambda _: factory.finished_deferred)
-    return d
+    return d
```
jobevers: does your editor remove newlines 😢

jobevers:
LGTM

jobevers:
Yes please: http://stackoverflow.com/a/729795

lbrynet/lbrynet_daemon/Daemon.py
jobevers: Yes please: http://stackoverflow.com/a/729795

jobevers:

PR #326, opened 2016-12-14
Make a new logger with a fail method that is useful for errbacks

jobevers:
Extracting useful tracebacks and line numbers from failures
within twisted's deferred can be a pain. Hopefully this is a step
in the right direction.


jobevers:

PR #328, opened 2016-12-14
rename platform to system_info

jobevers:
There is a system module named `platform`, which can conflict with `Platform`


jackrobison:
LGTM

jobevers:

PR #330, opened 2016-12-14
remove unused check_consistency function

jobevers:
Nothing seems to use this so burninate.

image


jobevers:

PR #331, opened 2016-12-14
Fix wallet concurrency issue 

kaykurokawa:
This patch fixes wallet concurrency issue happening due to the fact that lbryum is not designed  to be thread safe, but lbrynet was making asynchronous threaded calls to lbryum by using threads.deferToThread() .

This patch requires https://github.com/lbryio/lbryum/pull/59 from lbryum. 

https://github.com/lbryio/lbry/commit/84daa6b4885a99df5b09aae1e8753b986f98d77d
First we remove unused get_transaction function from both Daemon and Wallet. This function was not being used anywhere (not used in lbry-web-ui either), has concurrency issues, and was not even decoding the transaction in a useful manner so I just decided to remove it instead of  fixing it. 

https://github.com/lbryio/lbry/commit/9125adbba9c1160bab6cacbb61beb723ed4467b5
Use defer.succeed( ) instead of threads.deferToThread() when using lbryum. The only exception is when we are using lbryum functions that are just making network calls to lbryum-sever, these can be called in a separate thread asynchronously so we use threads.deferToThread() here.

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -17,7 +17,6 @@ from lbryum import SimpleConfig, Network
 from lbryum.lbrycrd import COIN
 import lbryum.wallet
 from lbryum.commands import known_commands, Commands
-from lbryum.transaction import Transaction
 
 from lbrynet.interfaces import IRequestCreator, IQueryHandlerFactory, IQueryHandler, IWallet
 from lbrynet.core.client.ClientRequest import ClientRequest
@@ -500,11 +499,6 @@ class Wallet(object):
     def support_claim(self, name, claim_id, amount):
         return self._support_claim(name, claim_id, amount)
 
-    def get_tx(self, txid):
-        d = self._get_raw_tx(txid)
-        d.addCallback(self._get_decoded_tx)
-        return d
-
     def get_block_info(self, height):
         d = self._get_blockhash(height)
         return d
@@ -715,13 +709,7 @@ class Wallet(object):
     def _check_first_run(self):
         return defer.fail(NotImplementedError())
 
-    def _get_raw_tx(self, txid):
-        return defer.fail(NotImplementedError())
-
     def _send_name_claim(self, name, val, amount):
-        return defer.fail(NotImplementedError())
-
-    def _get_decoded_tx(self, raw_tx):
         return defer.fail(NotImplementedError())
 
     def _abandon_claim(self, claim_outpoint):
@@ -749,9 +737,6 @@ class Wallet(object):
         return defer.fail(NotImplementedError())
 
     def _address_is_mine(self, address):
-        return defer.fail(NotImplementedError())
-
-    def _get_transaction(self, txid):
         return defer.fail(NotImplementedError())
 
     def _start(self):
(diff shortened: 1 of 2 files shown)


lbrynet/core/Wallet.py
```diff
@@ -908,162 +892,149 @@ def log_error(err):
...
 
-    def get_balance(self):
-        cmd = known_commands['getbalance']
+    # run commands as a defer.succeed,
+    # lbryum commands should be run this way , unless if the command
+    # only makes a lbrum server query, use _run_cmd_as_defer_to_thread()
+    def _run_cmd_as_defer_succeed(self, command_name, *args):
+        cmd = known_commands[command_name]
         func = getattr(self.cmd_runner, cmd.name)
+        return defer.succeed(func(*args))
```
jobevers: 👍

lbrynet/core/Wallet.py
```diff
@@ -908,162 +892,149 @@ def log_error(err):
...
 
+    def _broadcast_claim_transaction(self, claim_out):
+        if 'success' not in claim_out:
+            raise Exception('Unexpected claim command output:{}'.format(claim_out))
+        if claim_out['success']:
+            d = self._broadcast_transaction(claim_out['tx'])
+            d.addCallback(lambda _: claim_out)
+            return d
+        else:
+            return claim_out
```
jobevers: have this be `defer.succeed(claim_out)` so that its always returning a deferred

lbrynet/core/Wallet.py
```diff
@@ -908,162 +892,149 @@ def log_error(err):
...
-        return d
+        header = self.network.blockchain.read_header(height)
+        return defer.succeed(self.network.blockchain.hash_header(header))
 
     def _get_blockhash(self, height):
-        d = threads.deferToThread(self.network.blockchain.read_header, height)
-        d.addCallback(lambda header: self.network.blockchain.hash_header(header))
-        return d
+        header = self.network.blockchain.read_header(height)
+        return defer.succeed(self.network.blockchain.hash_header(header))
```
jobevers: Is `network.blockchain` actually making a network call?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1886,22 +1886,6 @@ def jsonrpc_get_transaction_history(self):
         d.addCallback(lambda r: self._render_response(r, OK_CODE))
         return d
 
-    def jsonrpc_get_transaction(self, p):
```
jobevers: 👍  - I'm a fan of getting rid of unused code.

jobevers:
I installed this, and the lbryum fix on a test reflector instance and was able to connect and request 50 blobs simultaneously.

jobevers:
Had the one question about `network.blockchain` but if that is a local call and not a network one, this LGTM.

lbrynet/core/Wallet.py
kaykurokawa: good catch

lbrynet/core/Wallet.py
kaykurokawa: no the network.blockchain object uses local storage

jobevers:
It would also be good to rebase with origin/master just to make sure that the build is working.

kaykurokawa:
rebased to master and fixed the broadcast_transaction bug

PR #335, opened 2016-12-14
Add TRACE level logging

jobevers:
Note that there currently isn't any command line flags or settings that would allow you to turn on TRACE logging


jackrobison:
LGTM

jobevers:

PR #336, opened 2016-12-14
Add timing to hash announcements

jobevers:
This could potentially be a performance issue
on reflector or any daemon with a large number of blobs.

### DIFF
--- a/lbrynet/core/server/DHTHashAnnouncer.py
+++ b/lbrynet/core/server/DHTHashAnnouncer.py
@@ -1,6 +1,7 @@
 import binascii
 import collections
 import logging
+import time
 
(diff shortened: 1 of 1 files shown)


lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -1,6 +1,7 @@
 import binascii
 import collections
 import logging
+import time
```
jackrobison: should be imported out of`lbrynet.core.utils`

jobevers: will make this a separate PR to fix this in multiple files.

jobevers:

PR #337, opened 2016-12-14
Improve BlobAvailabilityTracker performance

jobevers:
For daemons with a lot of blobs, getting mean availabity
will be slow. Samples the blobs in an attempt at getting
better performance.

### DIFF
--- a/lbrynet/core/BlobAvailability.py
+++ b/lbrynet/core/BlobAvailability.py
@@ -1,4 +1,6 @@
 import logging
+import random
+import time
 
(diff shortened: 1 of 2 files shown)


lbrynet/core/BlobAvailability.py
```diff
@@ -1,4 +1,6 @@
 import logging
+import random
+import time
```
jackrobison: these should be imported out of `lbrynet.core.utils` to make testing easier

jobevers:

PR #339, opened 2016-12-15
All The PRs

jobevers:
Instead of having a lot of separate PRs, this is all of them together - with the one that fixes the build first.

I'll leave the other ones up in case that is easier for feedback, but this is the one that should be merged.


lbrynet/core/BlobAvailability.py
```diff
@@ -1,4 +1,6 @@
 import logging
+import random
+import time
```
jackrobison: These should probably be imported from `lbrynet.core.utils utils for easier testing

jobevers: Indeed.  Thanks

PR #347, opened 2016-12-16
Log failure when attempting to get_lbry_files

jobevers:
I had an old version of lbryum installed and so no
names were coming back. But the daemon was failing
silently. This at least logs the error.


jackrobison:
LGTM

PR #348, opened 2016-12-16
remove unqlite requirement and old Settings.py

jackrobison:


jobevers:
Can you pull apart the first commit into separate ones?

This will be some good rebase practice :-)  While you're at it, you can get rid of the merge commit and make sure that unqlite & cython stick around for the tests.

jobevers:
I have to refresh my memory on the DHT, but I'm worried that if the LBRY id changes there will be bunch of stale DHT entries.

jackrobison:
@jobevers 

Stale DHT entries get rotated out as fresher ones come in. The ids are really supposed to be randomized at each start, [and ideally even be authenticatable](https://www.cl.cam.ac.uk/~lw525/publications/security.pdf). This allows more 'churn' in the routing table, and mixes up which peers are closest to the desired blobs.

PR #358, opened 2016-12-20
Fix wallet initialization 

kaykurokawa:
Redo of https://github.com/lbryio/lbry/pull/308

https://github.com/lbryio/lbry/commit/7d02e7d37065e3f55efcee49353efd08adb0907a changes the order of how the wallet is loaded. Before it would load the network, load the disk wallet, and then catch up on the blockchain (if any steps failed, the subsequent steps would not be performed) . If loading the network failed, the wallet would never be loaded (we can still query the wallet for balance, and transactions even if not connected to the network).

Now we load the disk wallet first, load the network, and than catch up on the blockchain so that the wallet is always loaded regardless of if the network loaded.

https://github.com/lbryio/lbry/commit/dafb8f1cf8db928a81c49f8971bd316608276f8d fixes cmd_runner so that the Commands object would always get loaded upon call. This patch is also needed for lbryum password entry for encryption.

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -771,7 +771,6 @@ class LBRYumWallet(Wallet):
         self._config = config
         self.network = None
         self.wallet = None
-        self.cmd_runner = None
         self.first_run = False
         self.printed_retrieving_headers = False
         self._start_check = None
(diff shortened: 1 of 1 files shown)


jobevers:
LGTM

PR #360, opened 2016-12-20
Only store the offer explicitly when its sent

jobevers:
Ran into a situation where I was receiving an offer response without
something in protocol_offers. I think this was because
get_and_set_rate doesn't always populate that dictionary.

I'm slightly worried that there still might be issues if the connection is
unreliable: if a response comes in twice it will error or
if a second offer is made without a response to the first
there will be an error.


jackrobison:
LGTM

PR #361, opened 2016-12-21
Analytics on startup

jobevers:
Two Goals here
* add lbryid and session_id to the startup event (https://github.com/lbryio/lbry/commit/3acd1de14198262ac086aea88e92ca3f3f664736)
* actually send the startup event (https://github.com/lbryio/lbry/commit/2a43b3433796ceae23c065cabe08f26d195db473)

There are a bunch of refactorings mixed in here, but they live in commits on there own, so this probably best reviewed on commit at a time.

### DIFF
--- a/lbrynet/analytics/api.py
+++ b/lbrynet/analytics/api.py
@@ -5,7 +5,7 @@ import logging
 from requests import auth
 from requests_futures import sessions
 
-from lbrynet.conf import settings
+from lbrynet import conf
 from lbrynet.analytics import utils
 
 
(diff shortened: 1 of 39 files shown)


jobevers:
Due to the size of conf.settings refactoring, I'd like to test the results of the builds before merging this into master.

jackrobison:
LGTM

jackrobison:
Is [this](https://github.com/lbryio/lbry/blob/20ca6edef46ac935a5e10f5460838e9d11befd4a/lbrynet/lbrynet_daemon/Daemon.py#L294) still used?

PR #364, opened 2016-12-28
Fix bug in server protocol

kaykurokawa:
This patch 
https://github.com/lbryio/lbry/commit/09846413bdf0aa5c647e8ca4e8198d748b4e6ff8 introduced a problem where BlobRequestHandler would never be loaded by ServerProtocol , because self.factory.query_handler_factories loaded by the Daemon ( see _add_query_handler() )was changed to be a dictionary of {query_id: query_handler_factory,} where before it was {query_handler_factory:enabled,}. This will cause the download to fail. 

 This patch changes ServerProtocol but I'm not sure if you want to change ServerProtocol or the Daemon function _add_query_handler() .

### DIFF
--- a/lbrynet/core/server/ServerProtocol.py
+++ b/lbrynet/core/server/ServerProtocol.py
@@ -33,10 +33,9 @@ class ServerProtocol(Protocol):
         peer_info = self.transport.getPeer()
         self.peer = self.factory.peer_manager.get_peer(peer_info.host, peer_info.port)
         self.request_handler = ServerRequestHandler(self)
-        for query_handler_factory, enabled in self.factory.query_handler_factories.iteritems():
-            if enabled is True:
-                query_handler = query_handler_factory.build_query_handler()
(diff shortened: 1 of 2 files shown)


jobevers:
LGTM

kaykurokawa:
the test is failing because it is setting up the query_handlers differently https://github.com/lbryio/lbry/blob/81187da5b7a6707c17a55f4dd736c3b75bcfab2d/tests/functional/test_misc.py#L149 from how the daemon sets it up

jobevers:
I think its worth changing the tests to match the new data structure.

kaykurokawa:
since this is a fix on jack's pull request, he should decide on what to do here, whether to keep the old way for setting up the query handlers or not.

jobevers:
On a call with Jack, he's good with this. I fixed the tests because I was running into problems in some code I was working on and needed to grab this fix.

jackrobison:
👍 +1

PR #365, opened 2016-12-30
Fix typo in set_settings API method

alexliebowitz:
Was causing it to throw errors


jobevers:
LGTM.  This is the kind of thing pylint should catch but I've disabled the no-member error because it unfortunately reports a lot of false positives with twisted.

PR #366, opened 2016-12-30
Add analytics event for download started / errored / finished

jobevers:
Also, way too much refactoring and switching to inlineCallbacks


jobevers:
closing in favor of #367

PR #369, opened 2016-12-30
track errors in analytics

jobevers:


jobevers:
merging so I can get a pre-release out

PR #370, opened 2016-12-31
Stop eating wallet connection error

jobevers:
In general, unless we have a way to actually handle an error, its better to shutdown then to leave the app in a weird state.


kaykurokawa:
tested and works LGTM

PR #372, opened 2017-01-02
fix the stop script

jobevers:


lbrynet/lbrynet_daemon/DaemonControl.py
```diff
@@ -21,20 +21,17 @@ def test_internet_connection():
...
 
 
+@defer.inlineCallbacks
 def stop():
-    def _disp_shutdown():
-        print "Shutting down lbrynet-daemon from command line"
+    conf.initialize_settings()
+    log_support.configure_console()
+    try:
+        yield JSONRPCProxy.from_url(conf.settings.API_CONNECTION_STRING).stop()
```
jackrobison: `yield  LBRYAPIClient.config().stop()`

jobevers:
PTAL.  It also turns out that the deferred wasn't necessary.

PR #373, opened 2017-01-02
Set ui_version for Bundled UI

jobevers:


lbrynet/lbrynet_daemon/UIManager.py
```diff
@@ -67,6 +67,10 @@ def setup(self, branch=None, check_requirements=None, user_specified=None):
         self.check_requirements = (check_requirements if check_requirements is not None
                                    else conf.settings.check_ui_requirements)
 
+        # Note that this currently overrides any manual setting of UI.
+        # It might be worth considering changing that behavior but the expectation
+        # is generally that any manual setting of the UI will happen during development
+        # and not for folks checking out the QA / RC builds that bundle the UI.
```
jackrobison: This confused Jeremy earlier today, it's probably worth having `configure_ui` indicate that it won't cooperate when it wont

jackrobison:
One small thing, other than that LGTM

PR #374, opened 2017-01-03
upgrade txJson-RCP to 0.5

jobevers:
The older version we were running didn't support jsonrpc v2 for
error codes, which both the command line client and ui client
were expecting.

Notably, this fixes this error:

```
  File "/home/jobevers/.virtualenvs/lbry/lib/python2.7/site-packages/jsonrpc/common.py", line 185, in get_result
    code = self.error['code']
KeyError: 'code'
```


jobevers:
this is only used for https://github.com/oubiwann/txjsonrpc/blob/master/txjsonrpc/jsonrpclib.py#L64, 
so its safe upgrade.

PR #379, opened 2017-01-03
open the ui as soon as possible

jackrobison:
this opens the ui after UIManager.setup rather than after the whole Daemon.setup chain

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -291,7 +291,7 @@ class Daemon(AuthJSONRPCServer):
         self.lbry_file_manager = None
 
     @defer.inlineCallbacks
-    def setup(self):
+    def setup(self, launch_ui):
         self._modify_loggly_formatter()
 
         def _announce_startup():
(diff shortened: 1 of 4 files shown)


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -329,7 +329,7 @@ def _announce():
 
         if conf.settings.host_ui:
             self.lbry_ui_manager.update_checker.start(1800, now=False)
-            yield self.lbry_ui_manager.setup()
+            yield self.lbry_ui_manager.setup(launch=launch_ui)
```
jobevers: I'd change this to be 
```
yield self.lbry_ui_manager.setup()
if launch_ui:
    self.lbry_ui_manager.launch()
```

lbrynet/lbrynet_daemon/UIManager.py
```diff
@@ -95,6 +94,8 @@ def setup(self, branch=None, check_requirements=None, user_specified=None):
 
         d = self._up_to_date()
         d.addCallback(lambda r: self._download_ui() if not r else self._load_ui())
+        if launch:
+            d.addCallback(lambda _: webbrowser.open(conf.settings.UI_ADDRESS))
```
jobevers: On line 75, 83, 89, the function returns which means launch won't happen regardless of the setting.

In general, functions should "do one thing" and now this function does two things.  It setups the UI and potentially launches it.

Also - I've harped on boolean flags before: http://www.informit.com/articles/article.aspx?p=1392524

I think it would be better to have this be two separate functions and call them individually.

jobevers:
LGTM once the changes are made.

PR #380, opened 2017-01-03
better lbrynet-cli connection error message

lyoshenka:
Actually lets you know what's going on with the daemon when it cannot connect

### DIFF
--- a/lbrynet/lbrynet_daemon/DaemonCLI.py
+++ b/lbrynet/lbrynet_daemon/DaemonCLI.py
@@ -4,6 +4,7 @@ import json
 from lbrynet import conf
 from lbrynet.lbrynet_daemon.auth.client import LBRYAPIClient
 from jsonrpc.common import RPCError
+from urllib2 import URLError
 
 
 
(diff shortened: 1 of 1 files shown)


lbrynet/lbrynet_daemon/DaemonCLI.py
```diff
@@ -43,16 +44,16 @@ def main():
...
+    except URLError:
+        print "Could not connect to lbrynet-daemon. Are you sure it's running?"
+        sys.exit(1)
+
+    if status.get('code', False) != "started":
+        print "Daemon is in the process of starting. Please try again in a bit."
+        message = status.get('message', False)
+        if message:
+            print "  Status: " + message
+        sys.exit(1)
```
jobevers: it might be possible that you'll fail because the daemon requires authentication and the CLI isn't providing any.  What is the message in that case?

jackrobison: might be worth adding in a check to see if a process bound to port 5279

lyoshenka: @jobevers fixed in https://github.com/lbryio/lbry/pull/397/commits/96252785d30a1f0ce078d1333508493488babae5

@jackrobison not a bad idea, but i didnt do it yet. maybe later.

lyoshenka:
Closing this, since https://github.com/lbryio/lbry/pull/397 includes these changes

PR #381, opened 2017-01-04
Fix watch

jobevers:
I changed the _download_name return type without realizing that
it was also used by the watch endpoint.  This switches the endpoint
to go through get so that watches can be tracked just like downloads


alexliebowitz:
LGTM

PR #382, opened 2017-01-04
silence "[Errno 11] Resource temporarily unavailable" error

lyoshenka:
not sure if good idea or bad idea

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ can and probably will change functionality and break backwards compatability
 at anytime.
 
 ## [Unreleased]
+### Fixed
+ * Change EWOULDBLOCK error in DHT to warning. #481
 
 ## [0.8.3rc0] - 2017-02-10
 ### Changed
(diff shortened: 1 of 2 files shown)


lyoshenka:
@kauffj this is the error we keep seeing

kauffj:
@lyoshenka do you understand root cause? If not, don't we need to know that before making a change here?

kaykurokawa:
From what I googled EWOULDBLOCK most likely means that the send buffer is full so it seems like turning down the chattiness of the DHT should make the '[Errno 11] Resource temporarily unavailable' error go away?

jobevers:
This bug is only 9 years old: https://twistedmatrix.com/trac/ticket/2790

And some good comments: http://stackoverflow.com/questions/7906615/udp-flow-control-with-python-twisted/7929158#7929158

I think the solution would be to reschedule the packet send and have the [delay](https://github.com/lbryio/lbry/blob/a8d45e4f1a080fd6939b922061df9fa9075cee41/lbrynet/dht/protocol.py#L51) dynamically adjust - slowing down after getting an EWOULDBLOCK.

lyoshenka:
I'm going to close this for now.

jobevers:
I'm re-opening this. In looking over the error reports, this is a huge source of unhandled errors.

Instead of straight up ignoring the error, I added a log.warning.

PR #383, opened 2017-01-04
Logging Improvements

jobevers:
Twisted does its own thing with logging, and without hooking into it we are left with the very useful "Unhandled error in deferred" message.

By hooking the twisted logging into the python logger, we'll actually get some useful error messages.

Twisted also logs when protocols start / stop, etc.  Which might be useful, might be noisy - can adjust the log level if its too noisy

### DIFF
--- a/lbrynet/core/log_support.py
+++ b/lbrynet/core/log_support.py
@@ -10,6 +10,7 @@ import traceback
 
 import requests
 from requests_futures.sessions import FuturesSession
+import twisted.python.log
 
 import lbrynet
 from lbrynet import analytics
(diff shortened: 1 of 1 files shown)


jobevers:
@jackrobison please review.

lbrynet/core/log_support.py
```diff
@@ -147,8 +149,14 @@ def get_loggly_url(token=None, version=None):
...
 
+def configure_loggly_handler(*args, **kwargs):
+    if build_type.BUILD == 'dev':
+        return
+    _configure_loggly_handler(*args, **kwargs)
+
+
 @_log_decorator
-def configure_loggly_handler(url=None, **kwargs):
+def _configure_loggly_handler(url=None, **kwargs):
```
lyoshenka: why did you make two functions?

jobevers: because this one has the `_log_decorator`, which helper for setting up handlers.  Maybe there are better names I could use for the two.

PR #386, opened 2017-01-05
upgrade twisted to 16.6

jobevers:


jackrobison:
not forgetting about this, but I'd like to hold off until simplification of requirements.txt is on master, unless it's specifically needed already.

jobevers:
sounds good; I'll reassign this back to me until its more appropriate

jobevers:
Yep, I'm still aware this is here

jobevers:
I'm going to close this until post-electron changes, as electron simplifies the build process for lbrynet.

PR #393, opened 2017-01-09
Fix obtaining version from github

kaykurokawa:
I created a tagged lbryum release on github, so now we can get version from there https://github.com/lbryio/lbryum/releases like we do with lbrynet instead of pulling it from the source code pages. 

Obtaining a response using requests can also hang the startup process indefinitely if the github server is slow to respond, or access to github is blocked, so add a conservative timeout.

Further improvement would be to use an asynchronous call to make requests.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -10,7 +10,6 @@ import base58
 import requests
 import urllib
 import simplejson as json
-from urllib2 import urlopen
 from decimal import Decimal
 
 from twisted.web import server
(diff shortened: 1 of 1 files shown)


jobevers:
An improvement, but I think we should move towards making all of our http requests async, either using request-futures, like the analytics library, or using twisted http://twistedmatrix.com/documents/current/web/howto/client.html.  Requests-futures seems simpler and has a nicer api.

kaykurokawa:
added pylint fixes, will rebase once passed. Will look into what's the best way to do this async

kaykurokawa:
rebase pylint fixes into single commit

jobevers:
LGTM.  Go ahead and merge - can do any async work in a separate PR.

PR #394, opened 2017-01-09
handle dropped api requests

jackrobison:
This cancels API requests that have been dropped/cancelled by a client before receiving a response.

https://app.asana.com/0/198587529699250/239832897034379

### DIFF
--- a/lbrynet/lbrynet_daemon/auth/server.py
+++ b/lbrynet/lbrynet_daemon/auth/server.py
@@ -4,7 +4,7 @@ import urlparse
 from decimal import Decimal
 from zope.interface import implements
 from twisted.web import server, resource
-from twisted.internet import defer, reactor
+from twisted.internet import defer
 from twisted.python.failure import Failure
 from twisted.internet.error import ConnectionDone, ConnectionLost
 from txjsonrpc import jsonrpclib
--- a/lbrynet/lbrynet_daemon/auth/server.py
+++ b/lbrynet/lbrynet_daemon/auth/server.py
@@ -217,15 +217,14 @@ class AuthJSONRPCServer(AuthorizedBase):
         else:
             d = defer.maybeDeferred(function, *args)
 
-        # finished_deferred will fire when the request is finished
-        # this could be because the request is really done, or because the connection dropped
-        # if the connection dropped, cancel the deferred stack
-        # otherwise it'll try writing to a closed request and twisted doesn't like that
+        # finished_deferred will callback when the request is finished and errback if something went wrong
+        # if the errback is called, cancel the deferred stack, this is to prevent request.finish()
+        # from being called on a closed request
 
-        # TODO: don't trap RuntimeError, which is presently done to handle deferredLists that
+        # TODO: don't trap RuntimeError, which is presently caught to handle deferredLists that
         # won't peacefully cancel, namely get_lbry_files
 
-        finished_deferred.addBoth(self._handle_dropped_request, d, function_name)
+        finished_deferred.addErrback(self._handle_dropped_request, d, function_name)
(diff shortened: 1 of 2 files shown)


lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -109,20 +122,38 @@ def __init__(self, use_authentication=None):
     def setup(self):
         return NotImplementedError()
 
+    def _set_headers(self, request, data, update_secret=False):
```
jobevers: why did you move this?

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -185,13 +217,23 @@ def expire_session():
...
-        notify_finish.addErrback(self._response_failed, d)
+        # finished_deferred will fire when the request is finished
+        # this could be because the request is really done, or because the connection dropped
+        # if the connection dropped, cancel the deferred stack
+        # otherwise it'll try writing to a closed request and twisted doesn't like that
+
+        # TODO: don't trap RuntimeError, which is presently done to handle deferredLists that
+        # won't peacefully cancel, namely get_lbry_files
+
+        finished_deferred.addBoth(self._handle_dropped_request, d, function_name)
```
jobevers: Given that `finished_deferred` is "A Deferred which will be triggered when the request is finished -- with a None value if the request finishes successfully or with an error if the request is interrupted by an error (for example, the client closing the connection prematurely)"

It seems worthwhile to treat the two cases differently.  Is there a situation in which finished_deferred will fire  a successful callback without `d` having been called?

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -109,20 +122,38 @@ def __init__(self, use_authentication=None):
...
         if response_code != AuthJSONRPCServer.FAILURE:
             request.setResponseCode(response_code)
-        request.write(fault)
-        request.finish()
+        self._render_message(request, fault)
+
+    def _handle_dropped_request(self, result, d, function_name):
+        if not d.called:
+            log.warning("Cancelling dropped api request %s", function_name)
+            reactor.callFromThread(d.cancel)
```
jobevers: why `callFromThread` instead of `calllater(d.cancel)` or simply`d.cancel()`?

lbrynet/lbrynet_daemon/auth/server.py
jackrobison: The addBoth should be (and has been updated to) addErrback.

>Is there a situation in which finished_deferred will fire a successful callback without d having been called?

There shouldn't be, d being called means request.finish() has been called. It seems like there's a  potential race condition if finished_deferred errbacks though: will d be cancelled before it calls request.finish()? If not, then there will be a RuntimeError instead of a ConnectionLos error. Race conditions aside, this also happens if d can't be successfully cancelled, as is the case with the deferredList returned by Daemon.get_lbry_files - but fixing that can go in a separate pr.

lbrynet/lbrynet_daemon/auth/server.py
jackrobison: Just because I'm using it in the render methods bellow it.

jobevers: I should have extended my comment - its useful for the code reviewer if simple housecleaning is in a commit on its own. As it is, its difficult to read through the diff and see whether `_set_headers` was only moved or if it was moved and changed.

jobevers:
Tested by clicking around a bunch really quick (which in the past would cause the error).  LGTM.

I'll look at fixing the merge conflicts.

jobevers:
Rebased as https://github.com/lbryio/lbry/pull/394; closing this.

PR #395, opened 2017-01-09
Fix Daemon RPC command publish output

kaykurokawa:
The Daemon RPC command "publish" did not return the claim ID of the claim and fee when going through Publisher class (Publisher class only returned txid and nout).  

It was also returning the claim Id in a dictionary with the key "claimid" which was inconsistent with other claim functions which were using the key "claim_id". This requires lbryum patch : https://github.com/lbryio/lbryum/pull/70

Added comments on what the claim_name() function does in Wallet.py

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -609,7 +609,21 @@ class Wallet(object):
             meta_for_return[k] = new_metadata[k]
         return defer.succeed(Metadata(meta_for_return))
 
+    """
+    Claim a name, update if name already claimed by user
+    @param name: name to claim
 
+    @param bid: bid amount
+
+    @param m: metadata
+
+    @return: Deferred which returns a dict containing below items
+        txid - txid of the resulting transaction
+        nout - nout of the resulting claim
+        fee - transaction fee paid to make claim
+        claim_id -  claim id of the claim
+
+    """
     def claim_name(self, name, bid, m):
         def _save_metadata(claim_out, metadata):
             if not claim_out['success']:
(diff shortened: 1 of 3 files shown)


jobevers:
Reassigning to grin as he's working on the API.

lbrynet/core/Wallet.py
```diff
@@ -609,7 +609,21 @@ def update_metadata(self, new_metadata, old_metadata):
...
+
+    @param m: metadata
+
+    @return: Deferred which returns a dict containing below items
+        txid - txid of the resulting transaction
+        nout - nout of the resulting claim
+        fee - transaction fee paid to make claim
+        claim_id -  claim id of the claim
+
+    """
```
jobevers: being pedantic, but I prefer the google style of docstrings, and most of lbrynet uses them.  

The google style is:

```Required short single line explanation.

Optional longer explanation

Args:
    name_of_argument: description

Returns: a description of what the function returns
```

More examples: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

kaykurokawa: Other functions in Wallet.py does not use the google style. I'll fix the commenting style of this and other function in Wallet.py  in a seperate pull request.

jobevers: Fair point.  Sounds good to me.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1686,7 +1686,7 @@ def jsonrpc_publish(self, p):
             'txid' : txid of resulting transaction if succesful
             'nout' : nout of the resulting support claim if succesful
             'fee' : fee paid for the claim transaction if succesful
-            'claimid' : claimid of the resulting transaction
+            'claim_id' : claimid of the resulting transaction
```
lyoshenka: is changing this key safe? it looks like `fee` and `claimid` were not return at all in the past, so they can't be in use anywhere, right?

lyoshenka:
If https://github.com/lbryio/lbry/pull/395#discussion_r96033190 is ok, then this is good to go

kaykurokawa:
It is not used in the web_ui or anywhere else in lbrynet

lyoshenka:
Cool. Merge it.

PR #397, opened 2017-01-10
The Great API Renaming

lyoshenka:
Several things happened here:

- Renamed many API endpoints. The goal is to use noun_verb style as much as possible (e.g. file_seed, claim_new, etc). 
- Old API endpoints still work, but are deprecated and ready for removal.
- CLI interaction is more pleasant. Error messages are nicer, help text is nicer, commands list is nicer.
- CLI no longer knows/cares whether or not api authentication is being used (its all handled in client)
- Subhandler functionality has been removed from AuthJsonRpc. We don't use it.
- Removed trailing ellipses from a bunch of messages.

@jackrobison the daemon ignores `use_auth_http` in daemon_settings.yml. this can cause the daemon and the CLI to be out of sync regarding when to use authenticated requests.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -94,10 +94,6 @@ CONNECTION_MESSAGES = {
                               "If this continues try restarting LBRY",
 }
 
-BAD_REQUEST = 400
-NOT_FOUND = 404
-OK_CODE = 200
-
 PENDING_ID = "not set"
 SHORT_ID_LEN = 20
 
@@ -206,10 +202,7 @@ class Daemon(AuthJSONRPCServer):
         AuthJSONRPCServer.__init__(self, conf.settings.use_auth_http)
         reactor.addSystemEventTrigger('before', 'shutdown', self._shutdown)
 
-        self.allowed_during_startup = [
-            'get_time_behind_blockchain', 'stop',
-            'status', 'version'
-        ]
+        self.allowed_during_startup = ['get_time_behind_blockchain', 'stop','status', 'version']
         last_version = {'last_version': {'lbrynet': lbrynet_version, 'lbryum': lbryum_version}}
         conf.settings.update(last_version)
         self.db_dir = conf.settings.data_dir
@@ -1122,9 +1115,9 @@ class Daemon(AuthJSONRPCServer):
 
             d.addCallback(_include_blockchain_status)
 
-        d.addCallback(lambda x: self._render_response(response, OK_CODE))
+        d.addCallback(lambda x: self._render_response(response))
         return d
-        # return self._render_response(response, OK_CODE)
+        # return self._render_response(response)
 
     def jsonrpc_get_best_blockhash(self):
         """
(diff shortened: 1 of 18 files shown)


kauffj:
Hooray!!!

lyoshenka:
Sorry about the large PR. Reading this commit-by-commit may be easier.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -318,5 +318,5 @@ def _callback_render(self, result, request, id_, version, auth_required=False):
             log.exception("Failed to render API response: %s", result)
             self._render_error(err, request, id_, version)
 
-    def _render_response(self, result, code):
-        return defer.succeed({'result': result, 'code': code})
+    def _render_response(self, result):
+        return defer.succeed({'result': result, 'code': 200})
```
jobevers: even here, the code is never used.  In _callback_render, it only pulls out the 'result'

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1052,95 +1054,113 @@ def _run_scripts(self):
...
-                r['message'] = "Catching up with the blockchain"
-                r['progress'] = 0
-        return self._render_response(r, OK_CODE)
+            d.addCallback(_include_session_status)
 
-    def jsonrpc_is_first_run(self):
-        """
-        Check if this is the first time lbrynet daemon has been run
+        if p.get('blockchain_status', False) and self.session and self.session.wallet:
+            # calculate blocks_behind more accurately
```
jobevers: Should this `# TODO: calculate blocks_behind more accurately`?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1301,22 +1309,36 @@ def jsonrpc_help(self, p=None):
...
-            return self._render_response(sorted(self.callable_methods.keys()), OK_CODE)
-        elif 'callable_during_start' in p.keys():
-            return self._render_response(self.allowed_during_startup, OK_CODE)
-        elif 'function' in p.keys():
-            func_path = p['function']
-            function = self.callable_methods.get(func_path)
-            return self._render_response(function.__doc__, OK_CODE)
+            return self._render_response(", ".join(sorted(self.callable_methods.keys())), OK_CODE)
+        elif 'callable_during_startup' in p:
+            return self._render_response(", ".join(sorted(self.allowed_during_startup)), OK_CODE)
```
jobevers: Why change from returning a list?

jobevers:
A message indicating something went wrong would be useful when the wrong params are provided:

```
$ lbrynet-cli help test=t
Function to retrieve docstring for API function

Args:
    'function': optional, function to retrieve documentation for
    'callable_during_startup': optional, returns functions that are callable during startup
Returns:
    if given a function, returns given documentation
    if given callable_during_startup flag, returns list of
    functions callable during the startup sequence
    if no params are given, returns the list of callable functions
```

jobevers:
Perhaps out of scope for this, `null` isn't useful to know if the command was successful or not - if thats an easy fix, might be worth doing while this is all still fresh.

```
$ lbrynet-cli configure_ui branch=master
null
```

jobevers:
A couple of small comments, but otherwise the UI works for me, the CLI works for me.

lbrynet/lbrynet_daemon/Daemon.py
lyoshenka: seems like help should return text. i added `jsonrpc_commands` to return a list of possible commands

PR #399, opened 2017-01-11
Add open_file() RPC method

alexliebowitz:


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2174,6 +2174,27 @@ def jsonrpc_configure_ui(self, p):
...
+        Returns:
+            True, opens file
+        """
+        path = p['path']
+        if sys.platform == 'darwin':
+            d = threads.deferToThread(subprocess.Popen, ['open', path])
+        elif os.name == 'posix':
+            d = threads.deferToThread(subprocess.Popen, ['xdg-open', path])
+        elif sys.platform == 'win32':
+            d = threads.deferToThread(os.startfile, path)
```
jobevers: given that this code is nearly identical to `reveal` I'd like there to be some abstraction over the OS for this sort of thing. Its better to do the if statement in a factory that gives you back a class that allows the rest of the code to not care what OS we're on.

roughly

```
def file_factory(path):
    if sys.platform == 'darwin':
        return DarwinFile(path)
    ...

class DarwinFile(object):
    def open(self):
        <code to open>
    def reveal(self):
        <code to reveal>

etc
```

jobevers:
Reassigning to grin so that he can give feedback on how this fits into the renaming.

lyoshenka:
I agree - abstract the OS differences as much as possible. Also happy to combine this with reveal. Maybe it should be called `file_open` and it could take an `open_in_folder` option.

I'm also concerned about security here. Do we really want the daemon opening whatever it's told to open? Shouldn't it check that this is a file it knows about, that its in the downloads folder, etc.

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: is deferToThread needed?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2174,6 +2174,27 @@ def jsonrpc_configure_ui(self, p):
...
+        Instruct the OS to open a file.
+
+        Args:
+            'path': path of file to be opened
+        Returns:
+            True, opens file
+        """
+        path = p['path']
+        if sys.platform == 'darwin':
+            d = threads.deferToThread(subprocess.Popen, ['open', path])
```
jackrobison: is deferToThread needed?

alexliebowitz:
Re: Job's idea, both of us and Jeremy have kicked it around and we decided to split the difference--create a new module called something like file_utils with functions revea()l and open() that have separate switch statements.

lbrynet/lbrynet_daemon/Daemon.py
alexliebowitz: Job, Jeremy and me have kicked this around and we decided to split the difference: create a new module called something like file_utils with functions revea() and open() that have separate switch statements. A lot simpler than making three classes and a factory but still abstracts out the OS handling stuff.

alexliebowitz:
Re: Grin's security concerns, the best way to handle this would be to pass in a `(file name || sd_hash)` instead of a path.* This is a more intuitive API anyway.

\* Although Jack pointed out today that SD hash doesn't quite work as a unique ID for entries in the file manager; you could have 2 files with the same SD hash. Same with claim ID. We're now planning to create a new unique ID field that will replace SD hash in most API calls. Not sure if that will end up getting merged before this PR or not.

alexliebowitz:
Closing this, work continues in #410

PR #402, opened 2017-01-11
Bug Fix for ConnectionManager

jobevers:
Fixes https://app.asana.com/0/142330900434470/239832897034382

See the individual commits comments for more details.


lbrynet/core/client/DownloadManager.py
```diff
@@ -59,8 +58,8 @@ def check_stop(result, manager):
         d1.addBoth(check_stop, "progress manager")
         d2 = self.connection_manager.stop()
         d2.addBoth(check_stop, "connection manager")
-        dl = defer.DeferredList([d1, d2])
-        dl.addCallback(lambda xs: False not in xs)
+        dl = defer.DeferredList([d1, d2], fireOnOneErrback=True, consumeErrors=True)
```
jackrobison: Why the `fireOnOneErrback` and `consumeErrors`?

jobevers: Reconsidering - `fireOnOneErrback` should probably be False so that the other deferred (the one that didn't fail) can continue doing its thing. I'll fix that.

`consumeErrors` should be set, otherwise if, for example, `d1` fails the error will be unhandled by d1 (since there are no errbacks attached to d1) and will also get passed along to possible errbacks of `dl`.  `consumeErrors` helps prevent spurious 'Unhandled error in Deferred' messages.

lbrynet/lbryfilemanager/EncryptedFileDownloader.py
```diff
@@ -96,18 +100,15 @@ def _reupload(self):
...
+        log.debug('Stopping download for %s', self.sd_hash)
         # EncryptedFileSaver deletes metadata when it's stopped. We don't want that here.
-        d = EncryptedFileDownloader.stop(self, err=err)
+        yield EncryptedFileDownloader.stop(self, err=err)
         if change_status is True:
-            self.saving_status = True
-            d.addCallback(lambda _: self._save_status())
-            d.addCallback(lambda _: set_saving_status_done())
-        return d
+            self._saving_status = True
```
jobevers: will do

PR #404, opened 2017-01-12
Allow requests that match allowed-origin

jobevers:
Also @jackrobison can you explain what check_headers does that `access-control-allow-origin` doesn't protect against.


jobevers:
@jackrobison please comment on this

jackrobison:
@jobevers 
checking the headers in the server prevents incoming requests from having a non complying origin - although the request can still be forged. access-control is just for the browser.

jackrobison:
LGTM

jobevers:
linux tests passed

PR #406, opened 2017-01-13
add installation id, re-initialize lbryid on each start

jackrobison:


lbrynet/analytics/events.py
```diff
@@ -85,7 +85,7 @@ def _event(self, event, event_properties=None):
     def _properties(self, event_properties=None):
         event_properties = event_properties or {}
         properties = {
-            'lbry_id': self.lbryid,
+            'installation_id': self.installation_id,
```
jobevers: the label on this needs to stay `lbry_id`.  Otherwise, you end up with some set of events that have a lbry_id column and another set of events that have an installation_id column.

An alternative would be keep lbry_id named as it is and create a new identifier, dht_id, or something.

jobevers:
LGTM

PR #408, opened 2017-01-13
non-blocking version check

jackrobison:
this fixes a bug where if github is down the app will fail to start.

-check for new version every 30 min instead of every 12 hours

-check connection problems every 30 seconds instead of every second

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -9,6 +9,7 @@ import requests
 import urllib
 import simplejson as json
 import textwrap
+from requests import exceptions as requests_exceptions
 from decimal import Decimal
 
 from twisted.web import server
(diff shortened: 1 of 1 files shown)


kaykurokawa:
instead of trapping urllib2.HTTPError you probably wanna trap requests.exception.RequestException http://docs.python-requests.org/en/master/_modules/requests/exceptions/ which should cover all the possible exceptions that can be raised from using requests.get

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -138,38 +139,18 @@ def __init__(self, daemon):
...
-                version, lbryum_version,
-                utils.version_is_greater_than(version, lbryum_version)
-            )
-            self.daemon.git_lbryum_version = version
-            return defer.succeed(None)
-        except Exception:
-            log.info("Failed to get lbryum version from git")
-            self.daemon.git_lbryum_version = None
-            return defer.fail(None)
+        self.daemon.git_lbryum_version = get_lbryum_version_from_github()
```
jobevers: Why take away the logging information?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -138,38 +139,18 @@ def __init__(self, daemon):
         self.daemon = daemon
 
     def __call__(self):
-        d = self._get_lbrynet_version()
+        d = threads.deferToThread(self._get_lbrynet_version)
+        d.addErrback(Failure.trap, HTTPError)
+        d.addErrback(log.fail(), "Error checking for lbrynet updates")
         d.addCallback(lambda _: self._get_lbryum_version())
+        d.addErrback(Failure.trap, HTTPError)
+        d.addErrback(log.fail(), "Error checking for lbryum updates")
```
jobevers: This would be nicer as an inlineCallback

jobevers:
I'd prefer to not do this in a threaded fashion and use requests-futures or twisted's web client instead.

jobevers:
To clarify my earlier comment - the threading approach is fine for now, but changing to an async request would be a good small project for the future.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -128,38 +129,34 @@ def __init__(self, daemon):
         self.daemon = daemon
 
     def __call__(self):
-        d = self._get_lbrynet_version()
+        d = threads.deferToThread(self._get_lbrynet_version)
+        d.addErrback(self._trap_and_log_error, 'lbrynet')
         d.addCallback(lambda _: self._get_lbryum_version())
```
jobevers: I don't think once the original function that was called in deferToThreads, control goes back to the reactor this. that means that get_lbryum_version happens synchronosly and will block the reactor thread until a response has come back.

jackrobison: Yup, TIL

jobevers:
LGTM.

PR #409, opened 2017-01-13
Fix wallet balance interfaces

kaykurokawa:
https://github.com/lbryio/lbry/commit/9230fb24dd0eaae379273e152a1d78a3f714702d just removes an unused get_transaction function from Wallet 

https://github.com/lbryio/lbry/commit/39d3947e88cece94209d049c21b3bf31ba6c436a fixes the Wallet interface for obtaining the balance, which was previously confusing and has lead to some bugs. 

Previously,  external classes used class member wallet_balance to fetch the balance of the wallet. However there is a method named get_balance()  which was used by the wallet to periodically pull the balance from lbryum, but should not be used by external classes to obtain the balance. We rename the previous get_balance() function as update_balance()  to avoid confusion , and change the function slightly to have it actually set class member wallet_balance. 

Further, wallet_balance did not consider reserved points (used for reserving credits for downloading) or queued payments (used to schedule payments at a later time) that should be subtracted to prevent external classes from accessing credits that are reserved for future uses.  Thus, a new function get_balance() is created that returns the balance with all the reserved and queued credits subtracted out. Now wallet_balance is only used to store the latest balance update from lbryum wallet. External classes will now exclusively uses get_balance() to retrieve the wallet balance. 

Claim commands also did not consider reserved points and queued payments before making the transaction, thus it could end up causing scheduled payments to fail. Now claim commands will use get_balance() to make sure that it has enough funds. 

Point reservation also did not consider queued payments, thus it could reserve more points than you had available. 

https://github.com/lbryio/lbry/commit/ebcddccd03325088bb8799c820dbf24be5b426a5 adds unit tests to test the above functionalities

Note that this does not solve all the wallet race condition issue, some lbryum changes are necessary (transactions need to be immediately added to wallet upon broadcast)

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -19,7 +19,8 @@ from lbryum.commands import known_commands, Commands
 from lbrynet.core.sqlite_helpers import rerun_if_locked
 from lbrynet.interfaces import IRequestCreator, IQueryHandlerFactory, IQueryHandler, IWallet
 from lbrynet.core.client.ClientRequest import ClientRequest
-from lbrynet.core.Error import UnknownNameError, InvalidStreamInfoError, RequestCanceledError
+from lbrynet.core.Error import (UnknownNameError, InvalidStreamInfoError, RequestCanceledError,
+                            InsufficientFundsError)
 from lbrynet.db_migrator.migrate1to2 import UNSET_NOUT
 from lbrynet.metadata.Metadata import Metadata
 
@@ -285,19 +286,12 @@ class Wallet(object):
             else:
                 d = defer.succeed(True)
 
-            d.addCallback(lambda _: self.get_balance())
-
-            def set_wallet_balance(balance):
-                if self.wallet_balance != balance:
-                    log.debug("Got a new balance: %s", str(balance))
-                self.wallet_balance = balance
-
             def log_error(err):
                 if isinstance(err, AttributeError):
                     log.warning("Failed to get an updated balance")
                     log.warning("Last balance update: %s", str(self.wallet_balance))
 
-            d.addCallbacks(set_wallet_balance, log_error)
+            d.addCallbacks(lambda _: self.update_balance(), log_error)
             return d
 
         d.addCallback(lambda should_run: do_manage() if should_run else None)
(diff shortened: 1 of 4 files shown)


lbrynet/core/Wallet.py
```diff
@@ -329,6 +322,18 @@ def set_manage_not_running(arg):
         d.addBoth(set_manage_not_running)
         return d
 
+    def update_balance(self):
```
jobevers: use inlineCallbacks:

```
@defer.inlineCallbacks
def update_balance(self):
    balance = yield self._update_balance()
    if self.wallet_balance != balance:
        log.debug("Got a new balance: %s", balance)
    self.wallet_balance = balance
```

lbrynet/core/Wallet.py
```diff
@@ -431,13 +436,13 @@ def set_address_for_peer(address):
         return d
 
     def _send_payments(self):
+
```
jobevers: this blank line isn't necessary.

lbrynet/core/Wallet.py
```diff
@@ -750,6 +760,7 @@ def get_status(claims):
         d.addCallback(get_status)
         return d
 
+
```
jobevers: only one blank line between functions

tests/unit/core/test_Wallet.py
```diff
@@ -128,3 +133,59 @@ def success_abandon_claim(self, claim_outpoint):
         d = wallet.abandon_claim("0578c161ad8d36a7580c557d7444f967ea7f988e194c20d0e3c42c3cabf110dd", 1)
         d.addCallback(lambda claim_out: check_out(claim_out))
         return d
+
+    def test_point_reservation_and_balance(self):
```
jobevers: inlineCallbacks would make this code more readable.

jobevers:
Overall, LGTM.  I added a couple of minor suggestions that are worth implementing.

kaykurokawa:
Fixed spacing and made update_balance into an inlinecallback (they are nice, did not know about them). 

Not sure how to rewrite the tests with them though, that will be on a separate pull request.

kaykurokawa:
Rebased to master and squashed

PR #410, opened 2017-01-13
Add open() API method and refactor/update reveal() API method (WIP)

alexliebowitz:
- Factors out the core logic for both of these into methods on a new module called lbry.core.file_utils
- Both methods now take an SD hash instead of a path (more logical and improves security)
- reveal() now works on Windows
- No longer spins off new threads for the OS calls since they're non-blocking (thanks to @jackrobison for pointing out this wasn't necessary)

Todo before this is ready:
- Tests
- The SD hash is not really the right identifier for this, as you can have multiple files downloaded with the same SD hash. Very soon Jack is planning to create a new unique ID for entries in the file manager, so we should probably wait until that change is in and then switch this to use the new ID.

Fixes this issue: https://app.asana.com/0/search/246741549780800/161594953616366

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -2220,7 +2220,7 @@ class Daemon(AuthJSONRPCServer):
         return d
 
     @AuthJSONRPCServer.auth_required
-    def jsonrpc_open_file(self, p):
+    def jsonrpc_open(self, p):
         """
         Instruct the OS to open a file.
 
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -38,7 +38,7 @@ from lbrynet.lbrynet_daemon.Downloader import GetStream
 from lbrynet.lbrynet_daemon.Publisher import Publisher
 from lbrynet.lbrynet_daemon.ExchangeRateManager import ExchangeRateManager
 from lbrynet.lbrynet_daemon.auth.server import AuthJSONRPCServer
-from lbrynet.core import log_support, utils
+from lbrynet.core import log_support, utils, file_utils
 from lbrynet.core import system_info
 from lbrynet.core.StreamDescriptor import StreamDescriptorIdentifier, download_sd_blob
 from lbrynet.core.StreamDescriptor import BlobStreamDescriptorReader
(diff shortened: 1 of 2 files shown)


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2255,25 +2255,52 @@ def jsonrpc_configure_ui(self, p):
         return d
 
     @AuthJSONRPCServer.auth_required
+    def jsonrpc_open(self, p):
```
jobevers: write this using inlineCallbacks:

@defer.inlineCallbacks
def jsonrpc_open(self, p):
    """<docstring>"""
    if 'sd_hash' not in p:
        raise ValueError('sd_hash is required')
    lbry_file = yield self._get_lbry_file(FileID.SD_HASH, p['sd_hash'])
    if not lbry_file:
        raise Exception('Unable to find file for {}'.format(p['sd_hash']))
    try:
        file_utils.start(lbry_file['download_path'])
    except IOError:
        pass
    defer.returnValue(True)
```

File: lbrynet/lbrynet_daemon/Daemon.py [hunk 2]
```diff
@@ -2255,25 +2255,52 @@ def jsonrpc_configure_ui(self, p):
...
+            except IOError:
+                pass
+
+        d = self._get_lbry_file(FileID.SD_HASH, p['sd_hash'])
+        d.addCallback(_open_lbry_file)
+        d.addCallback(lambda _: self._render_response(True, OK_CODE))
+
+        return d
+
+    @AuthJSONRPCServer.auth_required
```
jobevers: inlineCallback this too

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2255,25 +2255,52 @@ def jsonrpc_configure_ui(self, p):
...
+            except IOError:
+                pass
+
+        d = self._get_lbry_file(FileID.SD_HASH, p['sd_hash'])
+        d.addCallback(_reveal_lbry_file)
+        d.addCallback(lambda _: self._render_response(True, OK_CODE))
+
         return d
 
+
```
jobevers: No need for this, leave only a single line between functions

jobevers:
Also, make sure that the CI checks are passing.

lbrynet/lbrynet_daemon/Daemon.py
alexliebowitz: Thanks for the tip--done.

jobevers:
I wouldn't worry about adding tests for this; we don't have a good way to mock out the filemanager (for the _get_lbry_file call) and you don't really want files being opened / revealed during tests so its hard to say what you'd be testing.

jobevers:
pylint errors:
```
+./run_pylint.sh
************* Module lbrynet.lbrynet_daemon.Daemon
W:  7, 0: Unused import subprocess (unused-import)
W:  8, 0: Unused import sys (unused-import)
************* Module lbrynet.core.file_utils
C:  6, 0: Trailing whitespace (trailing-whitespace)
```

PR #413, opened 2017-01-16
fix log message in publisher

kaykurokawa:
claim_out['fee'] was not returned as float, so this log message would throw exception. Make sure we do the conversion here.

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -604,6 +604,12 @@ class Wallet(object):
             meta_for_return[k] = new_metadata[k]
         return defer.succeed(Metadata(meta_for_return))
 
+
+    def _process_claim_out(self, claim_out):
+        claim_out.pop('success')
+        claim_out['fee'] = float(claim_out['fee'])
+        return claim_out
+
     """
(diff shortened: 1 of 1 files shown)


jobevers:
I assume that if claim_out['fee'] is not a float, its an int?  Could it be None?

kaykurokawa:
its a string. Should be converted to decimal, but I'm looking right now if that's going to cause problem with the Daemon output. 

I don't really like these string formating methods, why don't we just use '{} {}'.format(var,var)

kaykurokawa:
I looked through the Daemon, and since it expects outputs from the wallet to be in float most every where else, I converted occurrences of claim_out['fee'] to be in float, instead of having the log message do the conversion. 

The daemon and wallet will need to be changed to work with Decimal instead of float, but that can be changed in a seperate pull request.

lyoshenka:
lgtm

jobevers:
`I don't really like these string formating methods, why don't we just use '{} {}'.format(var,var)`

In general, I agree that way is better. But the logging module offers lazy evaluation of the log message and arguments, and unfortunately, it only supports the `%`-stype formatting.

PR #417, opened 2017-01-17
refactor conf.settings

lyoshenka:
This is what happens when I need to change a single env variable and it doesn't work, and then I find a bunch of other bugs and weirdness.

Changes:
- **conf setup is much more sane. different conf types are stored differently and loaded with the right precedence. all conf variables work. there's no more danger of saving env/cli variables into daemon_settings.yml**
- conf.settings now uses dict style, not getattr style
- no more startup_scripts
- updated tests to use mock conf.Config()
- `API_INTERFACE` env var is renamed to `api_host` (change is backwards-compatible, but please update lbry-in-a-box eventually)

Consideration: invalid keys in daemon_settings.yml will now cause an exception. should that be handled better?


lyoshenka:
@kauffj public lbry server can go live after this PR is merged

lbrynet/conf.py
```diff
@@ -1,116 +1,99 @@
...
+    def _convert_key(key):
+        return ENV_NAMESPACE + key.upper()
 
-        if object or dict, follow default envparse rules, if tuple
-        it needs to be of the form (cast, default) or (cast, default, subcast)
+    @staticmethod
+    def _convert_value(value):
+        """
+        Allow value to be specified as a tuple or list. If you do this, the tuple/list
+        must be of the form (cast, default) or (cast, default, subcast)
```
jobevers: No need to change the docstring. The original docstring is inline with PEP257.  Docstrings are supposed to be:

```
"""A single line that provides a short description.

An optional longer explanation after a blank line. With the closing quotes
on a line on their own.
"""
```

lbrynet/conf.py
```diff
@@ -119,276 +102,289 @@ def _convert_value(self, value):
...
+    'upload_log': (bool, True),
+    'use_auth_http': (bool, False),
+    'use_upnp': (bool, True),
+    'wallet': (str, LBRYUM_WALLET),
+}
 
-    def get_conf_filename(self):
-        return get_settings_file_ext(self.ensure_data_dir())
 
+class Config:
```
jobevers: python classes inherit from object: `class Config(object):`

lbrynet/conf.py
```diff
@@ -119,276 +102,289 @@ def _convert_value(self, value):
...
+    def _init_defaults(self):
+        self.defaults = self.__fixed.copy()
+        for k, v in self.__adjustable.iteritems():
+            self.defaults[k] = v[1]
+
+    def _assert_valid_settings(self):
+        for s in [self.conf_file_settings, self.env_settings, self.cli_settings,
+                  self.runtime_settings]:
+            for name in s:
+                assert name in self.defaults, IndexError('%s is not a valid setting' % name)
```
jobevers: I don't like the pattern of using assertions like this.

```
Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.
```
https://wiki.python.org/moin/UsingAssertionsEffectively

I generally prefer this to look like:
```
for name in settings:
    if name not in self.defaults:
        raise IndexError('{} is not a valid setting'.format(name))
```

lbrynet/conf.py
```diff
@@ -119,276 +102,289 @@ def _convert_value(self, value):
...
+
+    def _combine_settings(self):
+        self.combined_settings = {}
+        for s in [self.defaults, self.conf_file_settings, self.env_settings, self.cli_settings,
+                  self.runtime_settings]:
+            self.combined_settings.update(s)
+
+    def get(self, name):
+        assert name in self.defaults, IndexError('%s is not a valid setting' % name)
+        return self.combined_settings[name]
```
jobevers: Its more 'pythonic' to write this as

```
try:
    return self.combined_settings[name]
except KeyError:
    raise IndexError('{} is not a valid setting'.format(name))
```

lbrynet/conf.py
```diff
@@ -119,276 +102,289 @@ def _convert_value(self, value):
...
+
+    def set(self, name, value, set_conf_setting=False):
```
jobevers: Can you document what `set_conf_setting` does

lbrynet/conf.py
```diff
@@ -1,116 +1,99 @@
...
     """An Env parser that automatically namespaces the variables with LBRY"""
-    NAMESPACE = 'LBRY_'
+
     def __init__(self, **schema):
         self.original_schema = schema
         my_schema = {
             self._convert_key(key): self._convert_value(value)
             for key, value in schema.items()
-        }
+            }
```
jobevers: I think the de-dented bracing is better formatting and more consistent with how the code is formatted in the rest of the library.

lbrynet/conf.py
```diff
@@ -119,276 +102,289 @@ def _convert_value(self, value):
...
+    'use_upnp': (bool, True),
+    'wallet': (str, LBRYUM_WALLET),
+}
 
-    def get_conf_filename(self):
-        return get_settings_file_ext(self.ensure_data_dir())
 
+class Config:
+    def __init__(self, fixed_defaults, adjustable_defaults, conf_file_settings={}, environment=None,
+                 cli_settings={}):
```
jobevers: Avoid using mutable objects as default arguments.

jobevers:
Fix up the long line errors and a couple of minor style comments, otherwise LGTM.  Merge when you're ready.

lbrynet/conf.py
lyoshenka: i've been fighting with pycharm to set up autoformat correctly, but its still not great. if you have a config that works, send it over. I'd love to use it

lbrynet/conf.py
lyoshenka: im not sure whether this is a case of bad user input. my understanding is that assertions are special because they can be turned off, so you cant rely on them for validating things like "did the user enter a number when we asked for one". but in this case, they are being used for internal validation. i can see validating the config file as user input, but not the get()/set() methods. does that make sense?

Issue #418, opened 2017-01-17
Unhandled exceptions in daemon commands don't return results


kaykurokawa:
If there is an unhandled exception thrown while processing a daemon command that was requested through lbrynet-cli , lbrynet-cli will hang forever. 

There should probably be a function wrapper for all daemon commands that will catch exceptions and return them to the cli.

┆Attachments: <a href="https:&#x2F;&#x2F;app.asana.com&#x2F;app&#x2F;asana&#x2F;-&#x2F;download_asset?asset_id&#x3D;248620640632445">lbrynet api doesn&#39;t respond.mov</a>

lyoshenka:
Fixed

PR #421, opened 2017-01-18
Add back id

jobevers:
When a command would error, the `_render_error` function was failing to call because the id_ parameter was missing. Additionally, twisted was silencing the error.


jobevers:
Fixes https://github.com/lbryio/lbry/issues/418

PR #423, opened 2017-01-18
Safegaurd against there not being a wallet on status call

jobevers:
Also switch status call to an inlineCallback


jobevers:
it passed on linux and windows, merging.

PR #428, opened 2017-01-19
Blob verification fixes

jobevers:
Move the blob verification to the actual Blob object; remove the check on verification time.

I'm not sure what checking verification time against ctime gets us, except some protection against an accidental modification of the blob.

I think a better approach, for future development, would be an occasional task that goes through and checks some subset of the blobs comparing the hash of their contents to their name and purging ones that don't match.

### DIFF
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -36,7 +36,7 @@ class BlobManager(DHTHashSupplier):
     def blob_completed(self, blob, next_announce_time=None):
         pass
 
-    def completed_blobs(self, blobs_to_check):
+    def completed_blobs(self, blobhashes_to_check):
         pass
 
     def hashes_to_announce(self):
@@ -136,8 +136,8 @@ class DiskBlobManager(BlobManager):
         d.addCallback(lambda _: self._immediate_announce([blob.blob_hash]))
         return d
 
-    def completed_blobs(self, blobs_to_check):
-        return self._completed_blobs(blobs_to_check)
+    def completed_blobs(self, blobhashes_to_check):
+        return self._completed_blobs(blobhashes_to_check)
 
     def hashes_to_announce(self):
         next_announce_time = time.time() + self.hash_reannounce_time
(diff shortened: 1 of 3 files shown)


lbrynet/core/HashBlob.py
```diff
@@ -206,6 +214,14 @@ def __init__(self, blob_dir, *args):
         self.file_path = os.path.join(blob_dir, self.blob_hash)
         self.setting_verified_blob_lock = threading.Lock()
         self.moved_verified_blob = False
+        if os.path.isfile(self.file_path):
+            self.set_length(os.path.getsize(self.file_path))
+            # This assumes that the hash of the blob has already been
+            # checked as part of the blob creation process. It might
+            # be worth having a function that checks the actual hash;
+            # its probably too expensive to have that check be part of
+            # this call.
```
lyoshenka: premature optimization? is this actually an issue?

lbrynet/core/BlobManager.py
```diff
@@ -126,35 +127,17 @@ def _make_new_blob(self, blob_hash, upload_allowed, length=None):
...
-
-        def check_completed(completed_blobs):
-
-            def set_length(length):
-                blob.length = length
-
-            if len(completed_blobs) == 1 and completed_blobs[0] == blob_hash:
-                blob.verified = True
-                inner_d = self._get_blob_length(blob_hash)
-                inner_d.addCallback(set_length)
```
lyoshenka: we're not losing `set_length`, right? I think its set below but im not sure

jobevers:
I will clean up the commits before merging.

lbrynet/core/HashBlob.py
```diff
@@ -78,14 +83,20 @@ class HashBlob(object):
...
         self.finished_deferred = None
-        self.verified = False
+        self._verified = False
         self.upload_allowed = upload_allowed
         self.readers = 0
 
+    @property
+    def verified(self):
+        # protect verified from being modified by other classes
+        return self._verified
```
lyoshenka: is this `@property` pattern the recommended thing to do in python? my Magic SenseTM is tingling

jobevers: @property is pretty common and definitely more pythonic than getters / setters

http://stackoverflow.com/questions/6618002/python-property-versus-getters-and-setters

lbrynet/core/BlobManager.py
jobevers: The length now gets calculated by the blob itself if its not passed into the HashBlob constructor.  Its only real use is when downloading a blob, you need to keep track of the data received.

I am going to take out `get_blob_length`

lbrynet/core/HashBlob.py
jobevers: on startup, if there are a lot of blobs, we'd end up having to read all of them from disk and calculate the hash, which can be quite slow.  Given that the hash was checked when the file was written, I don't think its necessary to recheck each time a HashBlob is initialized.

Also, its more code.  I'm attempting to justify not making the check.

lyoshenka: :+1:

lbrynet/core/HashBlob.py
lyoshenka: seems to violate the "no surprises" principle, but if that's what they recommend then its fine by me

PR #430, opened 2017-01-20
missed another verified time check

jobevers:


jobevers:
screw you osx.  merging.

PR #431, opened 2017-01-21
WIP: Add default blob peers

kaykurokawa:
This forces DHTPeerFinder to always append default blob peers to its found peers. Default blob peers are lbry servers that we know to be reliable. 

This patch is needed since the DHT is having problems returning peers which we know to have the blobs. 

This patch currently does not work because ConnectionManager class does a poor job of sorting through the peers.


lbrynet/conf.py
```diff
@@ -31,6 +31,21 @@
     ('lbrynet3.lbry.io', 4444)
 ]
 
+DEFAULT_BLOB_PEERS = [
+	('54.89.95.83',33),
```
jobevers: You want port 3333 for this.

kaykurokawa:
Fixed ports to 3333

Only add reflector nodes as they contain most of the blob files, adding other servers might make downloads longer as it has to look through all of them. 

Adding default blob peers needed to be done through the PeerManager, and make sure we don't add the same peer twice using a PeerSet class, the PeerSet class will also be useful in other places

kaykurokawa:
closing, as this won't be necessary for now.

PR #433, opened 2017-01-21
Clarify how to install, verify install, and make a change

kauffj:


INSTALL.md
```diff
@@ -66,20 +66,35 @@ Then, install the module by running `python setupmingw32.py install`.
...
  source lbry-venv/bin/activate
- ```
-
-Then, install the package in the new virtualenv:
  
- ```
- git clone https://github.com/lbryio/lbry.git
- cd lbry
  pip install -r requirements.txt
  pip install .
```
jobevers: probably worth changing this to `pip install -e .` as well

INSTALL.md
```diff
@@ -66,20 +66,35 @@ Then, install the module by running `python setupmingw32.py install`.
...
+If you are not already inside the virtual environment, run:
+
+```
+source lbry-venv/bin/activate
+```
+
+Then run:
+
+```
+python setup.py install
```
jobevers: `pip install -e .` is better

jobevers:
While using the raw `virtualenv` tool is fine, I do suggest using a virtualenv manager of some sort.  My favorite is `vex`.

I'm not sure if its worth adding to the instructions as its another step in the process, but if you're actually doing development, I recommend it.

Installation:
```
pip install vex
```

Making a new virtualenv: 
```
vex --make lbry
```

Switching virtualenvs:
```
vex lbry
```

lyoshenka:
@jobevers i agree, vex is great. but virtualenv is lower friction and more widely understood, so I'd rather the docs use that.

Issue #432, opened 2017-01-21
DHT Hash Announcer is not annoucing some blobs if it has a lot of blobs


kaykurokawa:
There is a hash_reannounce_time set to 1 hour in DHTHashSupplier , which determines how often hashes are reannounced. 

If a server takes more than 1 hour to announce all of its  blobs, than  there is going to be a constant back log of un- reannounced blobs that piles up. This means that some blobs will never be announced.  

The fix would be to make  hash reannounce time dynamic , based on how large a server's blob database is.

kaykurokawa:
each blob takes less than 1 second to announce, somewhere  around .2 to .7 seconds

jobevers:
The hashes are placed into a queue though which is processed first-in first-out, so it seems that it should get through all of them.  Maybe not in a timely fashion, but should get through.

kaykurokawa:
I think you are right, it should announce eventually, 
Its still a problem though because the queue will keep getting bigger without being cleared causing a memory leak.

jobevers:
Threw together https://github.com/lbryio/lbry/pull/435 the help investigate this issue

PR #435, opened 2017-01-23
WIP Debug DHT

jobevers:
To help debug: https://github.com/lbryio/lbry/issues/432.

Might be useful in general, but would want some way to clean up the announcement table otherwise it could get large for an installation with a lot of blobs.

### DIFF
--- a/lbrynet/core/server/DHTHashAnnouncer.py
+++ b/lbrynet/core/server/DHTHashAnnouncer.py
@@ -9,6 +9,29 @@ from twisted.internet import defer, reactor
 log = logging.getLogger(__name__)
 
 
+class DequeSet(object):
+    def __init__(self):
+        self._queue = collections.deque()
+        self._items = set()
+
+    def append(self, item):
+        if item in self._items:
+            return
+        self._queue.append(item)
+        self._items.add(item)
+
+    def popleft(self):
+        item = self._queue.popleft()
+        self._items.remove(item)
+        return item
+
+    def __len__(self):
+        return len(self._queue)
+
+    def __nonzero__(self):
+        return self._queue.__nonzero__()
+
+
 class DHTHashAnnouncer(object):
     """This class announces to the DHT that this peer has certain blobs"""
     def __init__(self, dht_node, peer_port):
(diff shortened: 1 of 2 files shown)


lyoshenka:
If we're gonna restructure the DB, let's do it right. I agree that storing last announce time is better than storing next announce time. Let's add the column you suggested, migrate the data, and drop the last_announce_time column on blobs (or at least ignore it)

jobevers:
I added a new table mostly for debugging; I didn't add a new column, this adds a new table that gets appended to each time a hash is announced.  Could run something like this on reflector and see if it does actually publish all of its blobs to the dht, and how frequently.

Definitely needs more work if we wanted to adapt this to be 'production' worthy.

lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -9,14 +9,37 @@
...
+
+    def popleft(self):
+        item = self._queue.popleft()
+        self._items.remove(item)
+        return item
+
+    def __len__(self):
+        return len(self._queue)
+
+    def __nonzero__(self):
```
jobevers: don't need this, if __nonzero__ doesn't exist, __len__ is called.

lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -9,14 +9,37 @@
 log = logging.getLogger(__name__)
 
 
+class DequeSet(object):
+    def __init__(self):
+        self._queue = collections.deque()
+        self._items = set()
+
+    def append(self, item):
```
jobevers: because the items being added are (h, announce_deferred), they aren't actually unique.  Should turn this into a special purposed data structure that is called like: `def append(self, blob_hash, announce_deferred)`
and have `_items` be a set of blobs, and `_queue` be a queue of (h, deferred) tuples.

kaykurokawa:
I think we can roll this out on a reflector branch instead of master for debugging purpose.

lyoshenka:
whats the next step here?

kaykurokawa:
https://github.com/lbryio/lbry/pull/465 will be a temporary fix for the same problem

This pull request, once worked on will be a superior fix but not high priority right now.

jobevers:
I'm going to close this until its time to revisit it.  The extra logging info though might be useful for reflector but I don't want to release it in general, and not sure I want the overhead of maintaining different version for reflector.

PR #450, opened 2017-01-29
Fix #449: Block to deep error

### LINKED ISSUE #449: Block too deep to generate proof
```
2017-01-20 13:50:39,046 ERROR    lbrynet.lbrynet_daemon.auth.server:214: Failed to process resolve_name: {u'message': u'Block too deep to generate proof', u'code': -32603}
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/twisted/python/threadpool.py", line 246, in inContext
    result = inContext.theWork()
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/twisted/python/threadpool.py", line 262, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/lbryum/commands.py", line 73, in func_wrapper
    return func(*args, **kwargs)
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/lbryum/commands.py", line 708, in requestvalueforname
    return self.network.synchronous_get(('blockchain.claimtrie.getvalue', [name, blockhash]))
  File "/Users/johnrobison/.virtualenvs/lbry/lib/python2.7/site-packages/lbryum/network.py", line 823, in synchronous_get
    raise BaseException(r.get('error'))
```

jobevers:
What was happening was the wallet claimed to be caught up before it
actually was and so the wallet’s local_height was still the value from
when lbry was last run, frequently more than 20 or 50 blocks
behind. _get_value_for_name uses the block at local_height as the
basis for the proof.  If _get_value_for_name is called during that
time between when the wallet claims to be caught up and it actually
is, the “Block too deep error” happens.  And since the discover page
of the UI does name resolution right away, the error basically happens
anytime somebody starts the app after not using it for a few hours.

This changes the startup behaviour of the wallet to
- use the `updated` callback provided by lbryum
- check that local_height and network_height match before declaring
  that the wallet has caught up

For reference, the error is raised here:
https://github.com/lbryio/lbrycrd/blob/1b896ae75b0fc7081ab9e4a1b7d45c94dcd73cbd/src/rpc/claimtrie.cpp#L688

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -887,7 +887,6 @@ class LBRYumWallet(Wallet):
         self._lag_counter = 0
         self.blocks_behind = 0
         self.catchup_progress = 0
-        self.max_behind = 0
 
     def _is_first_run(self):
         return (not self.printed_retrieving_headers and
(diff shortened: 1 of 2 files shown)


jobevers:
The windows test is failing, not due to this PR but because the build is broken for windows.

jobevers:
This depends on https://github.com/lbryio/lbryum/pull/73

kaykurokawa:
What was the point of using alert.info instead of log.info before ?

jobevers:
it was just logged differently, but I think its better to use consistent logging names.

kaykurokawa:
I think this code needs to take into account that remote block height could be lower than the local height (if you connect to an up to date server, disconnect, and than connect to a server that's behind).I think it would be better if the lbryum get_blocks_behind function could return a negative number instead of None so we can tell how far behind the  remote is. 

From the lbrynet side, not sure what would be good to do, as getvalueforname command would fail if local is ahead of remote . We could either wait for remote to catchup , fail immediately, or ignore and keep going. We should definitely log it as a log.warn so we can diagnose the issue if it happens.

kaykurokawa:
I think the is_lagging code that was removed in this patch was supposed to address some of that problem, I think jack wrote that part so he could comment on how frequently and when that was occurring.

jobevers:
The problem you describe (disconnect and reconnect to another server that is lagging) exists in the original code.

I think the solution should be in lbryum, not lbrynet: getvalueforname, or any other command that depends on the server and local chains to be caught up, should throw an error if the server is lagging.

jobevers:
Regarding the logic of whether the server is lagging, lbryum already has a function: https://github.com/lbryio/lbryum/blob/master/lib/network.py#L252

jobevers:
@kaykurokawa would you be okay with merging this code and have a future PR to address the further concerns you have?

lbrynet/core/Wallet.py
```diff
@@ -967,63 +968,34 @@ def _load_wallet(self):
...
-                # No progress. Might be lagging
-                self._lag_counter += 1
-                if self._lag_counter >= 900:
-                    self.is_lagging = True
-
-            self.blocks_behind = height_diff
+            assert self.blocks_behind == 0
+            self.network.unregister_callback(on_update_callback)
+            log.info("Wallet Loaded")
+            reactor.callFromThread(blockchain_caught_d.callback, True)
```
jackrobison: Why callFromThread?

jackrobison:
@kaykurokawa 

The off-behavior that `is_lagging` was originally intended for was when catchup would freeze up, which restarting would generally fix.

>From the lbrynet side, not sure what would be good to do, as getvalueforname command would fail if local is ahead of remote.

I think it should, but with the wallet setup only calling back after it's all caught up it shouldn't be a problem. If local is behind remote I think that should probably cause an error to be raised and logged in lbryum, and lbrynet should handle that situation in an errback.

jackrobison:
👍

kaykurokawa:
Wallet member variable "is_lagging" can be removed

Also same question as jack regarding callFromThread , 

other than that looks good to merge 

Future fixes needed:
lbryum commands that will fail if the remote is lagging , such as getvalueforname, should detect if it "is_lagging" before hand and throw exception 

there probably needs to be a way for lbrynet to be aware that it is stuck at a state where it is behind or ahead of remote server because startup will just get stuck without any feedback.

lbrynet/core/Wallet.py
jobevers: the `on_update_callback` is called by lbryum in the wallet network daemon thread.  `callFromThread` moves the  control back to the reactor thread.

jobevers:
removed the is_lagging attribute

PR #452, opened 2017-01-29
Retry startup

jobevers:


jobevers:
Jack commented that he's not sure some startup errors are recoverable in this way - like, if we've already started listening to 5279 it might fail on the second try while attempting to bind that port.

jackrobison:
It's probably worth testing crashing startup throughout the stages - teardown might be necessary (port binding and opening the databases come to mind as potential problems).

jobevers:
@jackrobison while there can be improvements to this that can be done, is there a reason not to merge it in?

jackrobison:
@jobevers LGTM, go for it

PR #453, opened 2017-01-30
Add details to error event

jobevers:


jobevers:
LGTM'd by grin in slack

PR #454, opened 2017-01-30
Claim after reflecting

jobevers:
Most of the changes in here are mechanical refactoring and conversion to inlinecallbacks.

The main change in logic is that I moved the name claim to happen AFTER the blobs have been pushed to reflector.  This will increase the amount of time it takes to publish but at the benefit of increasing the reliability of publishing

### DIFF
--- a/lbrynet/lbryfile/StreamDescriptor.py
+++ b/lbrynet/lbryfile/StreamDescriptor.py
@@ -80,19 +80,13 @@ def get_sd_info(stream_info_manager, stream_hash, include_blobs):
     return d
 
 
+@defer.inlineCallbacks
 def publish_sd_blob(stream_info_manager, blob_manager, stream_hash):
     descriptor_writer = BlobStreamDescriptorWriter(blob_manager)
-
-    d = get_sd_info(stream_info_manager, stream_hash, True)
-    d.addCallback(descriptor_writer.create_descriptor)
-
-    def add_sd_blob_to_stream(sd_blob_hash):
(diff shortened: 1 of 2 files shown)


jobevers:
Following up on a conversation with Jack, this will definitely slow down publishing.  An alternative solution would be to have a LoopingCall that checks with reflector and tries to upload/reupload.  We currently only reupload on startup, but as we want the daemon to be something that runs in the background having a LoopingCall is a better strategy.

The refactoring is worth keeping; if we'd rather go with the LoopingCall approach, I should take out https://github.com/lbryio/lbry/pull/454/commits/ce62c56312a240cace6610402d4666e7f71884c7.

jobevers:
@jackrobison thoughts on this?

jackrobison:
I think we should merge the publish_sd_blob refactor, the inline callbacks in [#477](https://github.com/lbryio/lbry/pull/477), and the Publisher refactor in [#476](https://github.com/lbryio/lbry/pull/476). (Sorry, I redid a bunch of what you already did)

476 moves the reflector logic fully out of Publisher and into Daemon. I think that pending all these refactors, we should remove `ManagedEncryptedFileDownloader._reupload` (called in `ManagedEncryptedFileDownloader.restore`) and move the reflect-on-startup behavior to happen in `EncryptedFileManager.start`. Presently we're reflecting the stream twice in a publish: once during `ManagedEncryptedFileDownloader.restore` called by`Publisher.set_status` and then again in `Publisher.start_reflector`. 

I think that uploading to reflector should be in Daemon after Publisher has returned, and probably not be blocking.

jobevers:
This can be closed in favor of #477

PR #456, opened 2017-01-30
nicer error output

lyoshenka:


lbrynet/lbrynet_daemon/auth/client.py
```diff
@@ -46,7 +46,7 @@ def call(self, method, params={}):
         pre_auth_post_data = {
             'version': '1.1',
             'method': method,
-            'params': [params],
+            'params': params,
```
jobevers: what is this change?  bugfix?

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -32,7 +32,10 @@ def __init__(self, err, code):
 
     @property
     def faultString(self):
-        return self.err.getTraceback()
+        try:
+            return self.err.getTraceback()
+        except AttributeError:
+            return self.err
```
jobevers: Does faultString need to be a string, maybe this should be `return str(self.err)`

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -163,13 +171,15 @@ def expire_session():
...
             log.warning("Unable to decode request json")
-            self._render_error(err, request, None)
+            self._render_error_string('Invalid JSON', request, None)
             return server.NOT_DONE_YET
 
         function_name = parsed.get('method')
-        args = parsed.get('params')
+        args = parsed.get('params', None)
+        if args is None:
+            args = {}
```
jobevers: Can be shortened to: `args = parsed.get('params', {})`

jobevers:
A couple of small comments / suggestions, after that LGTM

lbrynet/lbrynet_daemon/auth/client.py
lyoshenka: This is a change in how the api works. For some reason, the way params got passed in before was as a list with a dict as the first element of the list, and all the params in that dict. That's pretty dumb, and I fixed it when I made the jsonrpc functions explicitly declare their params. It's backwards-compatible for now (see `auth/server.py` lines 198-208) but that will be removed later.

lyoshenka: The change itself was done in a previous commit. This is just to get our client using the new way.

PR #457, opened 2017-01-31
Reflector v2

jackrobison:
add {‘sd_blob_hash’: …, ‘sd_blob_size ‘: …} query type with
{‘send_sd_blob’: True/False, ‘needed_blobs’: []} response

this allows the reflector client to know how much of a stream reflector
already has covered, as to minimize the number of subsequent requests
and prevent streams from being partially reflected

### DIFF
--- a/lbrynet/reflector/client/client.py
+++ b/lbrynet/reflector/client/client.py
@@ -1,61 +1,13 @@
-"""
-The reflector protocol (all dicts encoded in json):
-
-Client Handshake (sent once per connection, at the start of the connection):
-
-{
-    'version': 0,
-}
-
-
-Server Handshake (sent once per connection, after receiving the client handshake):
-
-{
-    'version': 0,
-}
-
-
-Client Info Request:
-
-{
-    'blob_hash': "<blob_hash>",
-    'blob_size': <blob_size>
-}
-
-
-Server Info Response (sent in response to Client Info Request):
-
-{
-    'send_blob': True|False
-}
-
-If response is 'YES', client may send a Client Blob Request or a Client Info Request.
-If response is 'NO', client may only send a Client Info Request
-
-
-Client Blob Request:
-
-{}  # Yes, this is an empty dictionary, in case something needs to go here in the future
-<raw blob_data>  # this blob data must match the info sent in the most recent Client Info Request
-
-
-Server Blob Response (sent in response to Client Blob Request):
-{
-    'received_blob': True
-}
-
-Client may now send another Client Info Request
-
-"""
 import json
 import logging
 
+from twisted.internet.error import ConnectionRefusedError
 from twisted.protocols.basic import FileSender
 from twisted.internet.protocol import Protocol, ClientFactory
 from twisted.internet import defer, error
 
-from lbrynet.reflector.common import IncompleteResponse
-
+from lbrynet.reflector.common import IncompleteResponse, ReflectorRequestError
+from lbrynet.reflector.common import REFLECTOR_V1, REFLECTOR_V2
 
 log = logging.getLogger(__name__)
 
@@ -63,24 +15,31 @@ log = logging.getLogger(__name__)
 class EncryptedFileReflectorClient(Protocol):
     #  Protocol stuff
     def connectionMade(self):
+        log.debug("Connected to reflector")
         self.blob_manager = self.factory.blob_manager
         self.response_buff = ''
         self.outgoing_buff = ''
         self.blob_hashes_to_send = []
         self.next_blob_to_send = None
-        self.blob_read_handle = None
-        self.received_handshake_response = False
-        self.protocol_version = None
+        self.read_handle = None
+        self.sent_stream_info = False
+        self.received_descriptor_response = False
+        self.protocol_version = self.factory.protocol_version
+        self.received_server_version = False
+        self.server_version = None
+        self.stream_descriptor = None
+        self.descriptor_needed = None
+        self.needed_blobs = []
+        self.reflected_blobs = []
         self.file_sender = None
         self.producer = None
         self.streaming = False
-        d = self.get_blobs_to_send(self.factory.stream_info_manager, self.factory.stream_hash)
+        d = self.load_descriptor()
         d.addCallback(lambda _: self.send_handshake())
         d.addErrback(
             lambda err: log.warning("An error occurred immediately: %s", err.getTraceback()))
 
     def dataReceived(self, data):
-        log.debug('Recieved %s', data)
         self.response_buff += data
         try:
             msg = self.parse_response(self.response_buff)
@@ -94,8 +53,8 @@ class EncryptedFileReflectorClient(Protocol):
 
     def connectionLost(self, reason):
         if reason.check(error.ConnectionDone):
-            log.info('Finished sending data via reflector')
-            self.factory.finished_deferred.callback(True)
+            log.debug('Finished sending data via reflector')
+            self.factory.finished_deferred.callback(self.reflected_blobs)
         else:
             log.debug('Reflector finished: %s', reason)
             self.factory.finished_deferred.callback(reason)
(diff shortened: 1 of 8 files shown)


jackrobison:
note: this uses jack.lbry.tech as the reflector server since reflector.lbry.io is running reflector v1.

lyoshenka:
I didn't test this, just read the code. Besides the unclear protocol description, this can go up unless @jobevers wants to read it too.

lbrynet/reflector/common.py
```diff
@@ -1,2 +1,18 @@
+REFLECTOR_V1 = 0
+REFLECTOR_V2 = 1
+
+
+class ReflectorClientVersionError(Exception):
```
jobevers: please add documentation to these errors.

tests/functional/test_reflector.py
```diff
@@ -98,7 +98,7 @@ def test_reflector(self):
             dht_node_class=Node
         )
 
-        self.stream_info_manager = EncryptedFileMetadataManager.TempEncryptedFileMetadataManager()
+        self.stream_info_manager = EncryptedFileMetadataManager.DBEncryptedFileMetadataManager(db_dir)
```
jobevers: why this change?

jobevers:
Since there are already tests for version 0 of the reflector protocol, I don't think adding tests for v1 would be that difficult and I'd like to see them.

jobevers:
There was some concern over race-conditions and what happens if multiple clients try to send a blob.  After looking over the blob manager and reflector code, I think that it gets handled fine.


When a client makes a request for a blob, reflector server asks the blob manager for the blob: https://github.com/lbryio/lbry/blob/master/lbrynet/reflector/server/server.py#L111
If the blob exists on disk, `blob.is_verified` will be true and so reflector will tell the client that it doesn't need to the blob.

If the blob doesn't exists, reflector will tell the client that it needs to blob.  The client will send over data.  This data gets written to a temporary file, which is opened here: https://github.com/lbryio/lbry/blob/master/lbrynet/core/HashBlob.py#L229

If another client comes through and initiates the same process for the same blob, the same `BlobFile` object will be returned, but a new writer will be added to the `BlobFile`. (Each writer is tracked: https://github.com/lbryio/lbry/blob/master/lbrynet/core/HashBlob.py#L89)

When a writer finishes, the blob gets moved from the temporary file to its hash based name: https://github.com/lbryio/lbry/blob/master/lbrynet/core/HashBlob.py#L284
and all of the other writers are cancelled: https://github.com/lbryio/lbry/blob/master/lbrynet/core/HashBlob.py#L157

jobevers:
I agree with @lyoshenka  that if we send reflector an SD blob and it tells us that it needs blobs X,Y,Z - the client should just send blobs X, Y, Z.  The client doesn't need to ask again if it needs X before sending it.

lbrynet/conf.py
```diff
@@ -193,7 +193,7 @@ def _convert_value(value):
     'pointtrader_server': (str, 'http://127.0.0.1:2424'),
     'reflector_port': (int, 5566),
     'reflector_reupload': (bool, True),
-    'reflector_servers': (list, [('reflector.lbry.io', 5566)], server_port),
+    'reflector_servers': (list, [('jack.lbry.tech', 5566)], server_port),
```
jackrobison: I'll revert that before merging, that's just so others can test it before going live on the production reflector.

lbrynet/reflector/client/client.py
```diff
@@ -4,16 +4,40 @@
...
+}
+
+Server Descriptor Response (sent in response to Client Descriptor Request):
+
+{
+    'send_sd_blob': True|False
+    'needed_blobs': []
+}
+
+If response is 'YES', client may send a Client Stream Descriptor or a Client Info Request.
```
jackrobison: -I'll rewrite the documentation - it isn't very clear.

-I see what you mean about not needing to make a {`send_blob`: x...} request before sending a blob out of the list populated by the initial `missing_blobs` response. I'd like to hold off on that for another pr.

-The requests and responses are json, the file transfer is raw.

lbrynet/reflector/client/client.py
```diff
@@ -118,32 +77,55 @@ def write(self, data):
...
-                if blob_hash is not None:
-                    self.blob_hashes_to_send.append(blob_hash)
-            log.debug("Preparing to send %i blobs", len(self.blob_hashes_to_send))
+    def get_validated_blobs(self, blobs_in_stream):
+        def get_blobs(blobs):
+            for (blob, _, _, blob_len) in blobs:
+                if blob:
+                    yield self.blob_manager.get_blob(blob, True, blob_len)
+
+        dl = defer.DeferredList(list(get_blobs(blobs_in_stream)))
```
jobevers: You'll want to consume errors here.

lbrynet/reflector/client/client.py
```diff
@@ -118,32 +77,55 @@ def write(self, data):
...
+        dl.addCallback(lambda blobs: [blob for _, blob in blobs if blob.is_validated()])
```
jobevers: I think you also want to check for success

lbrynet/reflector/server/server.py
```diff
@@ -68,26 +72,101 @@ def _get_valid_response(self, response_msg):
...
+            self.blob_manager.delete_blobs([blob.blob_hash])
+        else:
+            log.exception(err)
+        return False
+
+    def log_received_blob(self, blob, message):
+        log.info(message, blob.blob_hash[:16])
+        return True
+
+    def determine_descriptor_needed(self, sd_blob):
```
jobevers: Again, use inline callbacks.

This function is poorly named.  When I read the name, I expect that the result with be a boolean saying either yes or no that I need the descriptor.

Also, the function is too complicated.  I will yell this until I'm hoarse: functions need to do one thing.

Each block inside the if statement could be made into their own functions.

All of the callbacks added to the blob_finished_d should be in their own function.

lbrynet/reflector/server/server.py
```diff
@@ -68,26 +72,101 @@ def _get_valid_response(self, response_msg):
...
+            self.blob_finished_d.addCallback(lambda _: self.blob_manager.blob_completed(sd_blob))
+            self.blob_finished_d.addCallback(lambda _: self.close_blob())
+            self.blob_finished_d.addCallback(lambda _: self.log_received_blob(sd_blob, log_msg))
+            self.blob_finished_d.addErrback(self.clean_up_failed_upload, sd_blob)
+            self.blob_finished_d.addCallback(lambda result: {'received_sd_blob': result})
+            self.blob_finished_d.addCallback(self.send_response)
+            d = defer.succeed({'send_sd_blob': True})
+        return d
+
+    def determine_missing_blobs(self, sd_blob):
```
jobevers: switch this to inlinecallbacks and get rid of the nested / inner functions

lbrynet/reflector/client/client.py
```diff
@@ -118,32 +77,55 @@ def write(self, data):
...
-                    self.blob_hashes_to_send.append(blob_hash)
-            log.debug("Preparing to send %i blobs", len(self.blob_hashes_to_send))
+    def get_validated_blobs(self, blobs_in_stream):
+        def get_blobs(blobs):
+            for (blob, _, _, blob_len) in blobs:
+                if blob:
+                    yield self.blob_manager.get_blob(blob, True, blob_len)
+
+        dl = defer.DeferredList(list(get_blobs(blobs_in_stream)), consumeErrors=True)
+        dl.addCallback(lambda blobs: [blob for _, blob in blobs if blob.is_validated()])
```
jobevers: Why not check for success?

lbrynet/reflector/server/server.py
```diff
@@ -11,27 +11,89 @@
...
+        self.protocol_version = self.factory.protocol_version
         self.received_handshake = False
-        self.received_descriptor = False
-        self.protocol_version = REFLECTOR_V2
         self.peer_version = None
         self.receiving_blob = False
         self.incoming_blob = None
         self.blob_write = None
-        self.blob_finished_d = None
+        self.blob_f_d = None
```
jobevers: `f` usually stands for file not finished; generally fail on the side of being verbose.

lbrynet/reflector/server/server.py
```diff
@@ -11,27 +11,89 @@
...
+        self.transport.write(json.dumps(response_dict))
+
+    ############################
+    # Incoming blob file stuff #
+    ############################
+
+    def clean_up_failed_upload(self, err, blob):
+        log.warning("Failed to receive %s", blob.blob_hash[:16])
+        if err.check(DownloadCanceledError):
+            err.trap(DownloadCanceledError)
```
jobevers: you don't need to trap if you're already checking.  Trap is like check except that it re-raises if it not the expected type.

lbrynet/reflector/server/server.py
```diff
@@ -11,27 +11,89 @@
...
+            self.blob_manager.delete_blobs([blob.blob_hash])
+        else:
+            log.exception(err)
+        return False
+
+    def log_received_blob(self, blob):
+        log.info("Received %s", blob.blob_hash[:16])
```
jobevers: Since the __repr__ and __str__ functions of blob already do this (https://github.com/lbryio/lbry/blob/master/lbrynet/core/HashBlob.py#L201)

This should just be log.info('Received %s", blob).

And, think about it - what if we decide that we do want the full hash in the logs?  I don't want to have to change the code in multiple different places.

lbrynet/reflector/server/server.py
```diff
@@ -11,27 +11,89 @@
...
+        """
+
+        blob = self.incoming_blob
+        self.blob_f_d, self.blob_write, self.cancel_write = blob.open_for_writing(self.peer)
+        self.blob_f_d.addCallback(lambda _: self.blob_manager.blob_completed(blob))
+        self.blob_f_d.addCallback(lambda _: self.close_blob())
+        self.blob_f_d.addCallbacks(lambda _: self.log_received_blob(blob),
+                                   lambda err: self.clean_up_failed_upload(err, blob))
+        self.blob_f_d.addCallback(lambda result: {response_key: result})
+        self.blob_f_d.addCallback(self.send_response)
```
jobevers: I think you changed blob_finished_d to blob_f_d because blob_finished_d made this section too long. Instead, try:

```
blob = self.incoming_blob
self.blob_f_d, self.blob_write, self.cancel_write = blob.open_for_writing(self.peer)
self.blob_finished_d.addCallback(self._on_completed_blob, response_key)
```

```
@defer.inlineCallbacks
def _on_completed_blob(self, response_key):
    blob = self.incoming_blob
    try:
        yield self.blob_manager.blob_completed(blob)
        yield self.close_blob()
        log.info("Received %s", blob)
        yield self.send_respond({response_key: True})
    except Exception as e:
        yield self.clean_up_failed_upload(e, blob)
        yield self.send_respond({response_key: False})
```

lbrynet/reflector/server/server.py
```diff
@@ -47,15 +113,15 @@ def _get_valid_response(self, response_msg):
...
+        while not self.receiving_blob:
             next_close_paren = response_msg.find('}', curr_pos)
             if next_close_paren != -1:
                 curr_pos = next_close_paren + 1
                 try:
                     response = json.loads(response_msg[:curr_pos])
                 except ValueError:
-                    if curr_pos > 100:
-                        raise Exception("error decoding response")
+                    if curr_pos > 200:
```
jobevers: why 200?

lbrynet/reflector/server/server.py
```diff
@@ -67,71 +133,169 @@ def _get_valid_response(self, response_msg):
...
-            raise ValueError("Client should send version")
-        self.peer_version = int(request_dict['version'])
-        if self.peer_version != 0:
-            raise ValueError("I don't know that version!")
+            if VERSION not in request_dict:
+                raise ReflectorRequestError("Client should send version")
+            peer_version = int(request_dict[VERSION])
+            return self.handle_handshake(peer_version)
+
+        if SD_BLOB_HASH in request_dict and SD_BLOB_SIZE in request_dict:
```
jobevers: change this to read:

```
if self.is_descriptor_request(request_dict):
    return self.handle_descriptor_request(request_dict)
if self.is_blob_request(request_dict):
    return self.handle_blob_request(request_dict)
```

where `is_descriptor_request` is
```
def is_descriptor_request(request_dict):
    return SD_BLOB_HASH in request_dict and SD_BLOB_SIZE in request_dict
```

etc...
return

tests/functional/test_reflector.py
```diff
@@ -219,6 +219,37 @@ def verify_blob_completed(blob, blob_size):
...
+            for blob_hash, blob_size in self.expected_blobs:
+                check_blob_ds.append(verify_have_blob(blob_hash, blob_size))
+            return defer.DeferredList(check_blob_ds)
+
+        def verify_have_blob(blob_hash, blob_size):
+            d = self.server_blob_manager.get_blob(blob_hash, True)
+            d.addCallback(lambda blob: verify_blob_completed(blob, blob_size))
+            return d
+
+        def send_to_server(blobs_to_send):
```
jobevers: we need to be consistent with our variable name to distinguish between whether an object is a blob_hash string, a HashBlob, etc.  

I think `blobs_to_send` should be `blob_hashes`

lbrynet/reflector/server/server.py
jackrobison: client queries can be more than 100 bytes,
{
     'sd_blob_hash': '48d568facff0400581a1b449b4c624da061d9783063e118764b556c27f99fad3855f74155cf0ff431d3486580e279e13', 
    'sd_blob_size': 2097152
}

is 141

jobevers: Can you change 200 to be `MAXIMUM_QUERY_SIZE` or something similar

jobevers:
One last tiny request, otherwise, LGTM

PR #441, opened 2017-01-25
Better Connection manager and Client Protocol logging

kaykurokawa:
Add better logging ( log.debug's) for ConnectionManager and ClientProtocol (used by ConnectionManager to manage individual connections). 

In general be more verbose, and state what a specific instance of that class if for (what peer its connected to, or what its downloading)

PR #458, opened 2017-02-01
Connection Manager tests 

kaykurokawa:
Add trial tests for Connection Manager, this required some changes to Connection Manager to make it work but overall functionality is unchanged. Below are the list of changes.

The test caught a case where Connection Manager would not report a peer that was returning an unexpected response as being down, and thus it would attempt to reconnect to it later. This has been fixed in this patch. 

**Peer changes-** 

Add a report_success() function to the Peer class so we can figure out how many successful connections we've made to the peer.  Useful in unit testing. Also just added some comments to clarify how the class should be used. 

**ConnectionManager changes -** 

start() function is seperated into _start() and start(). _start() does the initialization and
start() handles the reactor scheduling. This is useful in unit test when we want to initialize without scheduling anything. 

Likewise, manage() function (the main loop of connection manager) is separated into _manage() and manage(). _manage() does the actual work, and manage() handles the reactor scheduling. This is useful when we want to test a single instance of a loop without scheduling anything. 

Instead of importing reactor.callLater , we make callLater a class method so in the unit test we can use clock.callLater. 

_connect_to_peer function, which sets up ClientProtocolFactory, will return a deferred , that fires when a connection has been made or lost. This allows us to unit test the connection to peers. 

num_peer_connections() function will return number of ongoing peer connections. Useful for debugging.

**ClientProtocol changes -** 

ClientProtocolFactory intializes a deferred which gets fired when a connection is made or lost. 

_ask_for_request() was being scheduled as a callLater for no particular good reason  

If the ClientProtocol has finished successfully, call report_success() on Peer. 

If ClientProocol runs into an error while in the middle of a request, call report_down() on Peer. Without this If a ClientProtocol failed due to some non standard behaving peer, ConnnectionManager
would attempt to reconnect to it. 

**Added Tests -**

Test a successful connection to an existing and well behaving server

Test a failed connection to a non existing server

Test a failed connection to a server that returns an improper response.

### DIFF
--- a/lbrynet/core/client/ClientProtocol.py
+++ b/lbrynet/core/client/ClientProtocol.py
@@ -23,7 +23,6 @@ def encode_decimal(obj):
 
 class ClientProtocol(Protocol):
     implements(IRequestSender, IRateLimited)
-
     ######### Protocol #########
 
     def connectionMade(self):
@@ -73,7 +72,7 @@ class ClientProtocol(Protocol):
             d.errback(err)
         if self._blob_download_request is not None:
             self._blob_download_request.cancel(err)
-        self._connection_manager.protocol_disconnected(self.peer, self)
+        self.factory.connection_was_made_deferred.callback(True)
 
     ######### IRequestSender #########
 
@@ -132,8 +131,8 @@ class ClientProtocol(Protocol):
                 log.info(
                     "Closing the connection to %s due to having no further requests to send",
                     self.peer)
+                self.peer.report_success()
                 self.transport.loseConnection()
-
         d = self._connection_manager.get_next_request(self.peer, self)
         d.addCallback(send_request_or_close)
         d.addErrback(self._handle_request_error)
@@ -208,10 +207,11 @@ class ClientProtocol(Protocol):
                         log.info(result.value)
                         log.info("The connection to %s is closing due to an error: %s",
                                  self.peer, result.getTraceback())
+
+                        self.peer.report_down()
             if failed is False:
                 log.debug("Asking for another request from %s", self.peer)
-                from twisted.internet import reactor
-                reactor.callLater(0, self._ask_for_request)
+                self._ask_for_request()
             else:
                 log.debug("Not asking for another request from %s", self.peer)
                 self.transport.loseConnection()
@@ -257,11 +257,15 @@ class ClientProtocolFactory(ClientFactory):
         self.rate_limiter = rate_limiter
         self.connection_manager = connection_manager
         self.p = None
+        # This defer fires and returns True when connection was
+        # made and completed, or fires and returns False if
+        # connection failed
+        self.connection_was_made_deferred = defer.Deferred()
 
     def clientConnectionFailed(self, connector, reason):
         log.debug("Connection failed to %s: %s", self.peer, reason)
         self.peer.report_down()
-        self.connection_manager.protocol_disconnected(self.peer, connector)
+        self.connection_was_made_deferred.callback(False)
 
     def buildProtocol(self, addr):
         p = self.protocol()
--- a/lbrynet/core/client/ConnectionManager.py
+++ b/lbrynet/core/client/ConnectionManager.py
@@ -1,5 +1,5 @@
 import logging
-from twisted.internet import defer
+from twisted.internet import defer, reactor
 from zope.interface import implements
 from lbrynet import interfaces
 from lbrynet import conf
@@ -19,6 +19,8 @@ class PeerConnectionHandler(object):
 
 class ConnectionManager(object):
     implements(interfaces.IConnectionManager)
+    callLater = reactor.callLater
+    MANAGE_CALL_INTERVAL_SEC = 1
 
     def __init__(self, downloader, rate_limiter,
                  primary_request_creators, secondary_request_creators):
@@ -44,16 +46,17 @@ class ConnectionManager(object):
             out = 'Connection Manager '+self.downloader.blob_hash
         return out
 
-    def start(self):
-        log.debug("%s starting", self._get_log_name())
-        from twisted.internet import reactor
-
+    def _start(self):
         self.stopped = False
-
         if self._next_manage_call is not None and self._next_manage_call.active() is True:
             self._next_manage_call.cancel()
-        self._next_manage_call = reactor.callLater(0, self._manage)
+
+    def start(self):
+        log.debug("%s starting", self._get_log_name())
+        self._start()
+        self._next_manage_call = self.callLater(0, self.manage)
         return defer.succeed(True)
+
 
     @defer.inlineCallbacks
     def stop(self):
@@ -68,8 +71,10 @@ class ConnectionManager(object):
         self._next_manage_call = None
         yield self._close_peers()
 
-    def _close_peers(self):
+    def num_peer_connections(self):
+        return len(self._peer_connections)
 
+    def _close_peers(self):
         def disconnect_peer(p):
             d = defer.Deferred()
             self._connections_closing[p] = d
@@ -137,37 +142,21 @@ class ConnectionManager(object):
         ]
         return defer.DeferredList(ds)
 
-    def protocol_disconnected(self, peer, protocol):
-        log.debug("%s protocol %s disconnected for %s",
-                    self._get_log_name(), type(protocol).__name__, peer)
-        if peer in self._peer_connections:
-            del self._peer_connections[peer]
-        if peer in self._connections_closing:
-            d = self._connections_closing[peer]
-            del self._connections_closing[peer]
-            d.callback(True)
-
     @defer.inlineCallbacks
-    def _manage(self):
+    def manage(self, schedule_next_call=True):
         self._manage_deferred = defer.Deferred()
-
-        from twisted.internet import reactor
         if len(self._peer_connections) < conf.settings['max_connections_per_stream']:
             log.debug("%s have %d connections, looking for %d",
                         self._get_log_name(), len(self._peer_connections),
                         conf.settings['max_connections_per_stream'])
-            try:
-                ordered_request_creators = self._rank_request_creator_connections()
-                peers = yield self._get_new_peers(ordered_request_creators)
-                peer = self._pick_best_peer(peers)
-                yield self._connect_to_peer(peer)
-            except Exception:
-                # log this otherwise it will just end up as an unhandled error in deferred
-                log.exception('Something bad happened picking a peer')
+            ordered_request_creators = self._rank_request_creator_connections()
+            peers = yield self._get_new_peers(ordered_request_creators)
+            peer = self._pick_best_peer(peers)
+            self._connect_to_peer(peer)
         self._manage_deferred.callback(None)
         self._manage_deferred = None
-        if not self.stopped:
-            self._next_manage_call = reactor.callLater(1, self._manage)
+        if not self.stopped and schedule_next_call:
+            self._next_manage_call = self.callLater(self.MANAGE_CALL_INTERVAL_SEC, self.manage)
 
     def _rank_request_creator_connections(self):
         """Returns an ordered list of our request creators, ranked according
@@ -213,11 +202,24 @@ class ConnectionManager(object):
         if peer is None or self.stopped:
             return
 
-        from twisted.internet import reactor
-
         log.debug("%s Trying to connect to %s", self._get_log_name(), peer)
         factory = ClientProtocolFactory(peer, self.rate_limiter, self)
+        factory.connection_was_made_deferred.addCallback(
+                lambda c_was_made: self._peer_disconnected(c_was_made, peer))
         self._peer_connections[peer] = PeerConnectionHandler(self._primary_request_creators[:],
                                                              factory)
         connection = reactor.connectTCP(peer.host, peer.port, factory)
         self._peer_connections[peer].connection = connection
+
+    def _peer_disconnected(self, connection_was_made, peer):
+        log.debug("%s protocol disconnected for %s",
+                    self._get_log_name(), peer)
+        if peer in self._peer_connections:
+            del self._peer_connections[peer]
+        if peer in self._connections_closing:
+            d = self._connections_closing[peer]
+            del self._connections_closing[peer]
+            d.callback(True)
+        return connection_was_made
+
+
--- a/tests/unit/core/client/test_ConnectionManager.py
+++ b/tests/unit/core/client/test_ConnectionManager.py
@@ -1,0 +1,214 @@
+import sys
+import time
+import logging
+
+from lbrynet.core import log_support
+from lbrynet.core.client.ConnectionManager import ConnectionManager
+from lbrynet.core.client.ClientRequest import ClientRequest
+from lbrynet.core.server.ServerProtocol import ServerProtocol
+from lbrynet.core.RateLimiter import RateLimiter
+from lbrynet.core.Peer import Peer
+from lbrynet.core.PeerManager import PeerManager
+from lbrynet.core.Error import ConnectionClosedBeforeResponseError, NoResponseError
+
+from twisted.trial import unittest
+from twisted.internet import defer, reactor, task
+from twisted.internet.task import deferLater
+from twisted.internet.protocol import Protocol, ServerFactory
+from lbrynet import conf
+from lbrynet.interfaces  import IQueryHandlerFactory, IQueryHandler, IRequestCreator
+
+from zope.interface import implements
+
+PEER_PORT = 5551
+LOCAL_HOST = '127.0.0.1'
+
+class MocDownloader(object):
+    def insufficient_funds(self):
+        pass
+
+class MocRequestCreator(object):
+    implements(IRequestCreator)
+    def __init__(self, peers_to_return):
+        self.peers_to_return = peers_to_return
+        self.sent_request = False
+
+    def send_next_request(self, peer, protocol):
+        if self.sent_request is True:
+            return defer.succeed(False)
+        response_identifier = 'moc_request'
+        r_dict = {'moc_request':0}
+        request = ClientRequest(r_dict, response_identifier)
+        d = protocol.add_request(request) # ClientRequest here
+        d.addErrback(self.request_err, peer)
+        d.addCallback(self.request_success)
+        self.sent_request = True
+        return defer.succeed(True)
+
+    def request_success(self, suc):
+        pass
+
+    def request_err(self, err, peer):
+        if isinstance(err.value, NoResponseError):
+            return err
+
+    def get_new_peers(self):
+        return self.peers_to_return
+
+class MocFunctionalQueryHandler(object):
+    implements(IQueryHandler)
+
+    def __init__(self, clock, is_good=True, is_delayed=False):
+        self.query_identifiers = ['moc_request']
+        self.is_good = is_good
+        self.is_delayed = is_delayed
+        self.clock = clock
+
+    def register_with_request_handler(self, request_handler, peer):
+        request_handler.register_query_handler(self, self.query_identifiers)
+
+    def handle_queries(self, queries):
+        if self.query_identifiers[0] in queries:
+            if self.is_delayed:
+                out = deferLater(self.clock, 10, lambda: {'moc_request':0})
+                self.clock.advance(10)
+                return out
+            if self.is_good:
+                return defer.succeed({'moc_request':0})
+            else:
+                return defer.succeed({'bad_request':0})
+        else:
+            return defer.succeed({})
+
+
+class MocQueryHandlerFactory(object):
+    implements(IQueryHandlerFactory)
+    # is is_good, the query handler works as expectd,
+    # is is_delayed, the query handler will delay its resposne
+    def __init__(self, clock, is_good=True, is_delayed=False):
+        self.is_good = is_good
+        self.is_delayed = is_delayed
+        self.clock = clock
+    def build_query_handler(self):
+        return MocFunctionalQueryHandler(self.clock, self.is_good, self.is_delayed)
+
+    def get_primary_query_identifier(self):
+        return 'moc_query'
+
+    def get_description(self):
+        return "This is a Moc Query"
+
+
+class MocServerProtocolFactory(ServerFactory):
+    protocol = ServerProtocol
+    def __init__(self, clock, is_good=True, is_delayed=False, has_moc_query_handler=True):
+        self.rate_limiter = RateLimiter()
+        query_handler_factory = MocQueryHandlerFactory(clock, is_good, is_delayed)
+        if has_moc_query_handler:
+            self.query_handler_factories = {
+                query_handler_factory.get_primary_query_identifier():query_handler_factory
+            }
+        else:
+            self.query_handler_factories = {}
+        self.peer_manager = PeerManager()
+
+
+class TestIntegrationConnectionManager(unittest.TestCase):
+    def setUp(self):
+
+        conf.initialize_settings()
+
+        self.TEST_PEER = Peer(LOCAL_HOST, PEER_PORT)
+        self.downloader = MocDownloader()
+        self.rate_limiter = RateLimiter()
+        self.primary_request_creator = MocRequestCreator([self.TEST_PEER])
+        self.connection_manager = ConnectionManager(self.downloader, self.rate_limiter,
+                                                    [self.primary_request_creator], [])
+
+        self.clock = task.Clock()
+        self.connection_manager.callLater = self.clock.callLater
+        self.connection_manager._start()
+        self.server_port = None
+
+    def tearDown(self):
+        if self.server_port is not None:
+            self.server_port.stopListening()
+        self.connection_manager.stop()
+        conf.settings = None
+
+    @defer.inlineCallbacks
+    def test_success(self):
+        # test to see that if we setup a server, we get a connection
+        self.server = MocServerProtocolFactory(self.clock)
+        self.server_port = reactor.listenTCP(PEER_PORT, self.server, interface=LOCAL_HOST)
+        yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        connection_made = yield self.connection_manager._peer_connections[self.TEST_PEER].factory.connection_was_made_deferred
+        self.assertEqual(0, self.connection_manager.num_peer_connections())
+        self.assertTrue(connection_made)
+        self.assertEqual(1, self.TEST_PEER.success_count)
+        self.assertEqual(0, self.TEST_PEER.down_count)
+
+    @defer.inlineCallbacks
+    def test_bad_server(self):
+        # test to see that if we setup a server that returns an improper reply
+        # we don't get a connection
+        self.server = MocServerProtocolFactory(self.clock, is_good=False)
+        self.server_port = reactor.listenTCP(PEER_PORT, self.server, interface=LOCAL_HOST)
+        yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        connection_made = yield self.connection_manager._peer_connections[self.TEST_PEER].factory.connection_was_made_deferred
+        self.assertEqual(0, self.connection_manager.num_peer_connections())
+        self.assertTrue(connection_made)
+        self.assertEqual(0, self.TEST_PEER.success_count)
+        self.assertEqual(1, self.TEST_PEER.down_count)
+
+    @defer.inlineCallbacks
+    def test_non_existing_server(self):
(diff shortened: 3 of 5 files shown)


lbrynet/core/client/ClientProtocol.py
```diff
@@ -132,8 +132,8 @@ def send_request_or_close(do_request):
                 log.info(
                     "Closing the connection to %s due to having no further requests to send",
                     self.peer)
+                self.peer.report_success()
```
kaykurokawa: this allows us to know that a peer connection and request/reply to it was  completed and successful

lbrynet/core/client/ClientProtocol.py
```diff
@@ -208,10 +208,11 @@ def get_next_request(results):
                         log.info(result.value)
                         log.info("The connection to %s is closing due to an error: %s",
                                  self.peer, result.getTraceback())
+
+                        self.peer.report_down()
```
kaykurokawa: without this, connection manager would attempt to reconnect to a peer that was returning an improper response

jobevers:
> Add trial tests for Connection Manager, this required some changes to Connection Manager to make it work but overall functionality is unchanged. Below are the list of changes.

Testing usually does and generally I find the parts changed to be better designed; one of the reasons I like testing.

lbrynet/core/client/ConnectionManager.py
```diff
@@ -213,11 +218,10 @@ def _connect_to_peer(self, peer):
...
 
-        from twisted.internet import reactor
-
         log.debug("%s Trying to connect to %s", self._get_log_name(), peer)
         factory = ClientProtocolFactory(peer, self.rate_limiter, self)
         self._peer_connections[peer] = PeerConnectionHandler(self._primary_request_creators[:],
                                                              factory)
         connection = reactor.connectTCP(peer.host, peer.port, factory)
         self._peer_connections[peer].connection = connection
+        return factory.deferred
```
jobevers: can you document what this returns and when the deferred that is returned gets fired?

It will save whoever is reading this code next from having to dig through a bunch of code.

lbrynet/core/client/ConnectionManager.py
```diff
@@ -147,27 +152,27 @@ def protocol_disconnected(self, peer, protocol):
...
                         self._get_log_name(), len(self._peer_connections),
                         conf.settings['max_connections_per_stream'])
+            ordered_request_creators = self._rank_request_creator_connections()
+            peers = yield self._get_new_peers(ordered_request_creators)
+            peer = self._pick_best_peer(peers)
             try:
-                ordered_request_creators = self._rank_request_creator_connections()
-                peers = yield self._get_new_peers(ordered_request_creators)
-                peer = self._pick_best_peer(peers)
                 yield self._connect_to_peer(peer)
```
jobevers: Due to the change in _connect_to_peer to return `factory.deferred` (that is call-backed when the connection is lost), this changes this code from only yielding until the peer is connected to to waiting until the peer connection is lost.

I don't think this is desirable as it will limit the number of parallel connections we have.

lbrynet/core/client/ConnectionManager.py
```diff
@@ -44,17 +46,18 @@ def _get_log_name(self):
             out = 'Connection Manager '+self.downloader.blob_hash
         return out
 
-    def start(self):
-        log.debug("%s starting", self._get_log_name())
-        from twisted.internet import reactor
-
+    def _start(self):
```
jobevers: this function name doesn't match what it does - not sure what a better name would be, but worth thinking about.

lbrynet/core/client/ConnectionManager.py
```diff
@@ -147,27 +152,27 @@ def protocol_disconnected(self, peer, protocol):
...
-                peer = self._pick_best_peer(peers)
                 yield self._connect_to_peer(peer)
             except Exception:
                 # log this otherwise it will just end up as an unhandled error in deferred
-                log.exception('Something bad happened picking a peer')
-        self._manage_deferred.callback(None)
-        self._manage_deferred = None
-        if not self.stopped:
-            self._next_manage_call = reactor.callLater(1, self._manage)
+                # Can happen if connection fails with ConnectionRefusedError
```
jobevers: can connectionrefusederror be caught and better handled?

Generally, blanket `except Exception` are an action of last resort (or laziness - in this case, mine) and if there is a more specific error that can be caught and handled, we should do that.  Maybe:

```
try:
    yield self._connect_to_peer(peer)
except ConnectionRefusedError:
    # do something with the peer, like peer.report_down() ?
except Exception:
    log.exception('Something bad happened connecting to a peer')
```

lbrynet/core/client/ConnectionManager.py
```diff
@@ -147,27 +152,27 @@ def protocol_disconnected(self, peer, protocol):
...
-        self._manage_deferred = defer.Deferred()
-
-        from twisted.internet import reactor
         if len(self._peer_connections) < conf.settings['max_connections_per_stream']:
             log.debug("%s have %d connections, looking for %d",
                         self._get_log_name(), len(self._peer_connections),
                         conf.settings['max_connections_per_stream'])
+            ordered_request_creators = self._rank_request_creator_connections()
+            peers = yield self._get_new_peers(ordered_request_creators)
+            peer = self._pick_best_peer(peers)
```
jobevers: are you certain that this code won't throw an exception?

I'm pretty sure that an exception here will just disappear into the ether, as indicated in the comment
`# log this otherwise it will just end up as an unhandled error in deferred`

lbrynet/core/client/ConnectionManager.py
```diff
@@ -147,27 +152,27 @@ def protocol_disconnected(self, peer, protocol):
...
             try:
-                ordered_request_creators = self._rank_request_creator_connections()
-                peers = yield self._get_new_peers(ordered_request_creators)
-                peer = self._pick_best_peer(peers)
                 yield self._connect_to_peer(peer)
             except Exception:
                 # log this otherwise it will just end up as an unhandled error in deferred
-                log.exception('Something bad happened picking a peer')
-        self._manage_deferred.callback(None)
-        self._manage_deferred = None
```
jobevers: See https://github.com/lbryio/lbry/pull/458#discussion_r98992954

lbrynet/core/client/ClientProtocol.py
```diff
@@ -74,6 +73,7 @@ def connectionLost(self, reason):
         if self._blob_download_request is not None:
             self._blob_download_request.cancel(err)
         self._connection_manager.protocol_disconnected(self.peer, self)
+        self.factory.deferred.callback(True)
```
kaykurokawa: self.factory.deferred is a deferred that is supposed to fire when a client connection has completely terminated (either it failed to connect, or succeeded to connect). The other places where this deferred gets fired is clientConnectionFailed() under ClientProtocolFactory. In the unit test we wait for this deferred to check if a ClientProtocol connection was sucessful or not.

lbrynet/core/client/ClientProtocol.py
```diff
@@ -208,10 +208,11 @@ def get_next_request(results):
                         log.info(result.value)
                         log.info("The connection to %s is closing due to an error: %s",
                                  self.peer, result.getTraceback())
+
+                        self.peer.report_down()
             if failed is False:
                 log.debug("Asking for another request from %s", self.peer)
-                from twisted.internet import reactor
-                reactor.callLater(0, self._ask_for_request)
+                self._ask_for_request()
```
kaykurokawa: but than why not call reactor.callLater(0,.) for everything and why this specific function? 

I think _ask_for_request() will obtain the next request for downloading (should be fast) and call transport.write() which is supposed to be a non blocking call right?

lbrynet/core/client/ConnectionManager.py
```diff
@@ -19,6 +19,8 @@ def __init__(self, request_creators, factory):
 
 class ConnectionManager(object):
     implements(interfaces.IConnectionManager)
+    callLater = reactor.callLater
```
kaykurokawa: I think that's good , in tests you need to be able to set callLater for all classes that uses it, and it would be better if you could just set a single function.  
I think this would be a seprate pull request, I see a lot callLaters in lbrynet that needs to be changed.

lbrynet/core/client/ConnectionManager.py
kaykurokawa: yea I'll check all the exceptions that can be returned here , and see if it needs to be handled

lbrynet/core/client/ConnectionManager.py
kaykurokawa: I see

lbrynet/core/client/ConnectionManager.py
kaykurokawa: OK, so I think what should happen is that it should yield until it found a peer to connect to and started to make a connection. That's what it was doing before ( not yielding until the peer is connected to ). Now this patch is yielding until a connection is fully terminated which you are right will limit the number of parallel connections.

lbrynet/core/client/ConnectionManager.py
```diff
@@ -147,27 +152,27 @@ def protocol_disconnected(self, peer, protocol):
             del self._connections_closing[peer]
             d.callback(True)
 
+    def manage(self):
```
kaykurokawa: sounds good.

lbrynet/core/client/ConnectionManager.py
kaykurokawa: I believe this exception catching is no longer needed. When connection fails (replicated in non existing server unit test),  ClientProtocolFactory.clientConnectionFailed() is called with reason ConnectionRefusedError. Before there was no logging information placed there (added inhttps://github.com/lbryio/lbry/pull/441), so this exception catching handled the logging.

lbrynet/core/client/ClientProtocol.py
```diff
@@ -258,13 +257,15 @@ def __init__(self, peer, rate_limiter, connection_manager):
         self.rate_limiter = rate_limiter
         self.connection_manager = connection_manager
         self.p = None
-        self.deferred = defer.Deferred()
+        # This defer fires and returns True when connection was
+        # made and completed, or fires and returns False if
+        # connection failed
+        self.connection_was_made_deferred = defer.Deferred()
```
kaykurokawa: a more useful name for the deferred, and comments on what it does

kaykurokawa: made a more useful name for the deferred, and comments on what it does

lbrynet/core/client/ClientProtocol.py
```diff
@@ -258,13 +257,15 @@ def __init__(self, peer, rate_limiter, connection_manager):
...
 
     def clientConnectionFailed(self, connector, reason):
         log.debug("Connection failed to %s: %s", self.peer, reason)
         self.peer.report_down()
```
kaykurokawa: Instead of firing a errback, we fire a callback and return False

lbrynet/core/client/ConnectionManager.py
```diff
@@ -142,37 +142,21 @@ def _send_secondary_requests(self, peer, protocol):
         ]
         return defer.DeferredList(ds)
```
kaykurokawa: protocol_disconnected() renamed to _peer_disconnected(), and does not require protocol as argument. ( only used for debugging here, and its always going to be ClientProtocol anways)

lbrynet/core/client/ConnectionManager.py
```diff
@@ -142,37 +142,21 @@ def _send_secondary_requests(self, peer, protocol):
...
-            del self._connections_closing[peer]
-            d.callback(True)
-
-    def manage(self):
-        self._manage_deferred = self._manage()
-        if not self.stopped:
-            self._next_manage_call = self.callLater(self.MANAGE_CALL_INTERVAL_SEC, self.manage)
-
-
     @defer.inlineCallbacks
```
kaykurokawa: Instead of using manage() and _manage() so that we can call _manage() in unit tests without scheduling. I added an argument to manage(), schedule_next_call that can be set to False for unit testing and is by default True

lbrynet/core/client/ConnectionManager.py
```diff
@@ -142,37 +142,21 @@ def _send_secondary_requests(self, peer, protocol):
...
             ordered_request_creators = self._rank_request_creator_connections()
             peers = yield self._get_new_peers(ordered_request_creators)
             peer = self._pick_best_peer(peers)
-            try:
-                yield self._connect_to_peer(peer)
-            except Exception:
-                # log this otherwise it will just end up as an unhandled error in deferred
-                # Can happen if connection fails with ConnectionRefusedError
-                log.exception('Something bad happened connecting to a peer')
+            self._connect_to_peer(peer)
```
kaykurokawa: make sure we don't yield here ( there is nothing to yield here now anyways, as connec_to_peer does not return deferred now

lbrynet/core/client/ConnectionManager.py
```diff
@@ -142,37 +142,21 @@ def _send_secondary_requests(self, peer, protocol):
...
-    def _manage(self):
+    def manage(self, schedule_next_call=True):
+        self._manage_deferred = defer.Deferred()
         if len(self._peer_connections) < conf.settings['max_connections_per_stream']:
             log.debug("%s have %d connections, looking for %d",
                         self._get_log_name(), len(self._peer_connections),
                         conf.settings['max_connections_per_stream'])
             ordered_request_creators = self._rank_request_creator_connections()
             peers = yield self._get_new_peers(ordered_request_creators)
             peer = self._pick_best_peer(peers)
```
kaykurokawa: remove the exception catching here, all exceptions should be caught and logged by ClientProtocolFactory.clientConnectionFailed()

lbrynet/core/client/ConnectionManager.py
```diff
@@ -220,8 +204,22 @@ def _connect_to_peer(self, peer):
 
         log.debug("%s Trying to connect to %s", self._get_log_name(), peer)
         factory = ClientProtocolFactory(peer, self.rate_limiter, self)
```
kaykurokawa: we attach a callback to the connection_was_made deferred that calls _peer_disconnected() , now ClientProtocol no longer needs to call disconnect peers in the ConnectionManager. This is good, because we need to reduce and eventually remove ClientProtocol's dependency on ConnectionManager.

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -12,14 +12,13 @@
 from lbrynet.core.Error import ConnectionClosedBeforeResponseError, NoResponseError
```
kaykurokawa: Ran pylint on tests

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -161,28 +162,74 @@ def tearDown(self):
...
-        d = yield self.connection_manager._manage()
+        d = yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        d = yield self.connection_manager._peer_connections[self.TEST_PEER].factory.connection_was_made_deferred
+        self.assertEqual(0, self.connection_manager.num_peer_connections())
+        self.assertEqual(False, d)
+        self.assertEqual(0, self.connection_manager.num_peer_connections())
+        self.assertEqual(0, self.TEST_PEER.success_count)
+        self.assertEqual(1, self.TEST_PEER.down_count)
+
```
kaykurokawa: Added tests to see if multiple parallel connections can be started using the manage() function

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -161,28 +162,74 @@ def tearDown(self):
...
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        self.assertEqual(0, self.TEST_PEER.success_count)
+        self.assertEqual(1, self.TEST_PEER.down_count)
+        d = yield self.connection_manager._peer_connections[test_peer2].factory.connection_was_made_deferred
+        self.assertEqual(False,d)
+        self.assertEqual(0, self.connection_manager.num_peer_connections())
+        self.assertEqual(0, test_peer2.success_count)
+        self.assertEqual(1, test_peer2.down_count)
+
+
```
kaykurokawa: Added test for the  stop() function

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -12,14 +12,13 @@
...
+from twisted.internet import defer, reactor, task
 from twisted.internet.task import deferLater
-from twisted.internet.protocol import Protocol,ServerFactory
+from twisted.internet.protocol import Protocol, ServerFactory
 from lbrynet import conf
-from lbrynet.interfaces  import IQueryHandlerFactory,IQueryHandler,IRequestCreator
+from lbrynet.interfaces  import IQueryHandlerFactory, IQueryHandler, IRequestCreator
 
 from zope.interface import implements
-conf.initialize_settings()
```
kaykurokawa: moved into setUp()

kaykurokawa:
Ok, new patch here based on your review, here is summary of the changes. 

Original Connection Manager, had no deferred that fired when connection was completed - 
	a) so there was no way to write a test in trial to check the state of connection manager after a finished connection
	b) this also meant that ClientProtocol had to be resposible for calling ConnectionManager.protocol_disconnect() when connection was finished ,  this is bad because ClientProtocol, which represents a single connection to a peer, should not need to know anything about ConnectionManager. We should be able to setup a ClientProtocol instance without setting up a ConnectionManager. In this current patch, now that we have this deferred in ClientProtocolFactory, ConnectionManager can attach a callback to it that cleans itself up after the connection has finished, and ClientProtocol does not need to call ConnectionManager.protocol_disconnect(). protocol_disconnect() has also been renamed to peer_disconnect(). 

So with previous patch, I fired the deferred callback when connection was made and completed, and fired the errback when connection failed. But now in this current patch, I will fire the deferred callback with return value True when connection was made and completed, and fire the deferred callback with return value False wen connection failed. This makes it easier to unit test the connection. 

The previous patch I made returned the deferred in the manage() function , but this was not necessary, and not correct as it was preventing parallel connections. The current patch, the deferred is not returned but stored in the ClientProtocolFactory, for unit testing we can acces this deferred through ConnectionManager._peer_connection[PEER].factory  

The manage loop should do the below in order. 

a) pick and find best peer to connect to 
b) setup the connection, but do not wait for the connection 
c) schedule the next manage call

The previous patch I made did c) without waiting for a) and b) so this was wrong. This new patch should fix this and I added a test to check to see that the manage() loop does what I described above. 

Also added a test to check to see that the stop() function cleans up the ConnectionManager properly (stop manage() from being scheduled, closing connections) and does not break https://github.com/lbryio/lbry/pull/458/files#diff-528ba4be45fedee1b75ecdc2f9c8719eL63

lbrynet/core/client/ClientProtocol.py
jobevers: Looking through the code again - you're right. I got confused and thought there was a loop between ask_for_request and handle_request.

jobevers:
Thanks.  I'll take a look

lbrynet/core/client/ConnectionManager.py
```diff
@@ -137,37 +142,21 @@ def _send_secondary_requests(self, peer, protocol):
...
-                peers = yield self._get_new_peers(ordered_request_creators)
-                peer = self._pick_best_peer(peers)
-                yield self._connect_to_peer(peer)
-            except Exception:
-                # log this otherwise it will just end up as an unhandled error in deferred
-                log.exception('Something bad happened picking a peer')
+            ordered_request_creators = self._rank_request_creator_connections()
+            peers = yield self._get_new_peers(ordered_request_creators)
+            peer = self._pick_best_peer(peers)
+            self._connect_to_peer(peer)
```
jobevers: You lost the error handling around _connect_to_peer

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertEqual(0, self.TEST_PEER.success_count)
+        self.assertEqual(1, self.TEST_PEER.down_count)
+
+    @defer.inlineCallbacks
+    def test_parallel_connections(self):
+        # Test to see that we make new connection for each manage call,
+        # without it waiting for the connection to complete
+        test_peer2 = Peer(LOCAL_HOST, PEER_PORT+1)
+        self.primary_request_creator.peers_to_return = [self.TEST_PEER, test_peer2]
+        d = yield self.connection_manager.manage(schedule_next_call=False)
```
jobevers: the variable `d` isn't right here as we use `d` to represent a deferred.  The yield resolves the deferred and actually gives back the result.  For this line, you aren't actually using the variable so you can just ignore it and use `yield self.connection_manager.manage(schedule_next_call=False)`

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        d = yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(2, self.connection_manager.num_peer_connections())
+        self.assertTrue(self.TEST_PEER in self.connection_manager._peer_connections)
```
jobevers: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIn

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertTrue(test_peer2 in self.connection_manager._peer_connections)
+        d = yield self.connection_manager._peer_connections[self.TEST_PEER].factory.connection_was_made_deferred
+        self.assertEqual(False,d)
```
jobevers: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertFalse

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        # without it waiting for the connection to complete
+        test_peer2 = Peer(LOCAL_HOST, PEER_PORT+1)
+        self.primary_request_creator.peers_to_return = [self.TEST_PEER, test_peer2]
+        d = yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        d = yield self.connection_manager.manage(schedule_next_call=False)
+        self.assertEqual(2, self.connection_manager.num_peer_connections())
+        self.assertTrue(self.TEST_PEER in self.connection_manager._peer_connections)
+        self.assertTrue(test_peer2 in self.connection_manager._peer_connections)
+        d = yield self.connection_manager._peer_connections[self.TEST_PEER].factory.connection_was_made_deferred
```
jobevers: here `d` should be a useful variable name, like, `conn_was_made`

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertEqual(False,d)
+        self.assertEqual(1, self.connection_manager.num_peer_connections())
+        self.assertEqual(0, self.TEST_PEER.success_count)
+        self.assertEqual(1, self.TEST_PEER.down_count)
+        d = yield self.connection_manager._peer_connections[test_peer2].factory.connection_was_made_deferred
```
jobevers: here `d` should be a useful variable name, like, `peer_2_conn_was_made`

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertEqual(1, test_peer2.down_count)
+
+
+    @defer.inlineCallbacks
+    def test_stop(self):
+        # test to see that when we call stop, the ConnectionManager waits for the
+        # current manage call to finish, closes connections,
+        # and removes scheduled manage calls
+        self.connection_manager.manage(schedule_next_call=True)
+        d = yield self.connection_manager.stop()
```
jobevers: can remove `d =`

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+        self.primary_request_creator = MocRequestCreator([self.TEST_PEER])
+        self.connection_manager = ConnectionManager(self.downloader, self.rate_limiter,
+                                                    [self.primary_request_creator], [])
+
+        self.clock = task.Clock()
+        self.connection_manager.callLater = self.clock.callLater
+        self.connection_manager._start()
+        self.server_port = None
+
+    def tearDown(self):
```
jobevers: add in a `conf.settings = None` here so that the settings created in `setUp` don't affect other tests

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+
+class MocQueryHandlerFactory(object):
+    implements(IQueryHandlerFactory)
+    # is is_good, the query handler works as expectd,
+    # is is_delayed, the query handler will delay its resposne
+    def __init__(self, clock, is_good=True, is_delayed=False):
+        self.is_good = is_good
+        self.is_delayed = is_delayed
+        self.clock = clock
+    def build_query_handler(self):
```
jobevers: nit: have a blank line between function definitions

tests/unit/core/client/test_ConnectionManager.py
```diff
@@ -0,0 +1,235 @@
...
+
+# Set up logging if debugging is necessary for writing more tests
+LOG_TO_CONSOLE = False
+if LOG_TO_CONSOLE:
+    handler = logging.StreamHandler(sys.stdout)
+    logging.basicConfig()
+    logger = logging.getLogger()
+    logger.setLevel(0)
+    log_support.configure_console()
+    log_support.configure_twisted()
```
jobevers: out of scope for this PR, but we need a better way to configure logging for the tests as this snippet of code can't be on each test.  Maybe worth looking into whether trial has a way to configure logging.

jobevers:
Made some new comments; largely cosmetic. 
Can you also add an entry to CHANGELOG.md succinctly summarizing the changes.

Feel free to merge when you're ready.

tests/unit/core/client/test_ConnectionManager.py
jobevers: I commented on a couple of cases, but not all.  Can you go through and fix all of the `d = yield ...` instances?

jobevers:
Closing to re-open to retrigger a travis build of the PR

lbrynet/core/client/ConnectionManager.py
kaykurokawa: I believe this exception catching is no longer needed. When connection fails (replicated in non existing server unit test), ClientProtocolFactory.clientConnectionFailed() is called. Before there was no logging information placed in clientConnectionFailed (added in https://github.com/lbryio/lbry/pull/441),  so the exception here would handle the catching and logging.   Now that we have 441, we don't need to catch and log here.

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmitive

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: affirmative

tests/unit/core/client/test_ConnectionManager.py
kaykurokawa: yea probably need a test runner script that can configure logging

kaykurokawa:
added changes to unit tests based on latest review
added changelog
rebased to master

kaykurokawa:
missed setting conf.settings to None for unit test tear down

kaykurokawa:
rebased to master as a single commit
merging to master now

PR #465, opened 2017-02-08
Adjust hash reannounce based on queue size , test for DHTHashAnnouncer

kaykurokawa:
This patch fixes https://github.com/lbryio/lbry/issues/432

It was discovered that after running a reflector node with 16000 hashes, the hash queue (hashes waiting to be announced) was growing at a very fast rate. After 18 hours of running, the hash queue had about 160,000 hashes (see below debug lines from reflector4). 

2017-02-08 00:04:00,459 DEBUG    lbrynet.core.server.DHTHashAnnouncer:67: There are now 14 hashes remaining to be announced
2017-02-08 18:41:07,384 DEBUG    lbrynet.core.server.DHTHashAnnouncer:67: There are now 164043 hashes remaining to be announced

Since each hash takes about half a second to announce, this means it would take 80,000 seconds (or about 22 hours) for any new hash that got put into the queue to be announced , and the condition would obviously worsen after running the node for longer. So basically the large hash queue size was preventing newly published blobs from being re announced until many days later, depending on how long the node had been running.

This patch makes the hash reannounce time adjustable based on the queue size. The default reannounce time is 1 hour, but if it detects that it would take more than 1 hour to announce the entire hash queue, it sets the reannounce time to a conservative estimate of how long it would take to clear the queue ( it assumes it takes 1 second to announce a single hash ). Note that if a node contains 16000 hashes, the reannounce time will be on average 16000 seconds or about 4.4 hours.

We also add a very basic unit test for DHTHashAnnouncer, that should be expanded. 

Some variables in DHTHashAnnouncer and DHTHashSupplier that should be constants, have been turned into constants. 

I'm currently running this on reflector4 to test this patch.

### DIFF
--- a/lbrynet/core/server/DHTHashAnnouncer.py
+++ b/lbrynet/core/server/DHTHashAnnouncer.py
@@ -39,7 +39,7 @@ class DHTHashAnnouncer(object):
 
     def immediate_announce(self, blob_hashes):
         if self.peer_port is not None:
-            return self._announce_hashes(blob_hashes)
+            return self._announce_hashes(blob_hashes, immediate=True)
         else:
             return defer.succeed(False)
 
@@ -56,7 +56,7 @@ class DHTHashAnnouncer(object):
         dl = defer.DeferredList(ds)
         return dl
 
-    def _announce_hashes(self, hashes):
+    def _announce_hashes(self, hashes, immediate=False):
         if not hashes:
             return
         log.debug('Announcing %s hashes', len(hashes))
@@ -67,7 +67,10 @@ class DHTHashAnnouncer(object):
         for h in hashes:
             announce_deferred = defer.Deferred()
             ds.append(announce_deferred)
-            self.hash_queue.append((h, announce_deferred))
+            if immediate:
+                self.hash_queue.appendleft((h, announce_deferred))
+            else:
+                self.hash_queue.append((h, announce_deferred))
         log.debug('There are now %s hashes remaining to be announced', self.hash_queue_size())
 
         def announce():
--- a/lbrynet/core/server/DHTHashAnnouncer.py
+++ b/lbrynet/core/server/DHTHashAnnouncer.py
@@ -10,6 +10,10 @@ log = logging.getLogger(__name__)
 
 
 class DHTHashAnnouncer(object):
+    callLater = reactor.callLater
+    ANNOUNCE_CHECK_INTERVAL = 60
+    CONCURRENT_ANNOUNCERS = 5
+
     """This class announces to the DHT that this peer has certain blobs"""
     def __init__(self, dht_node, peer_port):
         self.dht_node = dht_node
@@ -20,12 +24,9 @@ class DHTHashAnnouncer(object):
         self._concurrent_announcers = 0
 
     def run_manage_loop(self):
-
-        from twisted.internet import reactor
-
         if self.peer_port is not None:
             self._announce_available_hashes()
-        self.next_manage_call = reactor.callLater(60, self.run_manage_loop)
+        self.next_manage_call = self.callLater(self.ANNOUNCE_CHECK_INTERVAL, self.run_manage_loop)
 
     def stop(self):
         log.info("Stopping %s", self)
@@ -41,6 +42,9 @@ class DHTHashAnnouncer(object):
             return self._announce_hashes(blob_hashes)
         else:
             return defer.succeed(False)
+
+    def hash_queue_size(self):
+        return len(self.hash_queue)
 
     def _announce_available_hashes(self):
         log.debug('Announcing available hashes')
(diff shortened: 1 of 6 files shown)


lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -10,6 +10,10 @@
 
 
 class DHTHashAnnouncer(object):
+    callLater = reactor.callLater
```
jobevers: can you do the utils.callLater change that we talked about earlier?

lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -87,12 +90,32 @@ def announce():
...
     def __init__(self, announcer):
         if announcer is not None:
             announcer.add_supplier(self)
         self.hash_announcer = announcer
-        self.hash_reannounce_time = 60 * 60  # 1 hour
 
     def hashes_to_announce(self):
         pass
+
+    """
```
jobevers: put this comment in the docstring under the function

tests/unit/core/server/test_DHTHashAnnouncer.py
```diff
@@ -0,0 +1,50 @@
+from lbrynet.core.server.DHTHashAnnouncer import DHTHashAnnouncer,DHTHashSupplier
+from lbrynet.core.utils import random_string
+from lbrynet.core import log_support
+from twisted.trial import unittest
+from twisted.internet import defer,task
+import logging
+import sys
+import os,binascii
```
jobevers: import should be ordered by

```
system

third party

local
```

and alphabetical within each grouping.  Don't have multiple imports on the same line.  Instead do
```
import binascii
import os
```

lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -87,12 +90,32 @@ def announce():
...
+    unless we are announcing a lot of hashes at once which could cause the
+    the announce queue to pile up.  To prevent pile up, reannounce
+    only after a conservative estimate of when it will finish
+    to announce all the hashes.
+    """
+    def get_next_announce_time(self, num_hashes_to_announce=1):
+        queue_size = self.hash_announcer.hash_queue_size()+num_hashes_to_announce
+        reannounce = max(self.MIN_HASH_REANNOUNCE_TIME,
+                            queue_size*self.SINGLE_HASH_ANNOUNCE_DURATION)
+        return time.time() + reannounce
```
jobevers: I'm not convinced that this is a totally effective solution as it largely moves the queue from being in the queue data structure to being in the database.  The main difference is that the database will only have on entry for each blob so you decrease the chances of having repeated items in the queue - which is an improvement.

A better long term solution would be something along the lines of https://github.com/lbryio/lbry/pull/435/files - which is to record when a blob was last announced (instead of when to next announce it) and have a data structure that doesn't allow for repeated blobs.  As that requires a more substantial change, I'm okay with putting this in place for now but we should follow up with a more complete fix.

jobevers:
Question from tech meeting - can we prepend newly reflected blobs to the the queue instead of appending them.

kaykurokawa:
I don't see why not, we are using a deque .

lbrynet/core/server/DHTHashAnnouncer.py
kaykurokawa: Yea this patch is pretty bad in that we are increasing the dependency between BlobManager and DHTHashAnnouncer. BlobManager should not really need to handle announce time related logic at all except maybe to read and write the database entry for it. I agree that your patch would be be the direction to go next.

kaykurokawa:
Rebased to master 

Added immediate announce option, (when calling immediate_announce, the hash gets added to the front of the queue, instead of the back) 

Fixed doc string formatting, fixed unit test import formatting

Added changelog

lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -56,7 +56,7 @@ def _announce_available_hashes(self):
         dl = defer.DeferredList(ds)
         return dl
 
-    def _announce_hashes(self, hashes):
+    def _announce_hashes(self, hashes, immediate=False):
```
jobevers: 👍

kaykurokawa:
seems to be failing test tests.unit.lbryfilemanager.test_EncryptedFileCreator.CreateEncryptedFileTest, 

there is some weired dependency on blob manager that's causing it to fail I guess.

kaykurokawa:
Fixed the failing unit test tests.unit.lbryfilemanager.test_EncryptedFileCreator.CreateEncryptedFile

jobevers:
LGTM

kaykurokawa:
rebase to master
added a fix to functional test (mock Announcer class now requires function hash_queue_size()

PR #466, opened 2017-02-08
fix streams in stream_info_manager not being loaded by lbry_file_manager

jackrobison:
This migrates any streams tracked by stream_info_manager but not lbry_file_manager to also be included by the file manager

This commit should prevent new streams from having this problem: https://github.com/lbryio/lbry/commit/2126f69c930a4a4d22169cff3533362fe6e0bcaf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ at anytime.
 ### Fixed
  * fix recursion depth error upon failed blob
  * call stopProducing in reflector client file_sender when uploading is done
+ * ensure streams in stream_info_manager are saved in lbry_file_manager
 
 ## [0.8.1] - 2017-02-01
 ### Changed
--- a/lbrynet/core/Error.py
+++ b/lbrynet/core/Error.py
@@ -98,9 +98,3 @@ class InvalidAuthenticationToken(Exception):
 
 class NegotiationError(Exception):
     pass
-
-
-class MissingLBRYFile(Exception):
-    """
-    Raised by lbry file manager if stream hash has no associated lbry file
-    """
--- a/lbrynet/core/Error.py
+++ b/lbrynet/core/Error.py
@@ -98,3 +98,9 @@ class InvalidAuthenticationToken(Exception):
 
 class NegotiationError(Exception):
     pass
+
+
+class MissingLBRYFile(Exception):
+    """
+    Raised by lbry file manager if stream hash has no associated lbry file
+    """
--- a/lbrynet/lbryfilemanager/EncryptedFileManager.py
+++ b/lbrynet/lbryfilemanager/EncryptedFileManager.py
@@ -10,6 +10,7 @@ from twisted.internet import defer, task, reactor
 from twisted.python.failure import Failure
 
 from lbrynet.core.PaymentRateManager import NegotiatedPaymentRateManager
+from lbrynet.core.Error import MissingLBRYFile
 from lbrynet.lbryfilemanager.EncryptedFileDownloader import ManagedEncryptedFileDownloader
 from lbrynet.lbryfilemanager.EncryptedFileDownloader import ManagedEncryptedFileDownloaderFactory
 from lbrynet.lbryfile.StreamDescriptor import EncryptedFileStreamType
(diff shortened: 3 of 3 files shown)


lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -74,12 +75,49 @@ def _add_to_sd_identifier(self):
...
             EncryptedFileStreamType, downloader_factory)
 
-    def _start_lbry_files(self):
+    def _check_stream_is_managed(self, stream_hash):
+        # check that all the streams in the stream_info_manager are also
+        # tracked by lbry_file_manager and fix any streams that aren't.
+        def _check_rowid_result(rowid):
+            if rowid is not None:
+                return defer.succeed(True)
+            raise MissingLBRYFile
```
jobevers: this should be `raise MissingLBRYFile()` - ideally though with some info about what is missing

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -74,12 +75,49 @@ def _add_to_sd_identifier(self):
...
+                d = self.stream_info_manager.get_stream_info(stream_hash)
+                d.addCallback(lambda info: log.warning("Trying to fix missing lbry file for %s",
+                                                       info[1].decode('hex')))
+                d.addCallback(lambda _: self._save_lbry_file(stream_hash, rate))
+                return d
+
+        d = self._get_rowid_for_stream_hash(stream_hash)
+        d.addCallback(_check_rowid_result)
+        d.addErrback(_fix_missing_file)
+        return d
```
jobevers: this would be nicer as inlineCallbacks

```
@defer.inlineCallbacks
def _check_stream_is_managed(self, stream_hash):
    row_id = yield self._get_rowid_for_stream_hash(stream_hash)
    if rowid:
        defer.returnValue(True)
    else:
        rate = self.session.base_payment_rate_manager.min_blob_data_payment_rate
        info = yield self.stream_info_manager.get_stream_info(stream_hash)
        # as an aside, I think its bad practice to refer to tuples like this
        # and I prefer to unpack the tuple into descriptive names so that the reader doesn't have to look up the 
        # get_stream_info function to see what is returned.  Alternatively, use namedtuples
        # 
        # Also, we shouldn't need to remember how to log each object in our codebase.  Save yourself
        # the effort and put in place good __repr__ and __str__ functions so that we can just log
        #
        log.warning("Trying to fix missing lbry file for %s", info[1].decode('hex')))
        yield self._save_lbry_file(stream_hash, rate)
```

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -254,10 +294,17 @@ def _get_lbry_file_status(self, rowid):
         d = self.sql_db.runQuery("select status from lbry_file_options where rowid = ?",
                                  (rowid,))
         d.addCallback(lambda r: (
-            r[0][0] if len(r) else ManagedEncryptedFileDownloader.STATUS_STOPPED))
+            r[0][0] if len(r) else None))
```
jobevers: nit: can all be on one line

lbrynet/core/Error.py
```diff
@@ -98,3 +98,9 @@ class InvalidAuthenticationToken(Exception):
 
 class NegotiationError(Exception):
     pass
+
+
+class MissingLBRYFile(Exception):
```
jobevers: see below, I don't think this is necessary

lbrynet/lbryfilemanager/EncryptedFileManager.py
jobevers: doing it this way gets rid of the exception raising as well.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ can and probably will change functionality and break backwards compatability
 at anytime.
 
 ## [Unreleased]
+### Fixed
+ * ensure streams in stream_info_manager are saved in lbry_file_manager
```
jobevers: 👍

jobevers:
LGTM

PR #467, opened 2017-02-08
Fix reflector bugs

jackrobison:
-fix recursion depth bug upon failed blob upload

-stopProducing file_sender when set_not_uploading is called

-include stream uri in reflector logs to de-obfuscate them

### DIFF
--- a/lbrynet/reflector/client/client.py
+++ b/lbrynet/reflector/client/client.py
@@ -250,7 +250,7 @@ class EncryptedFileReflectorClient(Protocol):
     def skip_missing_blob(self, err, blob_hash):
         log.warning("Can't reflect blob %s", str(blob_hash)[:16])
         err.trap(ValueError)
-        return self.send_next_request()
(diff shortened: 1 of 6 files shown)


lbrynet/reflector/client/client.py
```diff
@@ -250,7 +250,7 @@ def send_descriptor_info(self):
     def skip_missing_blob(self, err, blob_hash):
         log.warning("Can't reflect blob %s", str(blob_hash)[:16])
```
jobevers: can you log the err traceback so that we know why this happens? I do worry that maybe we'll continue retrying something ad-infinitum without reason.

jobevers:
one comment, otherwise LGTM

lbrynet/reflector/client/client.py
jackrobison: I changed it to only retry the upload once if it fails, and to include the traceback in the log

PR #469, opened 2017-02-08
Blob commands

jackrobison:


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1982,28 +2057,113 @@ def jsonrpc_block_show(self, blockhash=None, height=None):
...
+
+        timeout = timeout or 30
+
+        if payment_rate_manager:
+            rate_managers = {
+                'only-free': OnlyFreePaymentsManager()
+            }
+            if payment_rate_manager in rate_managers:
+                payment_rate_manager = rate_managers[payment_rate_manager]
+                log.info("Downloading blob with rate manager: %s", payment_rate_manager)
```
jobevers: did you copy and paste this code?

jobevers:
one section of copy/pasted code, otherwise LGTM

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: moved into a function and rebased

Issue #470, opened 2017-02-09
file_delete does not remove data from stream_info_manager


jackrobison:
because EncryptedFileManager.get_count_for_stream doesn't return an int, it returns [(int,)]

https://github.com/lbryio/lbry/blob/d053db8dfd345f12989de50de0731084652d0062/lbrynet/lbrynet_daemon/Daemon.py#L824-L824

The stream info being in the EncryptedFileMetadataManager (after deleting the blobs) causes the reflector client to try sending blobs that it doesn't have (a cause of failed uploads).

PR #471, opened 2017-02-09
fix file_delete

jackrobison:
Fixes https://github.com/lbryio/lbry/issues/470

### DIFF
--- a/lbrynet/lbryfilemanager/EncryptedFileManager.py
+++ b/lbrynet/lbryfilemanager/EncryptedFileManager.py
@@ -281,8 +281,10 @@ class EncryptedFileManager(object):
 
     @rerun_if_locked
     def _get_count_for_stream_hash(self, stream_hash):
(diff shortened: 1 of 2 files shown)


lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -281,8 +281,10 @@ def _get_lbry_file_status(self, rowid):
 
     @rerun_if_locked
     def _get_count_for_stream_hash(self, stream_hash):
-        return self.sql_db.runQuery("select count(*) from lbry_file_options where stream_hash = ?",
+        d = self.sql_db.runQuery("select count(*) from lbry_file_options where stream_hash = ?",
                                      (stream_hash,))
+        d.addCallback(lambda r: (r[0][0] if len(r) else 0))
```
jobevers: can be `lambda r: r[0][0] if r else 0`

jobevers:
LGTM

PR #472, opened 2017-02-09
Fix how the bug report code accesses Slack API key setting

alexliebowitz:
It was trying to access this setting as an attribute, which doesn't work anymore. (Not sure if this is an API change in conf.settings or just a bug.)


jackrobison:
LGTM

PR #475, opened 2017-02-09
Adding reactor.callLater as a utils function

kaykurokawa:
Add reactor.callLater https://twistedmatrix.com/documents/16.4.0/api/twisted.internet.interfaces.IReactorTime.callLater.html as a utils function call_later() , so that trial tests can replace it with clock.callLater or some other non reactor function. Will be used in https://github.com/lbryio/lbry/pull/465 and https://github.com/lbryio/lbry/pull/458 , also need to replace instances of reactor.callLater in other parts of the code.

### DIFF
--- a/lbrynet/core/utils.py
+++ b/lbrynet/core/utils.py
@@ -45,6 +45,13 @@ def datetime_obj(*args, **kwargs):
     return datetime.datetime(*args, **kwargs)
 
 
+def call_later(delay, func, *args, **kwargs):
+    # Import here to ensure that it gets called after installing a reator
+    # see: http://twistedmatrix.com/documents/current/core/howto/choosing-reactor.html
+    from twisted.internet import reactor
+    return reactor.callLater(delay, func, *args, **kwargs)
+
+
 def generate_id(num=None):
     h = get_lbry_hash_obj()
     if num is not None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ at anytime.
 \#\# [0.8.6rc0] - 2017-02-19
 ### Changed
  * Add `file_get` by stream hash
+ * Add utils.call_later to replace reactor.callLater
 ###
  * Fix unhandled error in `get`
  * Fix sd blob timeout handling in `get_availability`, return 0.0
--- a/lbrynet/core/HashAnnouncer.py
+++ b/lbrynet/core/HashAnnouncer.py
@@ -11,5 +11,8 @@ class DummyHashAnnouncer(object):
     def add_supplier(self, *args):
         pass
 
+    def hash_queue_size(self):
+        return 0
+
     def immediate_announce(self, *args):
         pass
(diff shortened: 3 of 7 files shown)


lbrynet/core/utils.py
```diff
@@ -9,6 +9,8 @@
 
 import pkg_resources
 
+from twisted.internet import reactor
```
jackrobison: I think this should probably be in the `call_later` function, so that importing lbrynet.core.utils can't inadvertently configure an unconfigured reactor if it's imported early

kaykurokawa: that doesn't sound right. imports in python should not re import if the module already exists.

jackrobison: if twisted.internet.reactor was imported before:

https://github.com/lbryio/lbry/blob/267c6cbaca739fdce06c51b3251eecb75b9f31f5/packaging/osx/lbry-osx-app/lbrygui/main.py#L3-L3

I suppose it can't be a problem if we aren't using pyobjc for the app

jobevers: I have been curious about this - I've definitely seen example code where `reactor` was imported inside a function and I've never dug into why.  Guess this would be a good time to learn.

kaykurokawa: reactortest1.py
http://pastebin.com/6HPWQTBb
reactortest2.py
http://pastebin.com/gcvKnFUe

Doesn't look like a problem.  reactortest2.py is fully aware of the added system trigger added in reactortest1.py

jackrobison: http://twistedmatrix.com/documents/current/core/howto/choosing-reactor.html

kaykurokawa: choose a different reactor in reactortest1.py and check type(reactor) in reactortest2.py , you will see it will be the same reactor as initialized in reactortest1.py 

duplicate imports in python should do nothing but return the reference to the original import so it does not make sense that it would reconfigure reactor

jobevers: The worry is what happen if you don't want the default reactor:

"IMPORTANT: installing a reactor should be the first thing done in the app, since any code that does from twisted.internet import reactor will automatically install the default reactor if the code hasn’t already installed one." (from http://twistedmatrix.com/documents/current/core/howto/choosing-reactor.html)

And, as Jack referenced, this is largely a problem if the lbrynet library is being using in a gui app that needs to configure the reactor to use the GUI's event loop.  It might also be an issue for testing, but I think `trial` should take care of it.

Regardless, I think importing reactor at module load-time makes it harder to use the library.

jobevers:
This PR needs to move the reactor import and add a changelog entry.  After that, LGTM.

lbrynet/core/utils.py
kaykurokawa: ok lets import in the function then ,

kaykurokawa:
I will also add a commit that uses this function where it needs to be used

kaykurokawa:
Added changelog and rebased

PR #476, opened 2017-02-09
Simplify lbrynet.lbrynet_daemon.Publisher

jackrobison:


lbrynet/core/file_utils.py
```diff
@@ -33,3 +33,15 @@ def reveal(path):
...
+def get_read_handle(path):
+    """
+    Get os independent read handle for a file
+    """
+
+    if os.name == "nt":
+        file_mode = 'rb'
+    else:
+        file_mode = 'r'
+    return open(path, file_mode)
```
jobevers: Can you make this a context object so that we can do:

```
with file_utils.get_read_handle(path) as f:
    <stuff>
```

lbrynet/lbrynet_daemon/Publisher.py
```diff
@@ -21,154 +17,30 @@ def __init__(self, session, lbry_file_manager, wallet):
...
-        return threads.deferToThread(check_file_threaded)
-
-    def set_lbry_file(self, lbry_file_downloader):
-        self.lbry_file = lbry_file_downloader
-        return defer.succeed(None)
-
-    def add_to_lbry_files(self, stream_hash):
-        self.stream_hash = stream_hash
+        file_name = os.path.basename(file_path)
+        with file_utils.get_read_handle(file_path) as read_handle:
```
jobevers: I don't think this works.  See https://docs.python.org/2.7/reference/datamodel.html#context-managers for more info, and https://docs.python.org/2.7/library/contextlib.html for some helpers.

lbrynet/core/file_utils.py
```diff
@@ -33,3 +35,18 @@ def reveal(path):
...
+    Get os independent read handle for a file
+    """
+
+    if os.name == "nt":
+        file_mode = 'rb'
+    else:
+        file_mode = 'r'
+    read_handle = open(path, file_mode)
+    yield read_handle
+    read_handle.close()
```
jobevers: nice

lbrynet/lbrynet_daemon/Publisher.py
jobevers: you fixes this.  Thanks!

jackrobison:
included in https://github.com/lbryio/lbry/pull/477

PR #477, opened 2017-02-10
Refactor publish, file manager, and file downloader

jackrobison:


jackrobison:
@jobevers I rebased in your publish_sd_blob commit and moved reflecting the lbry file to a looping call in the file manager

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -108,13 +120,14 @@ def _start_lbry_files(self):
...
+            lbry_file = yield self.start_lbry_file(rowid, stream_hash, payment_rate_manager,
+                                                   blob_data_rate=options)
             try:
-                lbry_file = yield self.start_lbry_file(rowid, stream_hash, payment_rate_manager,
-                                 blob_data_rate=options)
                 yield lbry_file.restore()
             except Exception as err:
-                msg = "Failed to start lbry file, rowid: %i, stream_hash: %s, options: %s, err: %s"
-                log.error(msg, rowid, stream_hash, options, err)
+                log.error("Failed to start stream: %s, error: %s", stream_hash, err)
```
jobevers: this is better as `log.exception("Failed to start stream: %s", stream_hash)`

https://docs.python.org/2/library/logging.html#logging.Logger.exception

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -108,13 +120,14 @@ def _start_lbry_files(self):
         yield self._check_stream_info_manager()
         lbry_files_and_options = yield self._get_all_lbry_files()
         for rowid, stream_hash, options in lbry_files_and_options:
+            lbry_file = yield self.start_lbry_file(rowid, stream_hash, payment_rate_manager,
+                                                   blob_data_rate=options)
```
jobevers: why move this out of the try / except?

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -202,8 +215,17 @@ def toggle_lbry_file_running(self, lbry_file):
         else:
             return defer.fail(Failure(ValueError("Could not find that LBRY file")))
 
+    def _reflect_lbry_files(self):
+        for lbry_file in self.lbry_files:
+            yield reflect_stream(lbry_file)
+
+    @defer.inlineCallbacks
+    def reflect_lbry_files(self):
+        yield defer.DeferredList(list(self._reflect_lbry_files()))
```
jobevers: I think this is better using a list comprehension:
```
yield defer.DeferredList([reflect_stream(lf) for lf in self.lbry_files])
```

jobevers:
I only glanced through the inlinecallback refactorings - definitely makes things more readable.

A couple of small comments, otherwise LGTM

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -97,64 +113,75 @@ def _iter_streams(stream_hashes):
...
+        b_prm = self.session.base_payment_rate_manager
+        payment_rate_manager = NegotiatedPaymentRateManager(b_prm, self.session.blob_tracker)
         yield self._check_stream_info_manager()
-        files_and_options = yield self._get_all_lbry_files()
-        yield start_lbry_files(files_and_options)
+        lbry_files_and_options = yield self._get_all_lbry_files()
+        for rowid, stream_hash, options in lbry_files_and_options:
+            lbry_file = yield self.start_lbry_file(rowid, stream_hash, payment_rate_manager,
+                                                   blob_data_rate=options)
+            d = self._restore_lbry_file(lbry_file)
```
jobevers: mentioned in our phone chat - but this would be nice to have inside a deferredList so that we can keep track of the deferred instead of having it disappear into twisted.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -774,6 +774,26 @@ def _download_name(self, name, timeout=None, download_directory=None,
...
+        verify_name_characters(name)
+        if bid <= 0.0:
+            raise Exception("Invalid bid")
+        if fee:
+            metadata = yield publisher.add_fee_to_metadata(metadata, fee)
+        if not file_path:
+            claim_out = yield publisher.update_stream(name, bid, metadata)
+        else:
+            claim_out = yield publisher.publish_stream(name, file_path, bid, metadata)
+            yield threads.deferToThread(reupload.reflect_stream, publisher.lbry_file)
```
jobevers: I'm not sure this needs to be a separate thread - I think as long you don't yield, you can create a deferred and have twisted process it.  This can probably be:
```
# make this its own deferred so as to not block publishing more than it already is
d = reupload.reflect_stream(publisher.lbry_file)
d.addCallback(lambda _: log.info('Successfully reflected %s after publish', publisher.lbry_file))
d.addErrback(log.fail(), 'Failed to reflect %s after publish', publisher.lbry_file))
```

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1683,15 +1684,14 @@ def jsonrpc_claim_abandon(self, txid, nout):
...
-        d = defer.Deferred()
-        d.addCallback(lambda _: self.session.wallet.abandon_claim(txid, nout))
-        d.addCallback(_disp)
-        d.callback(None)  # TODO: is this line necessary???
-        return d
+        try:
+            abandon_claim_tx = yield self.session.wallet.abandon_claim(txid, nout)
+            response = yield self._render_response(abandon_claim_tx)
+        except Exception as err:
+            log.warning(err)
```
jobevers: use `exc_info` in log statements that you want to include the error message.

This should be more like, `log.warning('Failed to abandon claim %s - %s', txid, nout, exc_info=True)`

jobevers:
Still LGTM.  Go ahead and merge.

Issue #479, opened 2017-02-11
UnknownNameError


jobevers:
Lots of `UnknownNameErrors` show up in the logs.

PR #482, opened 2017-02-11
Catch unknown name error

jobevers:
See #479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ at anytime.
 ## [Unreleased]
 ### Fixed
  * Add better logging to help track down [#478](https://github.com/lbryio/lbry/issues/478)
-
+ * Catch UnknownNameErrors when resolving a name. [#479](https://github.com/lbryio/lbry/issues/479)
 
 ## [0.8.3rc3] - 2017-02-14
 
@@ -24,6 +24,7 @@ at anytime.
  * Change EWOULDBLOCK error in DHT to warning. #481
  * mark peers as down if it fails download protocol
  * Made hash reannounce time to be adjustable to fix [#432](https://github.com/lbryio/lbry/issues/432)
+
 
 ## [0.8.3rc0] - 2017-02-10
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ can and probably will change functionality and break backwards compatability
 at anytime.
 
 ## [Unreleased]
+### Fixed
+ * Catch UnknownNameErrors that don't need to be thrown. #479
```
jobevers: damn.  I was hoping github's markdown would recognize and link to this automatically.  It doesn't 😢

kaykurokawa:
I can review this

kaykurokawa:
There are some other places in the Daemon that could use the same catching 

a) jsonrpc_get_availability() 

b) get_est_cost_using_known_sizes() could also use the same catching. Alternatively since get_est_cost() calls both get_est_cost_using_known_sizes() and get_est_cost_from_name() , you could catch it in get_est_cost()

jobevers:
Thanks for pointing out my inconsistency.

For `jsonrpc_resolve_name` - that is used to see if a name is available so it is expected that names won't resolve.

I'm not sure what I was thinking with `get_est_cost_from_name` - looking this over, that seems like it should be an error (because, why would you request the cost for a name that doesn't exists?) and the root cause is coming from somewhere else.

I'm going to remove the check on get_est_cost_from_name.

jobevers:
@alexliebowitz this doesn't cause any backwards incompatability as you'll still get null if a name doesn't exists - but does mean that you can actually treat errors are errors: https://github.com/lbryio/lbry-web-ui/blob/7e77f529761b397999de04b0293e1cf23cb84f79/js/lbry.js#L165

kaykurokawa:
Before this patch though, resolve_name would throw an exception and you wouldn't get null though. Seems like the web-ui might need to change how it handles that? 

otherwise LGTM

jobevers:
The web-ui takes all errors and returns null to the callback.  With this patch, non-existing names will return null straight from the api, which will also get passed to the callback.  So the web-ui doesn't need to change.  It should change though so that errors are actually treated as errors.

Issue #480, opened 2017-02-11
"Failed to render API Response" has no traceback


jobevers:
Seeing lots of messages logged from here, but without a traceback. https://github.com/lbryio/lbry/blob/master/lbrynet/lbrynet_daemon/auth/server.py#L331

Makes it hard to debug what might have gone wrong.

lyoshenka:
pretty sure this is fixed

PR #484, opened 2017-02-11
record traceback in error event

jobevers:
In investigating #480, turns out log_record.message doesn't have traceback info - so adding another field to record it


kaykurokawa:
this is just for loggly right?

jobevers:
this goes into metabase - we have an error table that tracks each logged
error.  Loggly already gets the full log message.

kaykurokawa:
lgtm

jobevers:
this passed on linux, merging

PR #485, opened 2017-02-14
file_get by name for file with pending claim

jackrobison:
This fixes publish not redirecting to the published files page after publish() returns

-if a lbry file from the winning claim is in the file manager, return it. otherwise return a lbry file from a pending claim if one exists.

-update lbry_file.start to match lbry_file.restore

-add lbry_file.load_attributes

### DIFF
--- a/lbrynet/core/Error.py
+++ b/lbrynet/core/Error.py
@@ -84,6 +84,12 @@ class NoSuchStreamHashError(Exception):
     pass
 
 
+class NoSuchSDHash(Exception):
+    """
+    Raised if sd hash is not known
+    """
+
+
 class InvalidBlobHashError(Exception):
     pass
 
--- a/lbrynet/lbryfilemanager/EncryptedFileDownloader.py
+++ b/lbrynet/lbryfilemanager/EncryptedFileDownloader.py
@@ -7,6 +7,8 @@ from zope.interface import implements
 from twisted.internet import defer
 
 from lbrynet.core.client.StreamProgressManager import FullStreamProgressManager
+from lbrynet.core.Error import NoSuchSDHash, NoSuchStreamHashError
+from lbrynet.core.utils import short_hash
 from lbrynet.core.StreamDescriptor import StreamMetadata
 from lbrynet.lbryfile.client.EncryptedFileDownloader import EncryptedFileSaver
 from lbrynet.lbryfile.client.EncryptedFileDownloader import EncryptedFileDownloader
@@ -15,6 +17,18 @@ from lbrynet.interfaces import IStreamDownloaderFactory
 from lbrynet.lbryfile.StreamDescriptor import save_sd_info
 
 log = logging.getLogger(__name__)
+
+
+def log_status(uri, sd_hash, status):
+    if status == ManagedEncryptedFileDownloader.STATUS_RUNNING:
+        status_string = "running"
+    elif status == ManagedEncryptedFileDownloader.STATUS_STOPPED:
+        status_string = "stopped"
+    elif status == ManagedEncryptedFileDownloader.STATUS_FINISHED:
+        status_string = "finished"
+    else:
+        status_string = "unknown"
+    log.info("lbry://%s (%s) is %s", uri, short_hash(sd_hash), status_string)
 
 
 class ManagedEncryptedFileDownloader(EncryptedFileSaver):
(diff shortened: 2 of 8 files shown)


lbrynet/lbryfilemanager/EncryptedFileDownloader.py
```diff
@@ -59,10 +60,13 @@ def restore(self):
         self.claim_id = yield self.wallet.get_claimid(self.uri, self.txid, self.nout)
         status = yield self.lbry_file_manager.get_lbry_file_status(self)
         if status == ManagedEncryptedFileDownloader.STATUS_RUNNING:
+            log.info("lbry://%s (%s) is running", self.uri, self.sd_hash[:6])
```
jobevers: Please don't repeat the `self.sd_hash[:6]` part.  If you want that to be the standard way of logging an sd_hash, than make a function or anything that allows you to change how its logged if you so choose in the future.  It also ensures that you're consistent with how these sorts of things are logged.

lbrynet/core/Error.py
```diff
@@ -84,6 +84,12 @@ class NoSuchStreamHashError(Exception):
     pass
 
 
+class NoSuchSDHashError(Exception):
```
jobevers: a pet peeve of mine: its not necessary to have `Error` in your exception.  Its redundant because its always obvious in context that you're dealing with an error.. Ie - `raise NoSuchSDHash(...)` is enough.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -956,19 +958,22 @@ def get_est_cost(self, name, size=None):
...
+            stream_info = yield self._resolve_name(name)
+            sd_hash = stream_info['sources']['lbry_sd_hash']
+        except UnknownNameError:
+            sd_hash = None
+
+        if not sd_hash:
+            for lbry_file in self.lbry_file_manager.lbry_files:
+                if name == lbry_file.uri:
+                    defer.returnValue(lbry_file)
+        else:
```
jobevers: This else statement is a double-negative. My brain deals with positives better than double-negatives.

This should read

```
if sd_hash:
    lbry_file  = self.find_lbry_file_by_sd_hash(sd_hash)
else:
   lbry_file = self.find_lbry_file_by_uri(name)
defer.returnValue(lbry_file)
```

its also worth documenting that lbry_file can be None if there is no match found

lbrynet/lbrynet_daemon/Publisher.py
```diff
@@ -48,6 +48,7 @@ def publish_stream(self, name, file_path, bid, metadata):
         metadata['content_type'] = get_content_type(file_path)
         metadata['ver'] = Metadata.current_version
         claim_out = yield self.make_claim(name, bid, metadata)
+        yield self.lbry_file.start()
```
jobevers: What does this do?

jobevers: I think this calls 
https://github.com/lbryio/lbry/blob/master/lbrynet/cryptstream/client/CryptStreamDownloader.py#L97, which gives you back a finished deferred - and you're going to wait for that to be called. 

This seems odd to want to 'start' downloading a stream that you already have.  If you want to populate variables, it would be better to have a function that does just that and have `start()` also call that function.

kaykurokawa:
Having get_lbry_file_by_uri returning pending claims seems like a hack for fixing the real problem. What if you are publishing over an already claimed name?  IMO get_lbry_file_by_uri() should always return the winning claim. 

I guess the Web-UI rely on get_lbyr_file_by_uri to check for pending names, is that right? 

Why not just make a get_my_pending_lbry_files() function in the daemon?

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: changed it to raise an exception instead of returning None

lbrynet/lbrynet_daemon/Publisher.py
jackrobison: added a load_file_attributes function and removed the start() call

jackrobison:
@kaykurokawa if you have a lbry file from a winning claim it'll return that file, if you don't have the current winning claim but you do have a file from a pending one, then it'll return that.

jobevers:
LGTM

kaykurokawa:
Let's say you have two files (A and B) associated with two different claims for the same name "test". File A is currently associated with the winning claim. You call get_lbry_file_by_uri() and you get file A. Someone overtakes name "test" with some random file that you have not downloaded. You call get_lbry_file_by_uri() and you will either get file A or B depending on which one shows up first in the file manager. This is a very arbitrary behavior that's going to confuse the user. 

We need a separate function, something like get_lbry_file_by_name() that does not attempt to resolve the name through the blockchain, an just looks up the name in the file manager. get_lbry_file_by_uri() should be for the winning claim, and if you don't have the winning claim it should return nothing.

We should also return a list of files that is associated with a description whether it be sd_hash, name, or filename, instead of returning whatever is the first one in the file manager.

PR #486, opened 2017-02-14
Electron

jobevers:
Changes needed for electron, along with removing parts that aren't needed now that packaging happens in electron.


jobevers:
This won't build because I took out all of the packaging stuff.  I'd like to move this repo towards being a library... but not sure yet what to do with it yet as I've focused on the electron build.

jobevers:
Changed the build to only run on linux, it installs the library and runs the tests, no more packaging, etc.

lbrynet/core/utils.py
```diff
@@ -97,11 +95,5 @@ def check_connection(server="www.lbry.io", port=80):
         return False
 
 
-def setup_certs_for_windows():
```
jobevers: pyinstaller handles bundling the cacert just fine and so we don't need to set the environment variable.

lbrynet/lbrynet_daemon/auth/server.py
```diff
@@ -145,7 +145,7 @@ def _handle_dropped_request(self, result, d, function_name):
 
     def render(self, request):
         time_in = utils.now()
-        assert self._check_headers(request), InvalidHeaderError
+        # assert self._check_headers(request), InvalidHeaderError
```
jobevers: jack - you probably don't like that I did this, but electron is now essentially a separate process the queries the daemon via rpc and the checks in check_headers didn't seem that useful.  Open to being convinced otherwise.

packaging/travis/setup_build.sh
```diff
@@ -1,35 +0,0 @@
...
-set -o xtrace
-
-# changes to this script also need to be added to build.ps1 for windows
-add_ui() {
-    wget https://s3.amazonaws.com/lbry-ui/development/dist.zip -O dist.zip
-    unzip -oq dist.zip -d lbrynet/resources/ui
-    wget https://s3.amazonaws.com/lbry-ui/development/data.json -O lbrynet/resources/ui/data.json
-}
-
-set_build() {
```
jobevers: crap.  electron is currently building with the release set to dev.

lbrynet/lbrynet_daemon/auth/server.py
jackrobison: What's the new origin with electron, and why not update accordingly? If anything I think we should use the file system access in electron to get an authenticated api session for everything the ui does, removing the header check would be a regression on that front.

packaging/travis/setup_build.sh
jobevers: I fixed this.

lbrynet/lbrynet_daemon/auth/server.py
jobevers: While I agree that using authentication would be a good step forward, I don't think we should delay the electron launch due to that.

Also Electron is sending requests with essentially an empty origin, and no referrer.

image

I am not convinced checking headers buys us much in the way of security.  I'm open to being convinced otherwise.

jackrobison: >I don't think we should delay the electron launch due to that.

Agreed.

>I am not convinced checking headers buys us much in the way of security

It doesn't, they can be forged, but it does rule out requests that aren't forged. I suppose I think we should change the origin to `file://`, make note that double clicking a local html file can mess up your day, and start using auth as soon as possible to correct it. In addition to using auth, we should set the api keys to only be readable by the signed app and sudo as to prevent others with access to the file system from getting an authenticated session.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2064,28 +2054,6 @@ def jsonrpc_upload_log(self, name_prefix=None, log_type=None, exclude_previous=T
         return d
 
     @AuthJSONRPCServer.auth_required
-    def jsonrpc_configure_ui(self, branch=None, path=None, check_requirements=True):
```
jackrobison: Can you add this to the changelog?

requirements.txt
```diff
@@ -10,7 +9,6 @@ gmpy==1.17
 jsonrpc==1.2
 jsonrpclib==0.1.7
 jsonschema==2.5.1
-https://github.com/lbryio/lbryum/tarball/master/#egg=lbryum
```
jackrobison: this breaks `pip install -r requirements.txt` in development setup

setup.py
```diff
@@ -31,39 +19,38 @@
...
+    'six',
+    'slowaes',
+    'txJSON-RPC',
+    'wsgiref',
+    'zope.interface',
+    'base58',
+    'googlefinance',
+    'pyyaml',
+    'service_identity',
+    'ndg-httpsclient',
```
jackrobison: Can you read this list out of requirements.txt?

jackrobison:
Where is signing the app happening? (or is it not yet?)

jobevers:
No packaging happens in the lbry repo anymore.  The signing happens as part of the electron build.  Teamcity has a copy of the signing key.

packaging/osx/lbry-osx-app/setup_uri_handler.py
```diff
@@ -1,26 +0,0 @@
...
-               'LSUIElement': True,
-               'CFBundleIdentifier': 'io.lbry.LBRYURIHandler',
-               'CFBundleURLTypes': [
-                    {
-                    'CFBundleURLTypes': 'LBRYURIHandler',
-                    'CFBundleURLSchemes': [conf.PROTOCOL_PREFIX]
-                    }
-               ]
-           }
-        }
```
jackrobison: Does electron have an appropriate plist file to support uri handling on os x?

packaging/windows/lbry-win32-app/LBRY-URI.reg
```diff
@@ -1,15 +0,0 @@
...
-
-[HKEY_CLASSES_ROOT\lbry\DefaultIcon]
-@="\"LBRY.exe,1\""
-
-[HKEY_CLASSES_ROOT\lbry\shell]
-
-[HKEY_CLASSES_ROOT\lbry\shell\open]
-
-[HKEY_CLASSES_ROOT\lbry\shell\open\command]
-@="\"C:\\Program Files (x86)\\LBRY\\LBRY.exe\" \"%1\""
```
jackrobison: Is uri handling broken on windows?

setup.py
jobevers: Even this is is not considered best practice.  setup.py should only list the dependencies that lbrynet needs (not a list of dependencies of dependencies) along with any known version restrictions.  Its generally a bad idea to pin versions in setup.py

On the other hand - requirements.txt is intended to be an explicit set of libraries for deployment.

https://caremad.io/posts/2013/07/setup-vs-requirement/
https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/cxjs7n4/?utm_content=permalink&utm_medium=front&utm_source=reddit&utm_name=Python

jackrobison: I just wanna be able to add or remove requirements in one place and not worry about having them sync'ed

jobevers:
The push test is failing, but really don't care about that one (I'd disable it, but then it won't build master.  Travis is such a mess.)

PR #488, opened 2017-02-16
Remote version check

jobevers:
Change to checking the lbry-electron repo for updates.

I have a remaining todo to make sure that lbrynet.__version__ is synced up with the version number used in releasing  lbry-electron.  That will be in the lbry-electron repo


jobevers:
Oh, gross, I rebased the electron branch and now this happened.  I'm going to cherry-pick these commits onto the electron branch

Issue #498, opened 2017-02-22
Running trial tests broken on Ubuntu 16.04 after electron update 


kaykurokawa:
run_tests.sh script no longer works on Ubuntu 16.04 after the electron update which uses Twisted 16.6. 
Trial thinks that "tests" is some other random module instead of the one in our lbrynet folder. 

Renaming our tests folder to some other name such as lbrynet_tests fixes this problem. Probably there is some better way to force trial to use our tests folder.

jobevers:
You might have a different `tests` module installed in your python path.
In a python terminal, do:

```
> import tests
> print tests.__file__
```

lyoshenka:
​At one point, Job told me to run tests with `PYTHONPATH=. trial tests`.
Does that changed anything?​

On Wed, Feb 22, 2017 at 4:52 PM, Job Evers‐Meltzer <notifications@github.com

kaykurokawa:
It was our lighthouse tests , I think it installs the tests folder as module "tests" 
https://github.com/lbryio/lighthouse/blob/master/setup.py

jobevers:
Ah. This line https://github.com/lbryio/lighthouse/blob/master/setup.py#L19
should probably be `packages=find_packages('lighthouse')`

kaykurokawa:
This worked for me:

packages=find_packages(base_dir,exclude=['tests'])

We should probably do the same for setup.py in lbrynet

kaykurokawa:
Closing as proper pull requests has been made

PR #503, opened 2017-02-28
Exclude tests folders in setup.py

kaykurokawa:
Without this, it would install our tests folder as a module tests in python 

See https://github.com/lbryio/lbry/issues/498


lyoshenka:
@kaykurokawa in the future, you can just merge this without a PR :-)

PR #505, opened 2017-03-01
Database refactor

jackrobison:


kaykurokawa:
I think you don't want these classes to call storage.close()

lbrynet/core/BlobManager.py:        
lbrynet/lbryfilemanager/EncryptedFileManager.py        
lbrynet/lbryfile/EncryptedFileMetadataManager.py

Because they don't know if the other classes are done with using the Storage class and will result in a race condition.

lbrynet/core/Wallet.py
```diff
@@ -654,9 +618,15 @@ def claim_name(self, name, bid, metadata):
 
         claim = self._process_claim_out(claim)
         claim_outpoint = ClaimOutpoint(claim['txid'], claim['nout'])
-        log.info("Saving metadata for claim %s %d", claim['txid'], claim['nout'])
-
-        yield self._save_name_metadata(name, claim_outpoint, _metadata['sources']['lbry_sd_hash'])
```
kaykurokawa: This is wrong. Firstly because generate_claimid is broken because it hashes a hex encoded txid and not txid bytes. 
You also don't need to generate claim id because when it is a new claim, the lbryum claim command returns the claim_id in the output, and if it is not a new claim, you already know the old claim_id in my_claim. 

        if is_new_claim:
            claim_id = claim['claim_id']
        else:
            claim_id = my_claim['claim_id']

kaykurokawa:
This pull request fails https://github.com/lbryio/lbry-in-a-box/blob/master/integration_testing.py#L365 , 

It looks like reflector gets the blob but does not announce it immediately. Would be a problem in BlobManager.

lbrynet/core/BlobManager.py
```diff
@@ -1,473 +1,180 @@
...
-        blob = self.blob_type(self.blob_dir, blob_hash, length)
-        self.blobs[blob_hash] = blob
-        return defer.succeed(blob)
-
+    @defer.inlineCallbacks
     def blob_completed(self, blob, next_announce_time=None):
         if next_announce_time is None:
             next_announce_time = self.get_next_announce_time()
-        d = self._add_completed_blob(blob.blob_hash, blob.length, next_announce_time)
-        d.addCallback(lambda _: self._immediate_announce([blob.blob_hash]))
```
kaykurokawa: why did you delete immediate_announce here?

lbrynet/core/BlobManager.py
```diff
@@ -1,473 +1,180 @@
...
-        return self._get_blobs_to_announce()
+        hashes_to_announce = yield self.storage.get_blobs_to_announce()
+        next_announce_time = self.get_next_announce_time(len(hashes_to_announce))
+        yield self.storage.update_next_blob_announce(hashes_to_announce, next_announce_time)
+        defer.returnValue(hashes_to_announce)
 
+    @defer.inlineCallbacks
     def creator_finished(self, blob_creator):
-        log.debug("blob_creator.blob_hash: %s", blob_creator.blob_hash)
+        log.debug("creator %s finished (%s)", blob_creator.blob_hash, str(type(blob_creator)))
```
kaykurokawa: I don't see this get called on reflector

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1605,25 +1610,91 @@ def jsonrpc_stream_cost_estimate(self, name, size=None):
...
+            metadata['license_url'] = license_url
+        if thumbnail is not None:
+            metadata['thumbnail'] = thumbnail
+        if preview is not None:
+            metadata['preview'] = preview
+        if nsfw is not None:
+            metadata['nsfw'] = nsfw
+        if sources is not None:
+            metadata['sources'] = sources
+        if fee is not None:
```
kaykurokawa: if fee is specified in metadata instead of argument,the address will never be added
I fixed this in:
https://github.com/lbryio/lbry/pull/531

jackrobison:
I'll re-open this when all the conflicts are sorted out

PR #510, opened 2017-03-06
Removing unuseful/unused commands from daemon API

kaykurokawa:
Remove unuseful/unused commands from Daemon API

get_nametrie
log
upload_log
get_mean_availability


kaykurokawa:
also delete open and reveal commands, and their respective functions in file_utils.py

PR #512, opened 2017-03-06
File list refactor

jackrobison:


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1426,90 +1436,55 @@ def _display_shutdown_message():
...
-        """
-        Get a file, if no matching file exists returns False
+            'name' (optional): filter files by lbry name,
+            'sd_hash' (optional): filter files by sd hash,
+            'file_name' (optional): filter files by the name in the downloads folder,
+            'stream_hash' (optional): filter files by stream hash,
+            'claim_id' (optional): filter files by claim id,
+            'outpoint' (optional): filter files by claim outpoint,
+            'rowid' (optional): filter files by internal row id,
+            'full_status': (optional): bool, if true populate the 'message' and 'size' fields
```
kaykurokawa: not sure why this full_status option is necessary. Why not just always return full status

kaykurokawa: not sure why this full_status option is necessary. Why not just always return full status

lbrynet/lbryfilemanager/EncryptedFileManager.py
```diff
@@ -190,33 +190,40 @@ def add_lbry_file(self, stream_hash, payment_rate_manager, blob_data_rate=None,
...
+        self.lbry_files.remove(lbry_file)
 
-        d.addCallback(lambda _: remove_from_list())
-        d.addCallback(lambda _: self._delete_lbry_file_options(lbry_file.rowid))
-        return d
+        yield self._delete_lbry_file_options(lbry_file.rowid)
+
+        yield lbry_file.delete_data()
+        stream_count = yield self.get_count_for_stream_hash(lbry_file.stream_hash)
+        if stream_count == 0:
```
kaykurokawa: why do we delete here is steam_count is zero?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1680,41 +1661,40 @@ def jsonrpc_file_seed(self, status, **kwargs):
...
+            'stream_hash' (optional): delete files by stream hash,
+            'claim_id' (optional): delete files by claim id,
+            'outpoint' (optional): delete files by claim outpoint,
+            'rowid': (optional): delete file by rowid in the file manager
+            'delete_target_file' (optional): delete file from downloads folder, defaults to True
+                                             if False only the blobs and db entries will be deleted
         Returns:
-            confirmation message
+            True if deletion was successful, otherwise False
         """
```
kaykurokawa: This function should probably not succeed if there are multiple matches that can be deleted. So probably call get_lbry_files() and than if there is more than one file, throw exception saying something like 'multiple matches detected, specify by file_name or another unique search value'

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1553,64 +1553,65 @@ def jsonrpc_get(
...
 
         timeout = timeout if timeout is not None else self.download_timeout
         download_directory = download_directory or self.download_directory
-        sd_hash = get_sd_hash(stream_info)
         if name in self.waiting_on:
-            # TODO: return a useful error message here, like "already
-            # waiting for name to be resolved"
-            defer.returnValue(server.failure)
+            log.info("Already waiting on lbry://%s to start downloading", name)
+            yield self.streams[name].data_downloading_deferred
```
kaykurokawa: I don't think this code handles updated names. 

I think we need to resolve the name first and check if we have the file with the sd_hash and or stream hash in the file manager

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: ManagedEncryptedFileDownloader.status() can take upwards of a second, and if you have many lbry files this substantially slows down an unfiltered `file_list`

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: I'll change this to being indexed by outpoint, then each entry in Daemon.streams is unique to a claim - independent of updates.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -92,6 +92,19 @@
 SHORT_ID_LEN = 20
 
 
+class IterableContainer(object):
```
lyoshenka: @jackrobison :crying_cat_face:

PR #516, opened 2017-03-08
Delete reveal and open command from api, not needed for electron

kaykurokawa:
Delete the "reveal" and "open" daemon commands from api, as they are not needed for electron.


jackrobison:
`start` and `reveal` in `lbrynet.core.file_utils`  are also good to delete

kaykurokawa:
putting this with https://github.com/lbryio/lbry/pull/510 so closing

Issue #527, opened 2017-03-13
test_double_download fails occasionally 


kaykurokawa:
test_double_download under test_misc fails occasionally, likely due to some race condition. Below is the error log. I believe the problem happens due to _delete_blobs_marked_for_deletion() is called in manage() when db_conn may be closed already. 

[ERROR]

Traceback (most recent call last):

  File "/home/travis/build/lbryio/lbry/lbrynet/core/BlobManager.py", line 200, in delete_from_db

    d = self._delete_blobs_from_db(b_hs)

  File "/home/travis/build/lbryio/lbry/lbrynet/core/sqlite_helpers.py", line 19, in wrapper

    d = f(*args, **kwargs)

  File "/home/travis/build/lbryio/lbry/lbrynet/core/BlobManager.py", line 309, in _delete_blobs_from_db

    return self.db_conn.runInteraction(delete_blobs)

exceptions.AttributeError: 'NoneType' object has no attribute 'runInteraction'

tests.functional.test_misc.TestTransfer.test_double_download

kaykurokawa:
Example of build failing here: https://travis-ci.org/lbryio/lbry/builds/209860687

kaykurokawa:
test_last_blob_retrieval also fails occasionally , like for related reason. Error below:

Traceback (most recent call last):

  File "/home/travis/build/lbryio/lbry/venv/local/lib/python2.7/site-packages/twisted/enterprise/adbapi.py", line 474, in _runInteraction

    conn.rollback()

  File "/home/travis/build/lbryio/lbry/venv/local/lib/python2.7/site-packages/twisted/enterprise/adbapi.py", line 52, in rollback

    self._connection.rollback()

sqlite3.OperationalError: cannot rollback - no transaction is active

tests.functional.test_misc.TestTransfer.test_last_blob_retrieval

https://travis-ci.org/lbryio/lbry/builds/215692820

jackrobison:
fixed in https://github.com/lbryio/lbry/pull/682

PR #534, opened 2017-03-17
fix double_download tests failing on BlobManager

kaykurokawa:
This is a temporary fix for https://github.com/lbryio/lbry/issues/527

I don't think there's any need to set self.db_conn to None here in BlobManager. 

I think we probably are not shutting down the BlobManager properly (self.db_conn.close() never gets called) , but since the BlobManager should persist for the lifespan of the app this temporary fix should be ok.


kaykurokawa:
hmm there seems to be some other problems in double_download

   test_double_download ... print_error(): Server error: 'NoneType' object has no attribute '__getitem__'

Happens in this pull request too

https://github.com/lbryio/lbry/pull/532#partial-pull-merging

kaykurokawa:
closing for now

PR #545, opened 2017-03-20
Create download directory if it doesn't exist

kaykurokawa:
Create a download directory (~/Downloads by default) if it doesn't exist


kaykurokawa:
It does refuse to start, since exception will be thrown there. 

But _shutdown() will throw some exceptions since it does not expect the Daemon to fail there , so needs to be fixed. 

Also there is two places where daemon sets up directories , in __init__() and _setup_data_directory() , which is unnecessary.

kaykurokawa:
Move the creation of directories to _setup_data_directory() 

Move the initialization of stream_info_manager (which requires db_dir) to _setup_lbry_file_manager()

Issue #549, opened 2017-03-21
Daemon multiple start tries not working properly 


kaykurokawa:
DaemonControl.py has a procedure to try starting DaemonServer multiple times if it fails the initial time. 

https://github.com/lbryio/lbry/blob/master/lbrynet/lbrynet_daemon/DaemonControl.py#L125

However, it does not seem to work because DaemonServer is not cleaned up before it is started again, and also because DaemonServer will call sys.exit(1) when it detects that the lbry api port is being used. Calling sys.exit(1) does not seem to kill the lbrynet-daemon process either, probably due to how twisted handles it  

The below logs are generated on the second attempt to restart 

=========================================

2017-03-21 13:08:37,287 INFO     lbrynet.lbrynet_daemon.DaemonControl:126: Making attempt 2 / 5 to startup
2017-03-21 13:08:37,290 INFO     lbrynet.lbrynet_daemon.DaemonServer:56: Using non-authenticated API
2017-03-21 13:08:37,291 INFO     lbrynet.lbrynet_daemon.DaemonServer:40: Daemon already running, exiting app
Unhandled error in Deferred:
2017-03-21 13:08:37,291 CRITICAL twisted:154: Unhandled error in Deferred:
2017-03-21 13:08:37,291 DEBUG    lbrynet.analytics.api:76: Sending track event: {'timestamp': '2017-03-21T17:08:37.291910Z', 'userId': 'lbry', 'event': 'Error', 'context': {'app': {'build': 'dev', 'version': '0.9.1', 'name': 'lbrynet', 'wallet': {'version': '2.7.12', 'name': 'lbryum'}, 'python_version': '2.7.12'}, 'os': {'version': '4.4.0-66-generic', 'name': 'Linux'}, 'library': {'version': '1.0.0', 'name': 'lbrynet-analytics'}}, 'properties': {'lineno': 154, 'module': '_legacy', 'message': 'Unhandled error in Deferred:', 'traceback': None, 'lbry_id': '8Aag8cGyqvNrFpMSbEsJ9e6GgGapDpvUE5ku7oS4CGyWPXfL3diyT65TxESHH1V9ba', 'session_id': '7PnoEL7kaCxkBZj9yzUaWBXqQzC9CSzthdDUpnCuc8GDUdF8f5YfP4xQ5V9JHPSCK2', 'name': 'twisted'}}

2017-03-21 13:08:37,292 CRITICAL twisted:154: 
Traceback (most recent call last):
[...]

PR #554, opened 2017-03-22
Fix restart procedure in DaemonControl

kaykurokawa:
Fix for https://github.com/lbryio/lbry/issues/549

Create a stop() function in DaemonServer, which closes the listening port and shuts down the Daemon. 
This allows a DaemonControl to restart the DaemonServer in case something goes wrong while starting up the Daemon ( you can't restart if the port is already binded, and if you don't shut down the Daemon before restarting, you'll have two Daemon's running at the same time)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ at anytime.
   *
 
 ### Fixed
-  *
+  * Fix restart procedure in DaemonControl
   *
   *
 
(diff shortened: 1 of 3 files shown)


lyoshenka:
If the server_port is bound, this will still try to start the daemon multiple times. Is that correct behavior, or should it fail right away if a `CannotListenError` is raised?

Issue #533, opened 2017-03-17
Fails GET requests


Boruch-Baum:
I just now downloaded the daemon, successfully curl'ed a "resolve_name" request and a "wallet_new_address" request, but can't get GET requests working.

### Steps to reproduce
1.  ./lbrynet-daemon &>./lbrynet-daemon.log &
2.  curl 'http://localhost:5279/lbryapi' --data '{"method":"get","params":{"name":"what"} }'

### Expected behaviour
Exit code 0, mp4 file appearing in my /home/---/Downloads folder.

### Actual behaviour
The command line issues a timeout message. The abbreviated log file is attached (It was created using awk '{$1=""; $2=""; print $0}' lbrynet-daemon.log | uniq -c > log2, so the first column indicates the number of consecutive duplicated messages)

## System Configuration
curl 'http://localhost:5279/lbryapi' --data '{"method":"version"}'
[
  {
    "lbrynet_update_available": false,
    "lbrynet_version": "0.9.0",
    "lbryum_version": "2.7.13",
    "os_release": "4.5.0-0.bpo.2-amd64",
    "os_system": "Linux",
    "platform": "Linux-4.5.0-0.bpo.2-amd64-x86_64-with-glibc2.3.4",
    "remote_lbrynet": "0.9.0",
    "ui_version": "not loaded yet"
  }
]

## Anything Else
1] Notice in the abbreviated log that the daemon is issuing thousands of messages: WARNING lbrynet.dht.protocol:262: Can't send data to dht: EWOULDBLOCK.

2] Please consider either moving your not-so-hidden configuration folders .lbrynet and .lbryum from users' home folders to ~/.config, or make their location configurable.

[log.txt](https://github.com/lbryio/lbry/files/849298/log.txt)

aaronfrary:
I had a very similar issue: same steps to produce, same timeout message, but a few minutes later the download restarts in the background and completes successfully. I've attached my logs for the session; hope this is helpful.

#### System Configuration
built against master at commit c1a5806
```
[
  {
    "build": "dev",
    "ip": "2601:19b:4401:7320:41ec:8f14:e547:2307",
    "lbrynet_update_available": false,
    "lbrynet_version": "0.9.1",
    "lbryum_version": "2.7.13",
    "os_release": "15.6.0",
    "os_system": "Darwin",
    "platform": "Darwin-15.6.0-x86_64-i386-64bit",
    "processor": "i386",
    "python_version": "2.7.13",
[...]

Issue #504, opened 2017-03-01
StandaloneBlobDownloader needs to timeout and clean itself up 


kaykurokawa:
I noticed that when trying to download a name associated with a sd_hash that does not exist in the network, it leaves behind a perpetually running ConnectionManager that is trying to download the sd_hash which it will never find. ( Try running "lbrynet-cli get name=10mb") 

I think from lbrynet.core.StreamDescriptor.download_sd_blob which calls lbrynet.core.client.StandaloneBlobDownloader needs to have a time out and clean it self up.

kaykurokawa:
The proper place for the timeout would be in the DownloadManager class, 
as it is used both by GetStream (used to download full files), and StandaloneBlobDownloader. 

This means that we should replace the timeout mechanism currently present in GetStream.

alyssaoc:
fixed by #507

PR #560, opened 2017-03-23
Increase download timeout

kaykurokawa:
I'm finding that the daemon get command will often timeout , and there has also been reports here https://github.com/lbryio/lbry/issues/533 that report the same thing. Testing download of "what" several times, I've found that it times out about 40% of the time. 

The download_timeout is the number of seconds to wait for before the first blob (that's not the sd_blob) is downloaded. Setting it to 60 seconds and testing download of what, it succeeded 100% of the time. Not sure if the timeout is long enough for everyone though. 

I believe this is happening now because we have more peers on the network than before with many of them not being reliable, and the ConnectionManager often finds an unreliable peer and wastes time connecting to them.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -162,7 +162,7 @@ ADJUSTABLE_SETTINGS = {
     'delete_blobs_on_remove': (bool, True),
     'dht_node_port': (int, 4444),
     'download_directory': (str, default_download_directory),
-    'download_timeout': (int, 30),
+    'download_timeout': (int, 60),
     'host_ui': (bool, True),
     'is_generous_host': (bool, True),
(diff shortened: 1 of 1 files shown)


jackrobison:
I think that instead of raising the timeout we should reset the counter after having downloaded the sd blob. Most of those timeouts are happening when the sd blob takes ~20 seconds to acquire, and the odds of timeout increase as the time to download the sd blob approaches 30 - no time is left for the data to start downloading.

kaykurokawa:
I'm seeing the same thing, and thats a good solution. Would need to implement this https://github.com/lbryio/lbry/issues/504

Not sure if this patch is ok for now. Is 60 seconds timeout too unresponsive?

kaykurokawa:
closing this for now

PR #564, opened 2017-03-24
Rename API command file_seed 

kaykurokawa:
"seeding a file" generally refers to the act of making a file accessible to peers in a p2p network. 

API command file_seed did not actually do that, the command is either for stopping or restarting a download. 

I renamed it to file_set_status but open to other suggestions. 

Also ran the gen_api_docs.py script , I see that there are some other API changes that needs this.


lyoshenka:
merged

PR #565, opened 2017-03-24
WIP: connection manager speed up

kaykurokawa:
Some things to help ConnectionManager go through possible peers faster : 

Timeout TCP connection faster (default timeout was 30 seconds, set to 15 seconds) so it spends less time on peers it cant connect to. 

The old manage() loop was called every second, and each loop connected to a single new peer. Change it so that the manage() loop is called every five seconds, and connects to five new peers (at most) in a single loop (basically connect in parallel instead of in sequence)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ at anytime.
   *
 
 ### Changed
-  *
+  * Improved ConnectionManager speed
   *
   *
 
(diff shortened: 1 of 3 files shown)


kaykurokawa:
Tested and saw that this improved the speed of connection manager by significant margin:  

Over ten trial runs on claim 'what', I got 6.9 seconds with this new pull request vs 25.1 seconds with old. (Speed to download the sd_blob and the first blob of the file).

During this test, 'what' had about 20 peers.

lyoshenka:
merged

PR #569, opened 2017-03-27
Remove update_metadata() function

kaykurokawa:
During updates, publishing was silently using metadata fields in the old claim and placing it in the new claim. This would cause a problem for example if you specified a key fee in the old claim , but did not in the new claim (indicating free content). 

Treat metadata specified in the update as the de-facto metadata.

Am i missing something here?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ at anytime.
   *
 
 ### Fixed
-  *
+  * Removed update_metadata function that could cause update problems
   *
   *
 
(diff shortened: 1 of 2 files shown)


jackrobison:
The idea was to only submit changes to the metadata and not require a full dictionary

kaykurokawa:
K, I think that's not necessary and this  can be merged.
I don't know why the test is failing, it is completely unrelated. Its likely an issue with the test itself: 

https://github.com/lbryio/lbry/issues/527

kaykurokawa:
reran the tests and now its passed, so its definitely an issue with the test itself.

PR #570, opened 2017-03-28
proposal of API commands changes for channels

kaykurokawa:
This is the proposed API command for channels 

channel_new 

        WORK IN PROGRESS
        Make a channel claim, or update if already claimed
        by user.
        Args:
            'channel_name': name of channel, must start with @
        Returns:
            (dict) Dictionary containing result of the channel claim
            {
                'tx' : (str) hex encoded transaction
                'txid' : (str) txid of resulting claim
                'nout' : (int) nout of the resulting claim
                'fee' : (float) fee paid for the claim transaction
                'claim_id' : (str) claim ID of the resulting claim
            }

In addition, below are changes to the current API commands for channel 

1. jsonrpc_publish will have new argument "channel_name" to specify which channel a publish belongs to. 

2. jsonrpc_resolve_name will have a new argument "validate_channel" (other alternatives : "validate" , "authenticate", "authenticate_channel" ) which will check the signature, if there is any, and validate it. It will return None if it does not validate. 

Other commands, such as claim_show, claim_list_mine, and claim_list will remain the same, returning channel related claims in the same format. I think this is good enough for now, there could be more new commands (i.e. command for validating signatures) but its better not to make drastic changes all at once.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -1856,6 +1856,27 @@ class Daemon(AuthJSONRPCServer):
         return d
 
     @AuthJSONRPCServer.auth_required
+    def jsonrpc_channel_new(self, channel_name):
+        """
+        WORK IN PROGRESS
+        Make a channel claim, or update if already claimed
+        by user.
+        Args:
+            'channel_name': name of channel, must start with @
+        Returns:
+            (dict) Dictionary containing result of the channel claim
+            {
+                'tx' : (str) hex encoded transaction
+                'txid' : (str) txid of resulting claim
+                'nout' : (int) nout of the resulting claim
+                'fee' : (float) fee paid for the claim transaction
+                'claim_id' : (str) claim ID of the resulting claim
+            }
+
+        """
+        pass
+
+    @AuthJSONRPCServer.auth_required
     def jsonrpc_get_transaction_history(self):
         """
         DEPRECATED. Use `transaction_list` instead


lyoshenka:
why does `jsonrpc_resolve_name` take a validate argument. if the name being resolved is part of a channel, it should automatically validate and return that info in the response. here's an example:

```
$ lbrynet-cli resolve_name name=what
{
  "author": "Samuel Bryan",
  "channel": {
    "name":"@lbry",
    "valid":true,
  }
  "content_type": "video/mp4",
  "description": "What is LBRY? An introduction with Alex Tabarrok",
  "language": "en",
  "license": "LBRY inc",
  "nsfw": false,
  "sources": {
    "lbry_sd_hash": "d5169241150022f996fa7cd6a9a1c421937276a3275eb912790bd07ba7aec1fac5fd45431d226b8fb402691e79aeb24b"
  },
  "thumbnail": "https://s3.amazonaws.com/files.lbry.io/logo.png",
  "title": "What is LBRY?",
  "ver": "0.0.3"
}
```

lyoshenka:
alternatively, don't even include channel information if its not valid. we can define "in a channel" to mean "has a channel name and a valid channel signature". so if the sig is invalid, its not actually in a channel at all.

not sure which is better here.

kauffj:
1. Should it return outpoint?
2. Can `resolve_name` move to `resolve` and officially support claim/outpoint resolution?

kaykurokawa:
We could just validate by default, either removing the argument or setting it as default value True. But do note that validation takes more time, and if you don't need the validation you will be slowed down. 

I don't really like that you are inserting the validation into the metadata.

It becomes confusing what fields are part of the original metadata and what fields are something the daemon injects into the metadata. Furthermore, someone could craft a metadata that can confuse the user that it has valid authentication.  For example setting "content_type" to "valid" or something like that. 

Validation should be much more explicit. If you want resovle_name to return the actual validation state it should at least be crafted as {"metadata": metadata of claim, "validation":validation_result}

kaykurokawa:
1. I'm using txid , nout  instead of claim outpoint to be consistent with the other commands. But yes ,I think we should be returning claim outpoint for this and all other commands ( publish, abandon, support) 

2. yes but out of scope for this.

lyoshenka:
Agreed that validation should not be mingled with metadata. Didn't mean to put it there.

Let's validate by default. It's probably what you want most of the time. If you want to add an option to disable validation, I'm cool with that.

lyoshenka:
That said, `resolve` should return claim info somewhere. Wherever that goes, the validation should go with it.

jackrobison:
What should [the claim counter](https://github.com/lbryio/lbryum-server/blob/integrate-lbry-schema/lbryumserver/blockchain_processor.py#L721) by called? I have it as `claim_n` at the moment.

jackrobison:
@kaykurokawa 

It's the nth claim to the name (nth claim id for that name).

I claim lbry://test, my claim has n = 1.
You claim lbry://test, your claim has n = 2.
I update my claim to lbry://test, it still has n = 1.

kaykurokawa:
then claim_n sounds ok

lyoshenka:
Sounds like we're good here. Please reopen if we need to discuss anything else.

PR #576, opened 2017-03-29
remove update_stream

kaykurokawa:
Same problem as https://github.com/lbryio/lbry/pull/569 , identical code found in Publisher, triggered in case user specified the source instead of file_path in jsonrpc_publish()

Publishing was silently using metadata fields in the old claim and placing it in the new claim. This would cause a problem for example if you specified a key fee in the old claim , but did not in the new claim (indicating free content).


kaykurokawa:
closing this is not quite correct

PR #579, opened 2017-03-30
Return full LBRY ID and installation ID in jsonrpc_status()

alexliebowitz:
Better to let the client decide whether to display the short or long version


lyoshenka:
@alexliebowitz changelog please

PR #580, opened 2017-03-30
Fix update_stream in publisher 

kaykurokawa:
Redo of https://github.com/lbryio/lbry/pull/576, I misunderstood what this code was doing. 

In Publisher class, 

publish_stream() was used to create a lbry file and make a claim, triggered when file_path was specified in jsonrpc_publish(). 

update_stream() was used to only make a claim, without creating a lbry file, triggered when no file_path was specified and the source was specified indicating already created lbry file in jsonrpc_publish().  It also incorrectly assumed that it was an update and would try to fetch a previous claim, and reuse the fields in the previous claim (same issue as https://github.com/lbryio/lbry/pull/569). Note that just because a source is specified, does not mean it is necessarily an update. 

Rename publish_stream() to create_and_publish_stream()

Rename update_stream() to publish_stream() ,  remove the code that fetched the previous claim and reused its field.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -716,9 +716,9 @@ class Daemon(AuthJSONRPCServer):
         if bid <= 0.0:
             raise Exception("Invalid bid")
         if not file_path:
-            claim_out = yield publisher.update_stream(name, bid, metadata)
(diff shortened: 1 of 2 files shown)


kaykurokawa:
merged with https://github.com/lbryio/lbry/pull/586

PR #586, opened 2017-04-03
Lbryschema integration 

kaykurokawa:
lbryschema integration

PR #587, opened 2017-04-03
Decode hex value from getvalueforname

jackrobison:
https://github.com/lbryio/lbryum/pull/87

-also use `getvalueforname`, which does the same thing as the multiple calls before


kaykurokawa:
I have included hex decoding of lbryum getvalueforname command in this pull request https://github.com/lbryio/lbry/pull/586 so this one isn't necessary.

PR #592, opened 2017-04-06
Fix inefficient wallet look ups during publishing

kaykurokawa:
When publishing, Publisher.create_and_publish_stream() would call load_file_attributes() in EncryptedFileDownloader which would look up file attributes using the wallet. Obtaining the claim id in particular is expensive since it results in  making a request to lbryum server 

Loading attributes using the wallet is not necessary because all the attributes are available in the create_and_publish_stream() function. Create load_file_attributes_from_arg() function which sets the attributes from the argument instead of looking it up. 

Without this pr, publishing will not work using lbryum server with channels/lbryschema integration due to https://github.com/lbryio/lbryum-server/issues/18.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ at anytime.
 
 ### Fixed
   * Removed update_metadata function that could cause update problems
-  *
+  * Fixed inefficient wallet look ups during publishing
   *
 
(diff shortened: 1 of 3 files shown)


kaykurokawa:
fixed in https://github.com/lbryio/lbry/pull/593

PR #593, opened 2017-04-07
Channels

jackrobison:
https://github.com/lbryio/lbryum/pull/84

  * Use `claim_id` instead of outpoint for `claim_abandon`
  * Add `channel_name` parameter to `publish`
  * Add `delete_all` parameter to `file_delete` to allow deleting multiple files
  * Add `channel_list_mine`
  * Add `channel_new`
  * Add `resolve` to resolve lbry uris
  * Use `uri` instead of `name` for `get`, remove explicit `claim_id` parameter
  * Increase default download timeout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,17 +9,19 @@ at anytime.
 
 ## [Unreleased]
 ### Added
-  *
-  *
-  *
+  * Use `claim_id` instead of outpoint for `claim_abandon`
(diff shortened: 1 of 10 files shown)


kaykurokawa:
jsonrpc_resolve_name() no longer returns None for claims that does not exist (unclaimed name)

jsonrpc_resolve() needs to clarify what should happen if a claim does not exist, currently it returns an unrelated exception.

PR #605, opened 2017-04-10
use appdirs 1.4.3 in requirements.txt

kaykurokawa:
having trouble installing when I have most recent appdirs installed and lbrynet requirements.txt forces a downgrade into 1.4.0 https://github.com/ActiveState/appdirs/issues/89

appdirs changelog https://github.com/ActiveState/appdirs/blob/master/CHANGES.rst


jackrobison:
LGTM

PR #608, opened 2017-04-12
raise errors before we get further into the publish process

jackrobison:


kaykurokawa:
Do not use wallet.balance it doesn't take into account reserved points , use wallet.get_balance()

PR #614, opened 2017-04-18
Fix stream_cost_estimate

kaykurokawa:
stream_cost_estimate will throw exception when claim value is non decodeable instead of returning None. 

Have it return None, and clarify this in the docstring.


jackrobison:
LGTM

kaykurokawa:
rebased to master and merging

PR #618, opened 2017-04-20
Update prerequisites for Debian based OSes

blackknight36:
The pyparsing module is required to build lbry.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,7 +25,7 @@ sudo pip install virtualenv
 On Ubuntu (we recommend 16.04), install the following:
 
 ```
-sudo apt-get install libgmp3-dev build-essential python2.7 python2.7-dev python-pip git python-virtualenv libssl-dev libffi-dev python-protobuf
+sudo apt-get install libgmp3-dev build-essential python2.7 python2.7-dev python-pip git python-virtualenv libssl-dev libffi-dev python-protobuf python-pyparsing
 ```
 
 If you're running another Linux flavor, install the equivalent of the above packages for your system.


lyoshenka:
@blackknight36 what OS did you need this for? We build the app on Ubuntu 16.04 and we don't use that package.

blackknight36:
@lyoshenka I'm running raspbian 8.0 on a raspberry pi3 device.

lyoshenka:
We don't officially support raspbian, so I'd rather not put that in the install doc. I'm happy to create a "LBRY on Raspberry Pi" wiki page if you want to write about what you did there?

blackknight36:
@lyoshenka There really isn't much I had to do to get lbry running on Raspbian other than installing the python-pyparsing package.  I haven't tested on other Debian installs but I'm pretty sure the requirements would be the same.

lyoshenka:
Can you add a check that only installs pyparsing if you're on a pi?

blackknight36:
@lyoshenka It would be simple enough to check /etc/os-release to see what platform is being used.  You could even source the file to make the data in it available to the install script.  For example:

source /etc/os-release
if [ "$NAME" == 'Raspbian GNU/Linux' ]
    then apt-get -y install python-pyparsing
fi

The install instructions also say "On Ubuntu" so it might make sense to have a separate section for Raspbian.

lyoshenka:
I ended up just adding a comment here: 803cd98fef587cd6349c3cea96fe7e02ecbb2287

Thanks for your help. If you're in our slack, ping me and I'll send you some credits.

blackknight36:
@lyoshenka Thanks.  You can send credits to baXRAYzyZ5P7mSguUyBQUvfNXTTDKk5ipJ.

PR #619, opened 2017-04-21
Adding functionality to check the balance of a given wallet.

wallermadev:


CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ at anytime.
   *
   *
 
+
+
+## [0.10.0rc3] - 2017-04-21
+### Added
+  * Added functionality to check the balance of a given address.
```
lyoshenka: new changes should always go into the Unreleased section at the top. they'll get moved into the next release automatically when we create it

build/requirements_base.txt
```diff
@@ -1 +1,29 @@
-THIS FILE GETS FILLED IN BY BUILD SCRIPT AND INCLUDED IN requirements.txt
```
lyoshenka: this file shouldn't be in git. that my bad, i deleted it in master.

wallermadev: Removed from repo.

CHANGELOG.md
wallermadev: Updated change log to add changed to unreleased section

lbrynet/core/Wallet.py
```diff
@@ -98,7 +102,8 @@ def get_claim_metadata_for_sd_hash(self, sd_hash):
             return defer.succeed(None)
 
     def update_claimid(self, claim_id, name, claim_outpoint):
-        self.claimids[(name, claim_outpoint['txid'], claim_outpoint['nout'])] = claim_id
+        self.claimids[
+            (name, claim_outpoint['txid'], claim_outpoint['nout'])] = claim_id
```
jackrobison: Can you remove this kind of whitespace? We typically only split up lines like this if they're longer than 100 characters or if it helps readability.

lbrynet/core/Wallet.py
```diff
@@ -738,20 +764,26 @@ def get_transaction(self, txid):
     def get_claim_metadata_for_sd_hash(self, sd_hash):
         return self._get_claim_metadata_for_sd_hash(sd_hash)
 
-    def get_balance(self):
-        return self.wallet_balance - self.total_reserved_points - sum(self.queued_payments.values())
+    def get_balance(self, address=None):
```
jackrobison: It's probably worth making a separate function - like `lbrynet.core.Wallet.Wallet.get_address_balance` - since `lbrynet.core.Wallet.Wallet.get_balance` returns a figure based on the payments queued in lbrynet, which isn't exactly in sync with the live balance from lbryum. It also might be nice to optionally return the unconfirmed balance. 

You'd make a placeholder function that returns `defer.fail(NotImplementedError())` in the Wallet base class, and the real `get_address_balance` function in the LBRYumWallet class.

Then, you'd make the the new functionality accessible via the lbrynet api in `lbrynet.lbrynet_daemon.Daemon.Daemon.jsonrpc_wallet_balance`

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ at anytime.
 ## [Unreleased]
 ### Added
   * Add `lbryschema_version` to response from `version`
-  *
+  * Added functionality to check the balance of a given address.
```
jackrobison: Please add new/changed api arguments here, like "add optional `address` argument to `wallet_balance` to check the balance of a given address"

jackrobison:
Awesome!

lbrynet/core/Wallet.py
wallermadev: Hey Jack, thanks for the feedback will take a look over it now and makes some changes.

CHANGELOG.md
wallermadev: Will do.

jackrobison:
LGTM

PR #620, opened 2017-04-24
Add claim caching to wallet storage

jackrobison:


lbrynet/core/Wallet.py
```diff
@@ -101,24 +119,86 @@ def update_claimid(self, claim_id, name, claim_outpoint):
...
-        try:
-            return defer.succeed(
-                self.claimids[(name, claim_outpoint['txid'], claim_outpoint['nout'])])
-        except KeyError:
-            return defer.succeed(None)
+    def get_claimid_for_tx(self, claim_outpoint):
+        result = None
+        for k, claim_id in self.claimids.iteritems():
+            if k[1] == claim_outpoint['txid'] and k[2] == claim_outpoint['nout']:
+                result = claim_id
```
lyoshenka: should add a `break` here so we don't keep looping

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
-        d.addCallback(
-            lambda _: self.db.runQuery(
-                "insert into name_metadata values (?, ?, ?, ?)",
-                (name, claim_outpoint['txid'], claim_outpoint['nout'], sd_hash)))
-        return d
+        txid, nout = claim_outpoint['txid'], claim_outpoint['nout']
+        record_exists = yield self.db.runQuery("SELECT COUNT(*) FROM name_metadata "
+                                               "WHERE name=? AND txid=? AND n=?",
+                                               (name, txid, nout))
+        if record_exists:
```
lyoshenka: im confused about this part. you're only inserting if the record already exists? is that intentional, or did you mean `if not record_exists`?

in either case, you should only need one query to do this, not two (two queries causes a potential race condition). look at `INSERT OR IGNORE` and `INSERT OR ROLLBACK` (depending on which behavior you want). here are the relevant docs: [one](https://sqlite.org/lang_insert.html), [two](https://sqlite.org/lang_conflict.html). I can help if you have questions.

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
-                "insert into claim_ids values (?, ?, ?, ?)",
-                (claim_id, name, claim_outpoint['txid'], claim_outpoint['nout'])))
-        d.addCallback(lambda _: claim_id)
-        return d
+        txid, nout = claim_outpoint['txid'], claim_outpoint['nout']
+        result = yield self.db.runQuery("SELECT COUNT(*) FROM claim_ids WHERE claimId=?",
+                                           (claim_id,))
+        is_update = result[0][0]
+        if is_update:
+            yield self.db.runOperation("UPDATE claim_ids SET name=?, txid=?, n=? WHERE claimId=?",
```
lyoshenka: same as above. one query, using `INSERT OR REPLACE` this time

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
-                "delete from claim_ids where claimId=? and name=? and txid=? and n=?",
-                (claim_id, name, claim_outpoint['txid'], UNSET_NOUT)))
-        d.addCallback(
-            lambda r: self.db.runQuery(
-                "insert into claim_ids values (?, ?, ?, ?)",
-                (claim_id, name, claim_outpoint['txid'], claim_outpoint['nout'])))
-        d.addCallback(lambda _: claim_id)
-        return d
+        txid, nout = claim_outpoint['txid'], claim_outpoint['nout']
+        result = yield self.db.runQuery("SELECT COUNT(*) FROM claim_ids WHERE claimId=?",
```
lyoshenka: db columns should be underscored. `claimId` should be `claim_id`

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
+                                        (claim_outpoint['txid'], claim_outpoint['nout']))
+        response = None
+        if result:
+            response = result[0][0]
+        defer.returnValue(response)
+
+    @rerun_if_locked
+    @defer.inlineCallbacks
+    def get_cached_claim(self, claim_id, check_expire=True):
+        r = yield self.db.runQuery("SELECT * FROM claim_cache WHERE claim_id=?", (claim_id, ))
```
lyoshenka: i think these two queries could be combined into one, using some kind of JOIN

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
+            _, _, seq, claim_address, height, amount, supports, raw, chan_name, valid, ts = r[0]
+            last_modified = int(ts)
+            supports = [] if not supports else json.loads(supports)
+            name, txid, nout = claim_tx_info[0]
+            if check_expire and (time.time() - last_modified) > CLAIM_CACHE_TIME:
+                log.debug("Cached claim expired: %s", claim_id)
+            else:
+                deserialized = ClaimDict.deserialize(raw.decode('hex'))
+                response = {
+                    "height": height,
```
lyoshenka: is this a duplicate of https://github.com/lbryio/lbry/pull/620/files#diff-eafee006192cca2ffbca5ba91dc83eeeR145 ? something feels wrong about returning this dict here (and above too). shouldn't we return an object? the function is called `get_cached_claim`. we should be returning a claim, not a dict with some arbitrary fields

lbrynet/core/Wallet.py
```diff
@@ -130,54 +210,213 @@ def create_tables(transaction):
...
+    def save_claim_to_cache(self, claim_id, claim_sequence, claim, claim_address, height, amount,
+                            supports, channel_name, signature_is_valid):
+        log.debug("Saving %s to cache", claim_id)
+        result = yield self.db.runQuery("SELECT COUNT(*) FROM claim_cache WHERE claim_id=?",
+                                        (claim_id, ))
+        is_update = result[0][0]
+        serialized = claim.serialized.encode("hex")
+        supports = json.dumps([] or supports)
+        now = str(int(time.time()))
+        if is_update:
```
lyoshenka: same as before - this should be one query. this pattern is called an `upsert`

lbrynet/core/Wallet.py
```diff
@@ -576,33 +813,57 @@ def _handle_claim_result(self, results):
...
-                formatted = yield self._handle_claim_result(claim)
-                claims_for_return.append(formatted)
-            resolve_results['claims_in_channel'] = claims_for_return
-            result = resolve_results
-        elif 'error' in resolve_results:
-            raise Exception(resolve_results['error'])
+    def resolve_uri(self, uri, check_cache=True):
+        cached_claim = yield self._storage.get_cached_claim_for_uri(uri, check_cache)
+        if cached_claim:
+            result = cached_claim
```
lyoshenka: just return the result here. there's no need to put everything else into an `else` block. returning early makes your code more readable

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -934,7 +934,7 @@ def _get_lbry_file_dict(self, lbry_file, full_status=False):
             size = None
             message = None
 
-        claim = yield self.session.wallet.get_claim(lbry_file.claim_id)
+        claim = yield self.session.wallet.get_claim(lbry_file.claim_id, False)
```
lyoshenka: what does `False` mean here? using `check_expires=False` is a lot clearer.

lbrynet/core/Wallet.py
```diff
@@ -130,54 +236,161 @@ def create_tables(transaction):
                                 "    name text, " +
                                 "    txid text, " +
                                 "    n integer)")
+            transaction.execute("CREATE TABLE IF NOT EXISTS claim_cache (" +
+                                "    row_id INTEGER PRIMARY KEY AUTOINCREMENT, " +
+                                "    claim_id TEXT UNIQUE NOT NULL, " +
```
lyoshenka: is claim_id actually unique here? shouldn't it be something like outpoint? or is this the cache for winning claims?

lyoshenka: is claim_id actually unique here? shouldn't it be something like outpoint? or is this the cache for winning claims?

lbrynet/core/Wallet.py
```diff
@@ -130,54 +236,161 @@ def create_tables(transaction):
...
+                                "    height INTEGER NOT NULL, " +
+                                "    amount INTEGER NOT NULL, " +
+                                "    supports TEXT, " +
+                                "    claim_pb TEXT, " +
+                                "    channel_name TEXT, " +
+                                "    signature_is_valid BOOL, " +
+                                "    last_modified TEXT)")
+            transaction.execute("CREATE TABLE IF NOT EXISTS uri_cache (" +
+                                "    row_id INTEGER PRIMARY KEY AUTOINCREMENT, " +
+                                "    uri TEXT, " +
```
lyoshenka: uri should be unique here, right?

lbrynet/core/Wallet.py
```diff
@@ -130,54 +236,161 @@ def create_tables(transaction):
...
+    @defer.inlineCallbacks
+    def get_cached_claim_for_uri(self, uri, check_expire=True):
+        result = yield self.db.runQuery("SELECT "
+                                        "claim.claim_id, cert.claim_id, uri_cache.last_modified "
+                                        "FROM uri_cache "
+                                        "INNER JOIN claim_cache as claim "
+                                        "ON uri_cache.cache_row=claim.row_id "
+                                        "LEFT OUTER JOIN claim_cache as cert "
+                                        "ON uri_cache.certificate_row=cert.row_id "
+                                        "WHERE uri=?", (uri, ))
```
lyoshenka: `WHERE uri` needs table name (e.g. `WHERE uri_cache.uri`)

lbrynet/core/Wallet.py
```diff
@@ -130,54 +236,161 @@ def create_tables(transaction):
...
-            lambda _: self.db.runQuery(
-                "insert into name_metadata values (?, ?, ?, ?)",
-                (name, claim_outpoint['txid'], claim_outpoint['nout'], sd_hash)))
-        return d
+        txid, nout = claim_outpoint['txid'], claim_outpoint['nout']
+        record_exists = yield self.db.runQuery("SELECT COUNT(*) FROM name_metadata "
+                                               "WHERE name=? AND txid=? AND n=?",
+                                               (name, txid, nout))
+        if not record_exists:
+            yield self.db.runOperation("INSERT INTO name_metadata VALUES (?, ?, ?, ?)",
```
lyoshenka: if this can't be one query for now, at least add a `TODO` line so we know to come back to it later

lbrynet/core/Wallet.py
jackrobison: claim id is unique here. it'd only be outpoint if we wanted to cache non-current updates for the same claim.

lbrynet/core/Wallet.py
jackrobison: yup, will change that

lbrynet/core/Wallet.py
jackrobison: done

PR #631, opened 2017-05-03
PR

wallermadev:
Daemon will not start if correct password is not provided on startup


wallermadev:
This need the version of lbryum in my PR to work correctly.

kaykurokawa:
Hi, here are some general PR guidelines : 

Please rebase commits so that you don't have commits like https://github.com/lbryio/lbry/pull/631/commits/d328203f183587e21d37dcc96fdce14af7405fb3 and https://github.com/lbryio/lbry/pull/631/commits/e038431287924df2f8af6fa62b9579c970059866

Each commit should have a reason for existing. If not they should be rebased. 
 
A PR should achieve a single goal. If your goal is to have the daemon wait for lbryum to be decrypted, do not change unrelated things. If you want other unrelated changes, they should be put in a separate PR. 

Do not change the requirements.txt to point to your repo. If a PR in lbry requires a PR in lbryum, just assume that they will be both be merged to master at the same time. 

Now regarding this PR: 

Let's start first by creating a Deamon API function that allows users to enter the password and set the wallet password. Otherwise there's no point in waiting for lbryum to be decrypted right? Here is where Daemon API functions are placed:  https://github.com/lbryio/lbry/blob/master/lbrynet/lbrynet_daemon/Daemon.py , API functions are functions that start with jsonrpc_ . 

Please note that each API function requires a doc string and they all use the same formatting so please follow that. 


API function for entering the password is pretty simple, it takes the password as input and it just stores in a variable in the Wallet class which can be retreived whenever a lbryum function requiring the password is needed. The difficult part is knowing when passwords are needed in lbryum and putting the password in the required commands, but we can work on this later.

API function for setting the password on a lbryum wallet will take the new password (twice to handle mistypes) and the old password as input. You can look at how lbryum sets passwords on the wallet here:  https://github.com/lbryio/lbryum/blob/master/gui/qt/main_window.py#L1868

lyoshenka:
Closing this for now. Please reopen once you rebase and address kay's comments.

PR #636, opened 2017-05-10
Fix publish race conditions

jackrobison:
When issuing several `publish` calls concurrently, a race condition in lbryum causes double spends (and thus rejected transactions). This fixes the race condition by queueing the api calls that result in transactions being broadcast to be run sequentially rather than concurrently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ at anytime.
 
 ## [Unreleased]
 ### Added
-  *
+  * Add decorator to support queueing api calls
   *
 
 ### Changed
@@ -21,9 +21,10 @@ at anytime.
   * Cache claims in wallet storage for use looking claims up by id or outpoint
   * Try to use cached claim info for `file_list`
   * Convert wallet storage to inlinecallbacks
+  * Improve internal name_metadata sqlite table
   
 ### Fixed
-  *
+  * Fix race condition in publish that resulted in claims being rejected when making many publishes concurrently
   *
 
 ### Deprecated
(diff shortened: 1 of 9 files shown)


kaykurokawa:
I believe this doesn't fix the race condition, it only causes the race condition to not occur.  I am able to double spend through lbryum by issuing sequential transaction commands to it through the lbryum daemon directly. 

When I looked into it last time, Lbryum updates its sent transaction when  receive_history_callback is called by synchronizer.add_history_response(), which is not triggered by broadcasting a transaction or creating it.

lyoshenka:
This is a strange way of doing the migration. What if the daemon is killed right after you drop the table? Now you just lost everything.

You should do this

```
CREATE TABLE IF NOT EXISTS tmp_name_metadata_table (name TEXT UNIQUE NOT NULL, txid TEXT NOT NULL, n INTEGER NOT NULL, sd_hash TEXT NOT NULL);

INSERT OR IGNORE INTO tmp_name_metadata_table (name, txid, n, sd_hash) SELECT name, txid, n, sd_hash FROM name_metadata;

DROP TABLE name_metadata;

ALTER TABLE tmp_name_metadata_table RENAME TO name_metadata;
```

All SQL, no need to open multiple connections or store anything in memory. The only way this fails is if the last statement does not execute, but then all you need to do is detect that the tmp table exists and the original doesnt, and run just the last statement.

lyoshenka:
Btw, if this can be fixed for real in lbryum-server, that would be way better.

PR #643, opened 2017-05-16
Batched uri resolution

jackrobison:
Goes with https://github.com/lbryio/lbry/pull/652 and https://github.com/lbryio/lbryum/pull/103

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -26,6 +26,7 @@ from lbrynet.core.sqlite_helpers import rerun_if_locked
 from lbrynet.interfaces import IRequestCreator, IQueryHandlerFactory, IQueryHandler, IWallet
 from lbrynet.core.client.ClientRequest import ClientRequest
 from lbrynet.core.Error import RequestCanceledError, InsufficientFundsError, UnknownNameError
+from lbrynet.core.Error import UnknownClaimID, UnknownURI
 
 log = logging.getLogger(__name__)
 
@@ -702,9 +703,13 @@ class Wallet(object):
 
         if 'error' in results:
             if results['error'] in ['name is not claimed', 'claim not found']:
-                raise UnknownNameError(results['error'])
-            else:
-                raise Exception(results['error'])
+                if 'claim_id' in results:
+                    raise UnknownClaimID(results['claim_id'])
+                elif 'name' in results:
+                    raise UnknownNameError(results['name'])
+                elif 'uri' in results:
+                    raise UnknownURI(results['uri'])
+            raise Exception(results['error'])
 
         if 'certificate' in results:
             try:
@@ -832,7 +837,10 @@ class Wallet(object):
             cached_claim = None
         if not cached_claim:
             claim = yield self._get_claim_by_outpoint(txid, nout)
-            result = yield self._handle_claim_result(claim)
+            try:
+                result = yield self._handle_claim_result(claim)
+            except (UnknownNameError, UnknownClaimID, UnknownURI) as err:
+                result = {'error': err.message}
         else:
             result = cached_claim
         defer.returnValue(result)
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -685,18 +685,6 @@ class Wallet(object):
         defer.returnValue(my_claim)
 
     @defer.inlineCallbacks
-    def get_claim_info(self, name, txid=None, nout=None, claim_id=None, check_expire=True):
-        if claim_id is not None:
-            results = yield self.get_claim(claim_id, check_expire)
-            if results['name'] != name:
-                raise Exception("Name does not match claim referenced by id")
-        elif txid is None or nout is None:
-            results = yield self.get_claim_by_name(name)
-        else:
-            results = yield self.get_claim_by_outpoint(ClaimOutpoint(txid, nout), check_expire)
-        defer.returnValue(results)
-
-    @defer.inlineCallbacks
     def _handle_claim_result(self, results, update_caches=True):
         if not results:
             raise UnknownNameError("No results to return")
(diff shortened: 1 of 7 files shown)


build/build.sh
```diff
@@ -32,6 +32,9 @@ if [ "$FULL_BUILD" == "true" ]; then
   python "$BUILD_DIR/set_build.py"
 fi
 
+pip install git+https://github.com/lbryio/lbryschema.git
```
lyoshenka: why is this here? requirements should be in requirements.txt

lbrynet/core/Wallet.py
```diff
@@ -702,9 +702,13 @@ def _handle_claim_result(self, results, update_caches=True):
 
         if 'error' in results:
             if results['error'] in ['name is not claimed', 'claim not found']:
-                raise UnknownNameError(results['error'])
-            else:
-                raise Exception(results['error'])
+                if 'claim_id' in results:
```
lyoshenka: this feels wrong

lbrynet/core/Wallet.py
```diff
@@ -798,27 +802,35 @@ def _handle_claim_result(self, results, update_caches=True):
...
-        if resolve_results and 'claim' in resolve_results:
-            claim_id = resolve_results['claim']['claim_id']
-        certificate_id = None
-        if resolve_results and 'certificate' in resolve_results:
-            certificate_id = resolve_results['certificate']['claim_id']
-
-        result = yield self._handle_claim_result(resolve_results, cached_claim is None)
-        if claim_id:
-            yield self._storage.save_claim_to_uri_cache(uri, claim_id, certificate_id)
+    def resolve(self, check_cache, page, page_size, *uris):
```
lyoshenka: its weird that `*uris` (the thing you care about most of the time) comes last here. it should be the first arg, and the rest should come after and have sane default values. so, something like this:

```
resolve(uris, page=1, page_size=20, check_cache=True)
```

Also, I know arrays are 0-indexed, but "page 0" makes no sense. Pagination starts on page 1.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -787,9 +787,10 @@ def get_est_cost_using_known_size(self, uri, size):
 
         cost = self._get_est_cost_from_stream_size(size)
 
-        resolved = yield self.session.wallet.resolve_uri(uri)
-        if 'claim' in resolved:
-            claim = ClaimDict.load_dict(resolved['claim']['value'])
+        resolved = yield self.session.wallet.resolve(True, 0, 10, uri)
```
lyoshenka: Please use named params here in `resolve(True, 0, 10, uri)`. `True`, `0`, and `10` are meaningless when you read this code. Or if you rearrange params as in the above comment, you only have to pass in the uris.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1459,19 +1461,22 @@ def jsonrpc_claim_show(self, name=None, txid=None, nout=None, claim_id=None):
...
     @defer.inlineCallbacks
     @AuthJSONRPCServer.flags(force='-f')
-    def jsonrpc_resolve(self, uri, force=False):
+    def jsonrpc_resolve(self, force=False, uri=None, uris=[]):
         """
         Resolve a LBRY URI
 
         Usage:
-            resolve <uri> [-f]
+            resolve (<uri> | --uri=<uri>) [<uris>...] [-f]
```
lyoshenka: this is weird. shouldn't it just be `resolve [-f] uris...`? What's the point of `--uri` here? Is this just to maintain BC? If so, its probably best to fix this by handling this in the function itself (e.g. if you get one uri and its of the form `uri=whatever`, then remove the `uri=`). This way people are not encouraged to use the bad usage.

lyoshenka: Then you also dont need the weird special case of one uri being resolved. You always return a dict of resolved uris, keyed by uri (though this does break BC and I don't see a clean fix for it off the top of my head).

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1587,9 +1573,13 @@ def jsonrpc_get(self, uri, file_name=None, timeout=None, download_directory=None
         timeout = timeout if timeout is not None else self.download_timeout
         download_directory = download_directory or self.download_directory
 
-        resolved = yield self.session.wallet.resolve_uri(uri)
+        resolved_result = yield self.session.wallet.resolve(True, 0, 10, uri)
```
lyoshenka: another place where the `page` and `page_size` could be avoided if they are made defaults. DRY, maaaaan :-)

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -2165,6 +2155,81 @@ def jsonrpc_claim_list(self, name):
         claims = yield self.session.wallet.get_claims_for_name(name)
         defer.returnValue(claims)
 
+    @AuthJSONRPCServer.auth_required
+    @defer.inlineCallbacks
+    def jsonrpc_claims_in_channel_list(self, page=0, page_size=10, uri=None, uris=[]):
```
lyoshenka: we talked about this being in `claim_list`. we should not have a `claims_in_channels_list` RPC command. its dirty.

tests/unit/lbrynet_daemon/test_Daemon.py
```diff
@@ -53,7 +54,7 @@ def get_test_daemon(data_rate=None, generous=True, with_fee=False):
             {"fee": {"USD": {"address": "bQ6BGboPV2SpTMEP7wLNiAcnsZiH8ye6eA", "amount": 0.75}}})
     daemon._resolve_name = lambda _: defer.succeed(metadata)
     migrated = smart_decode(json.dumps(metadata))
-    daemon.session.wallet.resolve_uri = lambda _: defer.succeed({'claim': {'value': migrated.claim_dict}})
+    daemon.session.wallet.resolve = lambda *_: defer.succeed({"test": {'claim': {'value': migrated.claim_dict}}})
```
lyoshenka: shouldn't we test resolving multiple uris too?

build/build.sh
jackrobison: it's because when `pip install git+...`'ing, pip only looks at setup.py, so if the thing you're installing also depends on a non-pypy package it needs to be installed first.

jackrobison: actually, since build.sh is just running `pip install -r requirements.txt` to install lbrynet, the lbryum branch should indeed be specified in requirements.txt

kaykurokawa:
What should happen if a name is unclaimed? 

Before this PR, it was returning None.

Right now it is  returning dict {u'error': u'Name lbry://someunclaimedname1 is unknown'}

I think either way is ok. But others should state their opinion and behavior should be clarified in the docstring.

lyoshenka:
As per the tech meeting: `[u'error': u'Name lbry://someunclaimedname1 is unknown'}` is better, Jack should update docstring.

lbrynet/core/Wallet.py
jackrobison: Added UnknownClaimID and UnknownURI exceptions.

lbrynet/lbrynet_daemon/Daemon.py
jackrobison: yeah, [-f] should be at the beginning.

BC aside, it's (<uri>) [<uris>...] also so that at least one uri is required

tests/unit/lbrynet_daemon/test_Daemon.py
jackrobison: Yes - we should be testing Wallet.resolve in general. This test is just doing get_est_cost, which only does one resolution

PR #651, opened 2017-05-26
Allow zero fee in fee_schemas

kaykurokawa:
If a claim specifies the fee field, but sets amount to zero, the daemon will error out (see lbry://the-handbook-of-technical-analysis) 

Allow claims to have a fee field with amount zero , and treat it as free content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ at anytime.
 
 ### Fixed
   *
-  *
+  * Allow fee fields in metadata with amount 0
   * Allow claim_show to be used without specifying name
 
 ### Deprecated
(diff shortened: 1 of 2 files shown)


jackrobison:
I think we should delete `lbrynet.metadata` altogether instead. The 0 fee case would previously have thrown an error in the old JSON schema, and I think it's only because I didn't port the JSON tests to lbryschema that it crept in.

I think it's correct to throw an error when publishing with a 0 fee - it's dead space in the claim and another address in the publisher's wallet. I think the handling of such claims should be in lbryschema, and probably should strip such a fee out instead of handling it upon resolve/download in lbrynet.

PR #652, opened 2017-05-28
Parse cli args from docstrings and add positional arguments

jackrobison:


kauffj:
The quickstart guide needs to be updated in parallel with merging this.

kaykurokawa:
There's no option to filter by 'name' for file_list and file_delete commands.

kaykurokawa:
Quickstart doesn't need to be adjusted since it uses JSONRPC, which this PR should not change. 

1cf3243
Some docstrings missing a closing > bracked so added it. 

9003369
file_set_status was using argument names that were different from file_list and file_delete so made it the same. Corrected some docstrings descriptions. Added name as a argument for parsing by claim name. 

257d025
Added integration testing to this branch. 

58dd576
Added a test command to the Daemon, for testing docopt arg parsing.
Added docopt arg parsing to the integration testing.

PR #653, opened 2017-05-29
add ispath, and file size check before publishing

kaykurokawa:
Check for file, and that file size is not zero before publishing


jackrobison:
LGTM

PR #657, opened 2017-05-29
remove lbrynet.metadata

jackrobison:


kaykurokawa:
this is a good time to revisit and improve test_ExchageRateManager so I'm doing that

kaykurokawa:
So fee amount can be set to zero with this patch right? Was that your intention? 

I made the below changes to improve testing: 

I broke out the various Dummy classes which only gets used for testing from ExchangeRateManager file and placed it in the tests/mocks.py file. DummyExchangeRateManager should also not redefine its own convert_currency function since we want to test convert_currency function in the actual ExchangeRateManager class. 

Added an assert in ExchangeRate class to make sure price is more than 0 

Added new tests:
Check if setting Fee to zero is allowed 
Check ExchangeRate can't be set with price less than 0 
Check that if a feed is missing to do conversion, an exception is thrown.

jackrobison:
>So fee amount can be set to zero with this patch right? Was that your intention?

When publishing an empty fee will get stripped out, empty fees shouldn't make it into the claim.

PR #667, opened 2017-06-02
Add check for old fee format and throw releveant exception

kaykurokawa:
deprecating old fee format in publish, so throw relevant exception for the old format


kaykurokawa:
fixed pylint

PR #670, opened 2017-06-03
Add claim address as an option to publish

kaykurokawa:
Add claim address as an argument to publish API command using lbryum claim_addr parameter.


jackrobison:
LGTM

PR #672, opened 2017-06-05
Fix file delete

kaykurokawa:
Fixed various issues in API command file_delete

file_delete should have been preventing deletion of multiple files by default but it was not doing so. Fix this, and make sure that if delete_all flag is True, it allows deletion of multiple files.

delete_target_file was not accurately describing what it was doing, it was described as "delete only from downloads, do not delete blobs". However, the flag will cause the file to be delete from downloads, and blobs are always deleted. delete_target_file should be by default False, since presumably the user will most likely just want to delete the blobs so he can stop seeding them but will want to keep the actual file. 

Added longer flag options, --delete_all  and --delete_target_file (can be used instead of -a and -f) 

Added tests to make sure that the flags are being processed properly in the CLI.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -2501,22 +2501,25 @@ class Daemon(AuthJSONRPCServer):
         defer.returnValue(response)
 
     @defer.inlineCallbacks
-    def jsonrpc_cli_test_command(self, pos_arg, pos_args=[], pos_arg2=None, pos_arg3=None):
+    @AuthJSONRPCServer.flags(a_arg='-a', b_arg='-b')
+    def jsonrpc_cli_test_command(self, pos_arg, pos_args=[], pos_arg2=None, pos_arg3=None,
+                                 a_arg=False, b_arg=False):
         """
         This command is only for testing the CLI argument parsing
         Usage:
-            cli_test_command (<pos_arg> | --pos_arg=<pos_arg>)
+            cli_test_command [-a] [-b] (<pos_arg> | --pos_arg=<pos_arg>)
                              [<pos_args>...] [--pos_arg2=<pos_arg2>]
                              [--pos_arg3=<pos_arg3>]
 
         Options:
+            -a, --a_arg                        : a arg
+            -b, --b_arg                        : b arg
             <pos_arg2>, --pos_arg2=<pos_arg2>  : pos arg 2
             <pos_arg3>, --pos_arg3=<pos_arg3>  : pos arg 3
-
         Returns:
             pos args
         """
-        out = (pos_arg, pos_args, pos_arg2, pos_arg3)
+        out = (pos_arg, pos_args, pos_arg2, pos_arg3, a_arg, b_arg)
         response = yield self._render_response(out)
         defer.returnValue(response)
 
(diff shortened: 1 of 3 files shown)


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1684,19 +1684,20 @@ def jsonrpc_file_set_status(self, status, **kwargs):
...
-            file_delete [-a | -f] [--sd_hash=<sd_hash>] [--file_name=<file_name>]
+            file_delete [-a] [-f] [--sd_hash=<sd_hash>] [--file_name=<file_name>]
                         [--stream_hash=<stream_hash>] [--claim_id=<claim_id>]
                         [--outpoint=<outpoint>] [--rowid=<rowid>]
                         [--name=<name>]
 
         Options:
-            -a                          : delete file from downloads and delete stored blobs
-            -f                          : delete only from downloads, do not delete blobs
+            -a, --delete_all            : if there are multiple matching files, allow the deletion
```
lyoshenka: can we get rid of the short `-a` option, and leave only the long option. i dont want anyone to accidentally delete all their files

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1684,19 +1684,20 @@ def jsonrpc_file_set_status(self, status, **kwargs):
...
+                                            of multiple files. Otherwise do not delete anything.
+            -f, --delete_target_file    : delete file from download directory, instead of just blobs
```
lyoshenka: this flag needs a better name. the endpoint is already called `file_delete`. having a flag called `--delete_target_file` is confusing. doesnt it already delete the file?

the delete_target_file variable could also be renamed

kaykurokawa: how about "delete_from_download_directory" ?

kaykurokawa:
delete_target_file renamed to delete_from_download_dir
-a option for delete_all removed

kaykurokawa:
rebased to master

kauffj:
Is this a behavior BC break? If so should really be a minor version increment.

lyoshenka:
Not strictly necessary: http://semver.org/#spec-item-4. Plus we break BC
regularly and we're not that consistent about it.

If we want to get more consistent, we can. But that will require more work
and planning.

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1542,27 +1542,30 @@ def jsonrpc_file_set_status(self, status, **kwargs):
 
     @AuthJSONRPCServer.auth_required
     @defer.inlineCallbacks
-    @AuthJSONRPCServer.flags(delete_target_file='-f', delete_all='-a')
-    def jsonrpc_file_delete(self, delete_target_file=True, delete_all=False, **kwargs):
+    @AuthJSONRPCServer.flags(delete_from_download_dir='-f', delete_all='--delete_all')
+    def jsonrpc_file_delete(self, delete_from_download_dir=False, delete_all=False, **kwargs):
```
jackrobison: can you note this in the changelog?

kaykurokawa:
Rebased to master and added additional changelog for API change

lyoshenka:
is this good to go? if so lets merge it. if not, whats next?

kaykurokawa:
jack will do a version bump and merge this

jackrobison:
rebased on master

Issue #673, opened 2017-06-06
InsufficientFundsError gets swallowed when calling get


kaykurokawa:
When calling get API command for a stream with key fee more than available, InsufficentFundsError is encountered but not returned, causing the CLI or JSONRPC command to hang until it time outs. 

2017-06-06 17:45:00,886 CRITICAL twisted:154: 
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/usr/local/lib/python2.7/dist-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/usr/local/lib/python2.7/dist-packages/lbrynet-0.10.4rc1-py2.7.egg/lbrynet/lbrynet_daemon/Downloader.py", line 179, in download
    raise err
InsufficientFundsError: Unable to pay the key fee of 7.07302659546

PR #674, opened 2017-06-07
Fix GetStream class

kaykurokawa:
18fec8e
deferToThread was used for check_fee_and_convert() but is unnecessary, I think before this call made network calls to check the exchange rate but is now no longer the case. 
 
a004e92
finished_deferred was being initialized as a Deferred that was never used, and was over written with a different Deferred at the end of download() function. If a user used finished_deferred before download() function was called, this would be erronenous. Initialize finish_deferred to None , so that a user cannot accidentally use this deferred before it is ready. 

This is the reason why the download success metric in metabase is always 0. 

f7bbf81
Exceptions throw in download() function would be swallowed since we yield to data_downloading_deferred in the start() function and do not yield on download(). Thus you'd end up catching the timeout exception that gets error backed by data_downloading_deferred and the Exception thrown in download() would be swallowed.  Fixes https://github.com/lbryio/lbry/issues/673

b98363a
Just removing unnecessary initialization arguments from GetStream 

b00a8ae
remove _running variable in GetStream. I don't see the point of this variable, I guess its to prevent multiple calls of download()? The daemon will never do that so its not needed. 

1f2c9b9
It is possible for download to be manually stopped before a single blob is downloaded. Account for this case and callback data_downloading_deferred

db37791
break up download function to make it testeable. initialize() function will get sd_hash and key_fee , _create_downloader() function creates downloader instance from sd_blob. 

ceec3a3
add simple unit test 

ee3036b
have start() function return downloader and finished_deferred so that user of GetStream knows where to get the finished_deferred from. 

1ee1c77
I think we do not want to errback finished_deferred, a Deferred that GetSream is not responsible for managing. Rather we should call downloader.stop() to stop the download.

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -660,17 +660,15 @@ class Daemon(AuthJSONRPCServer):
             self.analytics_manager.send_download_started(download_id, name, claim_dict)
 
             self.streams[claim_id] = GetStream(self.sd_identifier, self.session,
-                                               self.session.wallet, self.lbry_file_manager,
                                                self.exchange_rate_manager, self.max_key_fee,
                                                conf.settings['data_rate'], timeout,
                                                download_directory, file_name)
             try:
-                download = self.streams[claim_id].start(claim_dict, name)
-                lbry_file = yield download
-                f_d = self.streams[claim_id].finished_deferred
-                f_d.addCallback(lambda _: self.analytics_manager.send_download_finished(download_id,
-                                                                                        name,
-                                                                                        claim_dict))
+                lbry_file, finished_deferred = yield self.streams[claim_id].start(claim_dict, name)
+                finished_deferred.addCallback(
+                    lambda _: self.analytics_manager.send_download_finished(download_id,
+                                                                            name,
+                                                                            claim_dict))
                 result = yield self._get_lbry_file_dict(lbry_file, full_status=True)
                 del self.streams[claim_id]
             except Exception as err:
--- a/lbrynet/lbrynet_daemon/Downloader.py
+++ b/lbrynet/lbrynet_daemon/Downloader.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from twisted.internet import defer, threads
+from twisted.internet import defer
 from twisted.internet.task import LoopingCall
 
 from lbryschema.fee import Fee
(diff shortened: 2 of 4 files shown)


lbrynet/lbrynet_daemon/Downloader.py
```diff
@@ -49,17 +49,16 @@ def __init__(self, sd_identifier, session, wallet, lbry_file_manager, exchange_r
...
         self.exchange_rate_manager = exchange_rate_manager
         self.payment_rate_manager = self.session.payment_rate_manager
-        self.lbry_file_manager = lbry_file_manager
         self.sd_identifier = sd_identifier
         self.downloader = None
         self.checker = LoopingCall(self.check_status)
 
         # fired when the download is complete
-        self.finished_deferred = defer.Deferred(None)
+        self.finished_deferred = None
```
jackrobison: was this intentional?

lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -660,17 +660,15 @@ def _download_name(self, name, stream_info, claim_id, timeout=None, download_dir
             self.analytics_manager.send_download_started(download_id, name, stream_info)
 
             self.streams[claim_id] = GetStream(self.sd_identifier, self.session,
-                                               self.session.wallet, self.lbry_file_manager,
                                                self.exchange_rate_manager, self.max_key_fee,
                                                conf.settings['data_rate'], timeout,
                                                download_directory, file_name)
             try:
-                download = self.streams[claim_id].start(stream_info, name)
+                lbry_file = yield self.streams[claim_id].start(stream_info, name)
```
jackrobison: this will block adding the download-finished analytics event callback until after `start` returns, and the callback can't be added if finished_deferred is `None`.

lbrynet/lbrynet_daemon/Downloader.py
kaykurokawa: Yes. Note that finished_deferred is set to downloader.start() at the end of download() function. 

Meaning that this initialized defer doesn't do anything. And if you try to attach a callback to it before downloader.start() is called, which someone did in the Daemon, it never gets called back.

kaykurokawa:
rebased to master

lbrynet/lbrynet_daemon/Downloader.py
```diff
@@ -49,28 +51,30 @@ def __init__(self, sd_identifier, session, wallet, lbry_file_manager, exchange_r
...
-        self.wallet = wallet
         self.session = session
+        self.wallet = self.session.wallet
         self.exchange_rate_manager = exchange_rate_manager
         self.payment_rate_manager = self.session.payment_rate_manager
-        self.lbry_file_manager = lbry_file_manager
         self.sd_identifier = sd_identifier
         self.downloader = None
         self.checker = LoopingCall(self.check_status)
+        self.download_sd_blob = download_sd_blob
```
jackrobison: I don't think this attribute is necessary

jackrobison:
One tidbit, otherwise LGTM

kaykurokawa:
fixed self.download_sd_blob
rebased to single commit
rebased to master

PR #681, opened 2017-06-09
remove unused add_stream() function from Daemon

kaykurokawa:
Just removing this unused function add_stream() from Daemon


lyoshenka:
this doesnt need to be a PR, you can just merge it

PR #687, opened 2017-06-12
DaemonControl is using deprecated is_running API command

kaykurokawa:
DaemonControl is using deprecated is_running API command to check if there is already a daemon running. Use status API command instead.


lyoshenka:
this doesnt need to be a PR, you can just merge it

PR #690, opened 2017-06-12
add unique_contacts to status response

jackrobison:


lyoshenka:
as per slack, `_total_contacts` should be `_unique_contacts`. after that, lgtm

PR #698, opened 2017-06-15
Remove download_directory from get API command

kaykurokawa:
Removes the download_directory argument from API command get 

Users should change the download_directory with settings_set 

https://github.com/lbryio/lbry/issues/645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ at anytime.
 
 ### Removed
   * Remove `claims_in_channel` from `resolve` response
-  *
+  * Removed download_directory argument from API command get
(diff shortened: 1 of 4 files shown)


zestyr:
Wouldn't this remove the possibility for unique per-file download locations, a la "save as"?

kauffj:
Inclined to agree with @zestyr. Another example could be a programmatic use case that needs to sort assets into folders.

Looking at the code, it doesn't seem like there is much of a complexity cost either.

lyoshenka:
merged in c7b254ec887d11a0d3da52ff61aa1d510a048f3c

PR #699, opened 2017-06-15
add claim_signature_update and channel_update

jackrobison:


lbrynet/lbrynet_daemon/Daemon.py
```diff
@@ -1694,6 +1694,39 @@ def jsonrpc_channel_new(self, channel_name, amount):
 
     @AuthJSONRPCServer.auth_required
     @defer.inlineCallbacks
+    @AuthJSONRPCServer.flags(revoke="-r")
+    def jsonrpc_channel_update(self, name, amount=None, revoke=False):
+        """
```
kaykurokawa: name should be 'channel_name' to be consistent with jsonrpc_channel_new

kaykurokawa:
I can't get channel_update command to work with revoke=False 

https://github.com/lbryio/lbryum/blob/master/lib/commands.py#L1605

Looks like you need to supply a val value to lbryum ? 

I'm going to force you to write tests for these on lbry-in-a-box. You can write the tests under _test_channels() in integration_testing.py

https://github.com/lbryio/lbry-in-a-box/blob/master/integration_testing.py#L539

You can make a PR with the tests on the lbry-in-a-box repo. To get travis tor run the integration test on the PR, change the line on .travis.yml that reads 'bash sync_to_master.sh' to 'bash sync_to_master.sh claim-signature-update master' which will use this branch for lbrynet and master for lbryum

lyoshenka:
@jackrobison how much work is left here?

eukreign:
Can this be closed and re-opened as a new issue in the new wallet?

jackrobison:
@eukreign sure, if still relevant

Issue #693, opened 2017-06-14
Functional tests fail occasionally 


kaykurokawa:
https://travis-ci.org/lbryio/lbry/builds/242981624?utm_source=github_status&utm_medium=notification

![screenshot from 2017-06-14 16-39-15](https://user-images.githubusercontent.com/7075051/27153617-26e49672-5120-11e7-985d-710637df01fc.png)

![screenshot from 2017-06-14 16-39-31](https://user-images.githubusercontent.com/7075051/27153628-2f0d09ce-5120-11e7-8294-9e070ff75e47.png)

Related issue:
https://github.com/lbryio/lbry/issues/675
https://github.com/lbryio/lbry/issues/527

kaykurokawa:
I still think this is because BlobManager's stop() function doesn't properly handle shutting down the manage() loop and closing the database properly

kaykurokawa:
Looking at BlobManager, it seems like it does not even need a looping manage() function. 

When deleting blobs, BlobManager queues them up for deletion, and then the manage() loops looks in the queue and deletes it. I don't think we need to queue up blobs for deletion, we can just delete them.

PR #704, opened 2017-06-16
Fix blobmanager manage

kaykurokawa:
44e7a04
BlobManager has a manage() function that is scheduled to run every second. The only thing it does is delete blobs marked for deletion by the delete_blobs() function. There is no really good reason why we can't just delete the blobs right away. Remove the manage() function and simplify blob deletion function by using inline callbacks.
 
1fed5f9
Added unit tests for BlobManager. 

4400d51
In test_misc.test_double_download() function test, check_lbry_file() was being called after delete_lbry_file(). 
Since we now delete blobs right away instead of scheduling them, this test was failing. Call check_lbry_file() after delete_lbry_file().

Fixes https://github.com/lbryio/lbry/issues/693

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ at anytime.
   * Fixed some log messages throwing exceptions
   * Fix shutdown of the blob tracker by Session
   * Fixed claim_new_support docstrings
+  * Fixed BlobManager causing functional tests to fail, removed its unneeded manage() loop
 
 ### Deprecated
   *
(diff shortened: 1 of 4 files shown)


lyoshenka:
Looks good to me.

FYI someone told me that if you put `Fixes #693` in your commit message, Github will automatically close the issue when the commit is merged. Not sure if this is true or not.

kaykurokawa:
rebased to master

PR #705, opened 2017-06-18
Fix default directories (fixes #574)

### LINKED ISSUE #574: Downloads are not XDG_DOWNLOAD_DIR compliant
### Steps to reproduce
1. Change the default download location on your system by setting the environment variable "XDG_DOWNLOAD_DIR" to something other than "~/Downloads"
2. Download any file through lbry, for example, by running `curl 'http://localhost:5279/lbryapi' --data '{"method":"get","params":{"name":"what"} }'`

### Expected behaviour
The file should be downloaded to the directory specified by "XDG_DOWNLOAD_DIR"

### Actual behaviour
The file is downloaded to "~/Downloads"

## System Configuration

 

    "build": "release",
    "ip": "Could not determine IP",
    "lbrynet_update_available": false,
    "lbrynet_version": "0.9.1",
    "lbryum_version": "2.7.13",
    "os_release": "4.10.3-1-ARCH",
    "os_system": "Linux",
    "platform": "Linux-4.10.3-1-ARCH-x86_64-with-glibc2.3.4",
    "processor": "",
    "python_version": "2.7.12",
    "remote_lbrynet": "0.9.1"

## Anything Else

This behaviour seems to stem from `lbry/lbrynet/conf.py`

zestyr:
When I installed lbry today I noticed it created a `Downloads` folder in my Linux home directory. As I've modified the default [xdg-user-dirs](https://freedesktop.org/wiki/Software/xdg-user-dirs/), I assumed lbry doesn't comply to the freedesktop standards. I checked the source code and found #574, so I wrote a fix.

Most importantly, this pull request moves the lbry directories out of the home directory, and into where they should be on Linux. 

In #574 a fix was suggested that included the use of GLib. It works, but it requires you to have that installed, and since `lbry` seemingly only supports python2, I couldn't get it to work. Instead I chose to write a couple lines more and not use a library.

For Linux, the new default locations are as follows (using the appdirs library):
```
default_data_dir = ~/.local/share/lbry/lbrynet
default_lbryum_dir = ~/.local/share/lbry/lbryum
```
To determine the correct downloads folder, we first look for `XDG_DOWNLOAD_DIR` in the `user_dirs.dirs` file which most likely exists, as it's created by default on pretty much every Linux distribution as part of the XDG standard. If this fails, the environment variable is checked, and if that's empty we simply default to `~/Downloads`.

I've also deleted some old seeming redundant code for creating some directories, they even had TODO comments about being duplicates. I did a tiny bit of linting, renamed the `default_downloads_directory` variable for consistency and readability, and a few other minor changes that should make some parts more readable.

I've tested the changes on two Linux machines, a macOS VM and a Windows machine, everything seems to work just fine.

I'm fully aware this is sort of a "breaking" change, since the folders wouldn't be moved automatically, leaving Linux users with a seemingly blank state after updating. I'd like to hear what you think about that, I'm open to any suggestions.

LBRY is an amazing project and this is likely not the last pull request you'll see from me, keep up the good work!


filipnyquist:
This is quite a game breaking bug for people that already have the app installed. Otherwise it seems good.

zestyr:
I somewhat agree, I also touched on that in the description. However, people can simple move the folders to their new locations, as I wrote in the changelog.

I agree because not everyone will check the changelog and such, so there needs to be a way to get it out there. I wouldn't mind creating something that checks if the old folders exist and the new ones doesn't, and then copying over the old folders to the new location, but I'm not sure that's a good way to do it.

filipnyquist:
That sounds like a good idea, waiting on some input from the core team :)

kauffj:
I support this as a change, but agree we're going to need to cleanly handle this for existing installs. Since the download directory is a setting value, wouldn't it be possible to only use the new logic when first setting value is empty/new?

Also, we should add a comparable check in Windows (and macOS?).

zestyr:
Not sure what you mean by when the value is empty/new. The download directory setting can be left unchanged, and it seems like it acts as a configuration setting just like the other two, with the difference being that you can modify it. 

I didn't change the Windows or macOS locations, so I don't think a check for them would do anything. I changed `lbry` to `LBRY` for consistency, but NTFS is case-insensitive, at least by default on Windows.

kaykurokawa:
This needs to be accompanied with one of the three options 

a) UI warnings that tells users to migrate their folder and instructions on how to do so. 
b) Automatic migration that moves the files to the new location 
c) If it detects that it was using the previous default folder configuration (basically it is not a fresh install with this PR) it uses the previous default folder configuration. Otherwise (its a fresh install with this PR) it uses the new configuration.

In b) and c), it needs to know about the previous default folder configuration. Thus, we need a function like old_default_dir() that resolves to ~/Downloads etc... and new_default_dir() for this PR. 

In a) we need a detailed summary for users on what to do for each platform.

kauffj:
C is by far the preferred solution IMO.

zestyr:
Alright, I believe I've implemented option C, though not the functions @kaykurokawa mentioned. If those are still wanted I can implement that, but everything appears to work fine. 
I've tested multiple configurations, both with a clean state, and using the old locations, and the new locations.

lbrynet/conf.py
```diff
@@ -43,25 +44,49 @@
...
+        default_download_dir = os.path.expanduser('~/Downloads')
+    else:
+        default_data_dir = user_data_dir('LBRY/lbrynet')
+        default_lbryum_dir = user_data_dir('LBRY/lbryum')
+        try:
+            with open(os.path.join(user_config_dir(), 'user-dirs.dirs'), 'r') as xdg:
+                down_dir = re.search(r'XDG_DOWNLOAD_DIR=(.+)', xdg.read()).group(1)
+                down_dir = re.sub('\$HOME', os.getenv('HOME'), down_dir)
+                default_download_dir = re.sub('\"', '', down_dir)
+        except:
```
lyoshenka: except is too broad. please specify error type(s)

lbrynet/conf.py
```diff
@@ -43,25 +44,49 @@
...
-        get_path(FOLDERID.RoamingAppData, UserHandle.current), 'lbryum')
+    default_download_dir = get_path(FOLDERID.Downloads, UserHandle.current)
+
+    # This checks whether the folders are still in their old locations
+    if os.path.isdir(user_data_dir('lbryum', 'lbry', roaming=True)) and \
+       os.path.isdir(user_data_dir('lbrynet', 'lbry', roaming=True)):
+        default_data_dir = user_data_dir('lbrynet', 'lbry', roaming=True)
+        default_lbryum_dir = user_data_dir('lbryum', 'lbry', roaming=True)
+    else:
+        default_data_dir = user_data_dir('lbrynet', 'LBRY')
```
lyoshenka: please user lowercase `lbry` for dir names

lbrynet/lbrynet_daemon/daemon_scripts/Autofetcher.py
```diff
@@ -1,21 +1,13 @@
...
 
-if sys.platform != "darwin":
-    log_dir = os.path.join(os.path.expanduser("~"), ".lbrynet")
-else:
-    log_dir = user_data_dir("LBRY")
-
-if not os.path.isdir(log_dir):
-    os.mkdir(log_dir)
-
+log_dir = conf.settings['data_dir']
```
lyoshenka: you have to call `conf.initialize_settings()` before using conf settings

lyoshenka:
Will gladly merge this if you fix the above. Thanks for the detailed discussion/PR :-)

lbrynet/conf.py
zestyr: Sure, I was just going by the way you handled the data location on Mac

lyoshenka: unless uppercase is the standard here? lowercase feels right to me and its what we've done elsewhere.

zestyr:
Alright, that should be about it. Thanks :)

lbrynet/conf.py
zestyr: I guess it depends, Microsoft tends to love PascalCasing. If the program's name is LBRY, I guess the standard is the name the folder LBRY as well

kaykurokawa:
Its worth consider if we really want to move our data directories to

default_data_dir = ~/.local/share/lbry/lbrynet
default_lbryum_dir = ~/.local/share/lbry/lbryum

The standard practice in cryptocurrency related project is to have them in ~/.cryptocurrency_name

zestyr:
That would be breaking an important standard, and littering people's home directories with folders that belong elsewhere.
Sia puts its things under `~/.config/Sia-UI`, which isn't correct either

jackrobison:
@zestyr 

I think you should split up the migration into two PRs: the one here (for migrating only the lbrynet data folder), and another in lbryio/lbryum. The migration would have to handle cases where it is canceled midway through or where it otherwise fails, since it could otherwise lead to a lbryum wallet being lost, which would be no good.

zestyr:
There is no migration @jackrobison, so I'm not sure what you mean. If it's not a fresh install, or the user hasn't moved the folders manually, it simply falls back to the old folder locations.

filipnyquist:
The standard for most crypto project is what Kay describes! (For every crypto I've tried)

zestyr:
I understand, but in my opinion open standards for a given operating system are more important than following the mistakes made by similar projects. If you all disagree, fair enough, but then I'd like to turn it into a configurable option like the downloads directory.

lyoshenka:
@kaykurokawa @filipnyquist you have a good point that we should consider how other crypto projects do it. In this case, I think we're better off doing what's right rather than following the mistake that many projects make, so I'd like to merge this.

@zestyr did you test this on windows and mac?

zestyr:
@lyoshenka yep, all works as expected :)

lyoshenka:
great. could you please fix the merge conflicts and I'll merge it

zestyr:
Done @lyoshenka

PR #711, opened 2017-06-19
adding unit tests for GetStream

kaykurokawa:
Just adding some more tests for GetStream class , no change to the class itself


kaykurokawa:
need to fix this for some changes

PR #712, opened 2017-06-19
[WIP] Daemon interface description for wallet encryption

kaykurokawa:
This PR contains the daemon interface that should be implemented in order to support wallet encryption.  Below is the interface: 

Suggestions to changes are welcomed.

    def jsonrpc_password_set(self, pw, pw_verify, old_pw=None):
        """
        Encrypt the wallet with password pw, and pw_verify must equal pw.
        If wallet encryption passsword was set previously, old_pw
        must be set to the previous encryption password.

        Usage:
            password_set (<pw> | --pw=<pw>) (<pw_verify> | --pw_verify=<pw_verify>)
                         [<old_pw> | --old_pw=<old_pw>]

        Options:
            <old_pw>, --old_pw=<old_pw>        : old password if password has been set before

        Returns:
            (bool) True if setting password was succesful, False otherwise

        """
        pass

    def jsonrpc_password_enter(self, pw):
        """
        If wallet password has been set with API command password_set,
        use this command to enter the password. It will attempt to
        decrypt the wallet with this password and return True if it succeeds.

        If decryption is succesful, it will store the password until the daemon
        closes for any subsequent operation that require acess to the wallet.

        The daemon will pause startup until this command is used if
        the password has been set.

        Usage:
            password_enter (<pw> | --pw=<pw>)

        Returns:
            (bool) True if the entered password was correct, False otherwise

        """
        pass

### DIFF
--- a/lbrynet/lbrynet_daemon/Daemon.py
+++ b/lbrynet/lbrynet_daemon/Daemon.py
@@ -2542,6 +2542,49 @@ class Daemon(AuthJSONRPCServer):
         response = yield self._render_response(mean_availability)
         defer.returnValue(response)
 
+
+    def jsonrpc_password_set(self, pw, pw_verify, old_pw=None):
+        """
+        Encrypt the wallet with password pw, and pw_verify must equal pw.
+        If wallet encryption passsword was set previously, old_pw
+        must be set to the previous encryption password.
+
+        Usage:
+            password_set (<pw> | --pw=<pw>) (<pw_verify> | --pw_verify=<pw_verify>)
+                         [<old_pw> | --old_pw=<old_pw>]
+
+        Options:
+            <old_pw>, --old_pw=<old_pw>        : old password if password has been set before
+
+        Returns:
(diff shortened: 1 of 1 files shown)


lyoshenka:
Let's stick with our `noun_verb` pattern and call the endpoints `jsonrpc_wallet_encrypt` and `jsonrpc_wallet_decrypt`. Decrypt may imply that the wallet is no longer encrypted, so I'm open to another verb here (unlock?)

Why do you have `pw` and `pw_verify` on encrypt. If this is an API, its just as easy to include a string twice as it is to include it once. If we're gonna go with the "enter it twice" route, we should get it interactively after the command is run.

The wallet password should be handled securely. Idk what the current best practices are for storing keys in memory. Whoever implements this, please read about it and do the right thing.

I know this an MVP, but I like how lbrycrdd asks you for a time and only keeps the wallet decrypted for that period of time. That would be a nice-to-have feature.

lyoshenka:
I'm going to close this so we don't have an outstanding PR waiting for us.

Issue #716, opened 2017-06-20
Daemon hangs after timeout on get request


billbitt:
## The Issue
After making a 'get' request that times out, when the daemon reaches the end of the timeout it hangs without responding to the curl request.

### Steps to reproduce
1. open the daemon
2. use curl to perform a 'get' request on a url that will time out. Example: 
`$ curl 'http://localhost:5279/lbryapi' --data '{"method":"get","params":{"uri":
"fucksiacoin"}}'`
3. wait for the deamon to timeout

### Expected behaviour
I expect to receive a response from the daemon once it times out

### Actual behaviour
daemon just hangs and does not respond. (see attached screenshots)

## System Configuration

$ curl 'http://localhost:5279/lbryapi' --data '{"method":"version"}'
{
  "id": null,
  "jsonrpc": "2.0",
  "result": {
    "build": "release",
    "ip": "********",
    "lbrynet_version": "0.13.1",
    "lbryschema_version": "0.0.7",
    "lbryum_version": "2.8.4",
    "os_release": "7",
    "os_system": "Windows",
    "platform": "Windows-7-6.1.7601-SP1",
    "processor": "Intel64 Family 6 Model 61 Stepping 4, GenuineIntel",
    "python_version": "2.7.13"
  }
}

## Screenshots

![screenshot 2017-06-19 19 12 24](https://user-images.githubusercontent.com/19243959/27313625-b0f1815c-5523-11e7-964b-40d4f8b88405.png)

billbitt:
follow up... if you do a new curl request after the hang, the result is as follows:

second curl request:
`$ curl 'http://localhost:5279/lbryapi' --data '{"method":"get","params":{"uri":
"fucksiacoin"}}'
{
  "id": null,
  "jsonrpc": "2.0",
  "result": null
}`

new output in daemon:
`2017-06-19 19:39:26,036 INFO     lbrynet.core.Wallet:809: Resolving fucksiacoin`
`2017-06-19 19:39:26,503 INFO     lbrynet.lbrynet_daemon.Daemon:1511: Already waiting on lbry://fucksiacoin to start downloading`

kaykurokawa:
Can confirm that if get timeouts before it downloads a sd blob, the daemon will hang.

billbitt:
same thing is happening when getting the claim for `boni#8315fa0b7a3ba2b7806a4411214b0bf5bfc31677`

PR #724, opened 2017-06-21
Add timeout to download_sd_blob

kaykurokawa:
Fixes https://github.com/lbryio/lbry/issues/716 and https://github.com/lbryio/lbry/issues/504 by adding a timeout argument to function download_sd_blob(). The timeout detects if the blob has been downloaded within a specified time, and if not, it throws DownloadTimeoutError and cleans everything up. 

As a side effect, the timeout behaviour when calling API command get is a bit different now.
Now it will wait 180 seconds (or whatever the timeout is set to) for the sd blob to download, and if it does , waits another 180 seconds for the first blob to download. Before it waited 180 seconds for both the sd blob and first blob. 

f1845c1
In SingleProgressManager, which we will add the timeout mechanism, use twisted's LoopingCall mechanism instead of reactor.callLater to do looping calls as it is much simpler. safe_start_looping_call and safe_stop_looping_call is added to lbrynet.utils and should be used by other classes which need looping calls. 

ccb7991
Add timeout do download_sd_blob(). Make SingleProgressManager , which checks when a single blob is downloaded, also responsible for keeping track of the timeout. Create DownloadTimeoutError in lbrynet.Error to use as a standard exception for indicating timeouts. 

 a5f4396
Use the newly added timeout mechanism for download_sd_blob in GetStream class. Note that GetStream still has its own timeout mechanism for checking when a single blob (that is not the sd blob) is downloaded. Getstream's timeout mechanism is started after download_sd_blob() is finished. 

345a1f2
Just breaking up GetStream for unit testing 

 6995169
use DownloadTimeoutError instead of generic Exception 

 0452338
use safe_start_looping_call in lbrynet.utils 

7ee0489
added unit tests for GetStream

eba854a
Use timeout mechanism in download_sd_blob() instead of creating its own in Daemon._download_blob()

 f058f18
This is a fix for test_double_download in function/test_misc. It was revealed in test_double_download that when starting the DownloadManager, progress_manager.start() could stop the download (due to it already having the blobs), but then after it would call connection_manager.start() and leave the ConnectionManager running (the tests would then fail as Reactor is unclean).

So have ConnectionManager start up first , and then start the progress manager. Also simplify the resume_downloading() and stop_downloading() command by using inline callbacks instead of Deferred lists.

Note also here the complexity and exception swallowing in CryptStreamDownloader that uses resume_downloading() and stop_downloading(). This should be fixed in a future PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ at anytime.
   *
 
 ### Fixed
-  *
+  * Fixed timeout behaviour when calling API command get
   *
 
 ### Deprecated
--- a/lbrynet/core/Error.py
+++ b/lbrynet/core/Error.py
@@ -9,6 +9,10 @@ class DuplicateStreamHashError(Exception):
 class DownloadCanceledError(Exception):
     pass
 
+class DownloadTimeoutError(Exception):
+    def __init__(self, download):
+        Exception.__init__(self, 'Failed to download {} within timeout'.format(download))
+        self.download = download
 
 class RequestCanceledError(Exception):
     pass
--- a/lbrynet/core/StreamDescriptor.py
+++ b/lbrynet/core/StreamDescriptor.py
@@ -236,7 +236,7 @@ class StreamDescriptorIdentifier(object):
         return d
 
 
-def download_sd_blob(session, blob_hash, payment_rate_manager):
+def download_sd_blob(session, blob_hash, payment_rate_manager, timeout=None):
     """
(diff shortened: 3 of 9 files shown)


kaykurokawa:
rebased to master

jackrobison:
Nice, LGTM

Can you rebase (again) and merge? @kaykurokawa

kaykurokawa:
rebased and merging

Issue #644, opened 2017-05-21
claim_show can return incorrect claim if name and outpoint are provided


kauffj:
(**Only number 1 exists, number 2 is fixed.**)

A few issues with `claim_show`:

1) `claim_show` requires name, which it then ignores if txid/nout or claim_id is provided. Example:

`./lbrynet-cli claim_show name=foo txid=e43ff68bb91df13df57b87e6a43c70f5c584e88feacf75789308b9c02ad2657b nout=0`

This returns a claim for name "chillin" (which is correct for the outpoint), but should not be returned if I provide the name "foo".

2) **(FIXED)** `claim_show` should not require a name if it's possible to resolve by claim_id and/or outpoint directly. I should be able to run:

`/lbrynet-cli claim_show txid=e43ff68bb91df13df57b87e6a43c70f5c584e88feacf75789308b9c02ad2657b nout=0`

But it currently tells me a name is required.

kaykurokawa:
https://github.com/lbryio/lbry/pull/647 fixes 2, 

For 1, do you need it to filter by multiple arguments? (for example by both name and claim ID) 
Or should it only allow you to filter by one thing only , to be less confusing, (throw exception if you specify both name and claim id)

kauffj:
I think it should be possible to filter by name and claim_id. When both are provided, you are essentially saying find a claim with this id iff it has this name. An example use case would be spee.ch, where it will support URLs like spee.ch/name/claimId.

hackrush01:
@kauffj, I wanted to clarify point 1. So if a user provides both name and claim id or name and txid + nout then it should check whether the name returned by claim_show matches the name provided by the user, and if not then display some sort of error?

kauffj:
Correct.

hackrush01:
@kauffj This can be closed now I guess.

Fixed in #730

PR #730, opened 2017-06-23
[WIP] Fix claim_show

kaykurokawa:
Initially was just fixing docstring bug where name was required in claim_show but found a bunch of other problems. This will be an API change. 

1. Removed argument 'name' from claim_show. This is redundant because API command resolve should handle name resolution. The purpose of this command is to check a claim if the only information you have about it is the outpoint, or the claim ID. If want to look up a claim by both name and claim_id you can use resolve (https://github.com/lbryio/lbry/issues/644)

2. The command would return different outputs if the claim didn't exist depending on if you fed it a txid/nout or Claim ID. Make them consistent and return a {'error':'reason...'} dictionary like you would get in API command resolve.


lyoshenka:
@kaykurokawa is this still work in progress?

kaykurokawa:
yea need to test it still

kaykurokawa:
Done with this.

made some tests for claim_show for lbry in a box which can be merged when this PR is merged
https://github.com/lbryio/lbry-in-a-box/pull/9

kaykurokawa:
Made some adjustments with https://github.com/lbryio/lbryum/pull/124/commits/267230ddc2da3fd7777d8bf1460c0330b74eb9a6  in https://github.com/lbryio/lbryum/pull/124 

We now have an UnknownOutpoint exception in core/Error.py which we raise in _handle_claim_result()

PR #731, opened 2017-06-26
module cleanup

jackrobison:
- Added convenient import for setting up a daemon client, `from lbrynet.daemon import get_client`
- Renamed `lbrynet.lbryfile` to `lbrynet.lbry_file`
- Renamed `lbrynet.lbryfilemanager` to `lbrynet.file_manager`
- Renamed `lbrynet.lbrynet_daemon` to `lbrynet.daemon`
- Initialize lbrynet settings when configuring an api client if they are not set yet


kaykurokawa:
what's point of d9ced3c "convenient import for api client?" 

otherwise untested but LGTM

jackrobison:
@kaykurokawa I was starting to write up a brief article on using the daemon from python

Before, you'd set up a client like this:

```
from lbrynet import conf
from lbrynet.lbrynet_daemon.auth.client import LBRYAPIClient

conf.initialize_settings()
client = LBRYAPIClient.get_client()
```

The above was too ugly, IMO. That commit lets you do the same with:

```
from lbrynet.daemon import get_client

client = get_client()
```

kaykurokawa:
gotcha, LGTM

PR #732, opened 2017-06-26
Add support for downloading HTTP files

stephenrblake:
I was a little unsure of the new structure of the metadata (and where would be the best place to insert the code), so feel free to comment. I assumed the sd_hash field (the "source") would contain a URL if the sourceType was "http", but I could be wrong.

If all goes well, my email is stephenrblake@gmail.com and I will send a request for the bounty.


stephenrblake:
(Ignore the files not on the "master" branch for now.)

lyoshenka:
@stephenrblake this is great, thanks for helping out. I'll ask our protocol expert to look at this. I think these changes will also require a change to the metadata schema, most likely here: https://github.com/lbryio/lbryschema/blob/master/lbryschema/proto/source.proto

PR #733, opened 2017-06-26
Add docstring and unit test for CryptBlob.py

kaykurokawa:
Added unit tests and docstrings for CryptBlob.py,  the class responsible for encrypting and decrypting blobs.

No functionality changes.

### DIFF
--- a/tests/unit/cryptstream/test_cryptblob.py
+++ b/tests/unit/cryptstream/test_cryptblob.py
@@ -1,0 +1,79 @@
+from twisted.trial import unittest
+from twisted.internet import defer
+from lbrynet.cryptstream import CryptBlob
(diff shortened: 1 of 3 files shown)


tests/unit/cryptstream/test_cryptblob.py
```diff
@@ -0,0 +1,79 @@
...
+
+from Crypto import Random
+from Crypto.Cipher import AES
+import tempfile
+import random
+import string
+import StringIO
+import time
+
+class MocBlob(object):
```
lyoshenka: you missed a letter

kaykurokawa: is intentional other tests use the same naming scheme.

PR #736, opened 2017-06-27
Increased max_key_fee

### LINKED ISSUE #721: Re-think max_key_fee
There's a `max_key_fee` config in the daemon that prevents you from downloading anything over that fee. It's set to 25 USD by default. I think this needs to be changed somehow.

Option 1: remove it completely. the app already confirms before spending

Option 2: remove the default, or set it very high. Then if you want to actually enforce a limit, you can set it yourself.

Option 3: leave it as is

I'd prefer option 2.

Related app issue: https://github.com/lbryio/lbry-app/issues/268

hackrush01:
The default max_key_fee value is increased to US$ 50.
Fixes #721

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ at anytime.
   * Fix shutdown of the blob tracker by Session
   * Fixed claim_new_support docstrings
   * Fixed BlobManager causing functional tests to fail, removed its unneeded manage() loop
+  * Increased max_key_fee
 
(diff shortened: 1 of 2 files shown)


jackrobison:
LGTM

hackrush01:
LGTM means?

jackrobison:
@hackrush01 Looks good to me

hackrush01:
Oh OK Thanks!

PR #741, opened 2017-06-29
WIP: Extend protocol support

zestyr:
Goes with [this PR](https://github.com/lbryio/lbryschema/pull/9).

This change adds the ability to publish claims with the source types lbry_sd_hash, btih and http. If no source type is specified, regex functions are used to determine which type it is.

With some assistance from @jackrobison, I've setup a regtest network to do a bunch of tests, and everything seems to work fine. It's possible to publish any of the three source types, they're detected properly, you can omit various properties from the source dictionary except the source itself, publishing from the LBRY App still works, and the claims can be resolved. 

It's not yet possible to download the extra source types, running "get" results in "Failed to get ...", but I'm working on that now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ at anytime.
   * Missing docstring for `blob_list`
   * Added convenient import for setting up a daemon client, `from lbrynet.daemon import get_client`
   * Added unit tests for CryptBlob.py
+  * Add support for publishing source types `btih` and `http`
+  *
 
 ### Changed
   * Change `max_key_fee` setting to be a dictionary with values for `currency` and `amount`
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -594,8 +594,9 @@ Options:
     --license_url=<license_url>    : publication license url
     --thumbnail=<thumbnail>        : thumbnail url
     --preview=<preview>            : preview url
-    --nsfw=<nsfw>                  : title of the publication
-    --sources=<sources>            : {'lbry_sd_hash':sd_hash} specifies sd hash of file
+    --nsfw=<nsfw>                  : mark as nsfw (not safe for work)
+    --sources=<sources>            : {'source': source} specifies source of file. Source
+                                     can be a lbry_sd_hash, http url, or a btih hash.
     --channel_name=<channel_name>  : name of the publisher channel name in the wallet
     --channel_id=<channel_id>      : claim id of the publisher channel, does not check
                                      for channel claim being in the wallet. This allows
@@ -956,4 +957,3 @@ Usage:
 Returns:
     (str) Unused wallet address in base58
 ```
-
--- a/docs/index.md
+++ b/docs/index.md
@@ -449,7 +449,8 @@ Args:
     'thumbnail'(optional): (str) thumbnail URL for the file
     'preview'(optional): (str) preview URL for the file
     'nsfw'(optional): (bool) True if not safe for work
-    'sources'(optional): (dict){'lbry_sd_hash':sd_hash} specifies sd hash of file
+    'sources'(optional): (dict) {'source': source} specifies source of file. Source
+                          can be a lbry_sd_hash, http url, or a btih hash.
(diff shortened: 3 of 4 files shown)


lbrynet/daemon/Publisher.py
```diff
@@ -54,6 +55,42 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
...
+
+        source = claim_dict['stream']['source'].get('source', None)
+        source_type = claim_dict['stream']['source'].get('sourceType', None)
+        content_type = claim_dict['stream']['source'].get('contentType', 'application/octet-stream')
+        version = claim_dict['stream']['source'].get('version', '_0_0_1')
+
+        if source is None:
+            raise Exception("Cannot publish a stream with no source")
+
+        if source_type is None:
```
jackrobison: These checks should go in lbryschema, an dictionary containing invalid values passed to `ClaimDict.load_dict` should throw a `DecodeError` or an error specific to the field (like `UnknownSourceType`)

lbrynet/daemon/Publisher.py
```diff
@@ -54,6 +55,42 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
     @defer.inlineCallbacks
     def publish_stream(self, name, bid, claim_dict, claim_address=None, change_address=None):
         """Make a claim without creating a lbry file"""
+        log.info('Starting publish for %s', name)
+
+        source = claim_dict['stream']['source'].get('source', None)
+        source_type = claim_dict['stream']['source'].get('sourceType', None)
```
jackrobison: I think it's worth making a helper `ClaimDict.source_type` for this, similar to https://github.com/lbryio/lbryschema/blob/acb05b52f13bd0028c2b2d4d8939fb313c83c079/lbryschema/claim.py#L67

lbrynet/daemon/Publisher.py
```diff
@@ -54,6 +55,42 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
...
+        content_type = claim_dict['stream']['source'].get('contentType', 'application/octet-stream')
```
jackrobison: This should use `get_content_type(file_path)` to pick the `contentType`

lbrynet/daemon/Publisher.py
```diff
@@ -54,6 +55,42 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
...
+            btih = re.findall(btih_re, source.lower())
+
+            if len(lbry_sd_hash) == 1:
+                source_type = "lbry_sd_hash"
+            elif len(btih) == 1:
+                source_type = "btih"
+            elif len(http) == 1:
+                source_type = "http"
+            else:
+                raise Exception("Unknown source type")
```
jackrobison: As indicated in the above comment, this stuff should go in lbryschema. When you move it there the exception you're looking for is https://github.com/lbryio/lbryschema/blob/60ab79474b4e70e13b8b302f91fdfbcd4772f4ab/lbryschema/error.py#L1

lbrynet/daemon/Publisher.py
jackrobison: I think you should also rename the `ClaimDict.source_hash` property to `ClaimDict.source_value`, since with these changes it could be a url and not a BTIH or a SD hash

lbrynet/daemon/Publisher.py
zestyr: There is no file_path though, that's the prerequisite for this function running in the first place

lyoshenka:
This feels like its close to done. The issues raised in review need to be addressed, and it needs to be rebased.

eukreign:
@jackrobison @lyoshenka is this PR out of date and not mergeable? Can we close it? Or is there still something salvageable here we can fix and merge?

eukreign:
Closing this as significantly out-of-date and it appears the original PR author has abandoned the effort.

Issue #692, opened 2017-06-14
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcc in position 62: ordinal not in range(128)


lyoshenka:
@jackrobison happens reliably when downloading `lbry://casually-successful`

```
017-06-14 08:51:55,726 INFO     lbrynet.lbrynet_daemon.Downloader:101: Download lbry://casually-successful status changed to initializing
2017-06-14 08:51:55,728 INFO     lbrynet.lbrynet_daemon.Downloader:101: Download lbry://casually-successful status changed to downloading_metadata
2017-06-14 08:51:55,728 INFO     lbrynet.core.client.ConnectionManager:38: Connection Manager 8d20e99a127c90c7b67641fefbe23609346058ab55ed878c4f3b66555c120ba65dcfdaf648410d171be6d027d53bc4b1 initialized
2017-06-14 08:51:55,754 INFO     lbrynet.lbryfile.EncryptedFileMetadataManager:246: Saved sd blob hash 8d20e99a127c90c7b67641fefbe23609346058ab55ed878c4f3b66555c120ba65dcfdaf648410d171be6d027d53bc4b1 to stream hash 9dba62280f52ecb18c8c675e9128b8e0b3ebe9e68e8357e8bbeb3f85c9ba5867802dab08efe70818f048035082ebba10
2017-06-14 08:51:55,765 INFO     lbrynet.lbrynet_daemon.Downloader:101: Download lbry://casually-successful status changed to running
Traceback (most recent call last):
  File "logging/handlers.py", line 76, in emit
  File "logging/handlers.py", line 156, in shouldRollover
  File "logging/__init__.py", line 734, in format
  File "logging/__init__.py", line 465, in format
  File "logging/__init__.py", line 329, in getMessage
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcc in position 62: ordinal not in range(128)
```

jackrobison:
The serialized claim value: 
[...]

PR #746, opened 2017-06-30
Make sure window paths of directories are in string not unicode

kaykurokawa:
Likely fix for 

https://github.com/lbryio/lbry/issues/692 
https://github.com/lbryio/lbry/issues/661

data directories in windows were being returned as unicode type, which will result in problems when using os.path.join with filenames that are of string type. 

Added unit tests in test_conf.py to make sure the data directories are in string not unicode. 

relevant discussion where I got the solution from :
https://github.com/pypa/pip/issues/3463
the resulting pr where I got the solution from:
https://github.com/pypa/pip/pull/4000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ at anytime.
   * Fixed BlobManager causing functional tests to fail, removed its unneeded manage() loop
   * Increased max_key_fee
   * Fixed unit tests on appveyor Windows build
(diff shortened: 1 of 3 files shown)


kaykurokawa:
Depends on https://github.com/lbryio/lbry/pull/751 for unit test to work

kaykurokawa:
@tzarecbczan is reporting that this fixes #692 for him

lyoshenka:
lgtm

PR #747, opened 2017-06-30
Remove point reservation (payment queueing)

kaykurokawa:
Remove payment queuing/point reservation feature in the wallet and send transactions immediately. This removes ability for clients to send data fee payments. However, there is no changes that will prevent servers from asking for data fee payments (non generous hosting). 

a) sending is no longer queued , API command send_amount_to_address() will immediately send 

d) The Wallet class's manage() function has been reworked to use LoopingCall . The only thing it does now is check expected balances (from recieving data fees).  Before it queued payments and updated the balance from lbryum which it no longer needs to do. Calls to get the wallet balance is a direct query now to lbryum.

e) Checks for insufficients funds that the Daemon and Wallet used to make has been deleted. These checks no longer need to exist in lbry daemon because lbryum can do it now since there is no queued payments that it doesn't know about. Fixes https://github.com/lbryio/lbry/issues/748

c) Clients will no longer attempt to send data fee payments in BlobRequester. If data fee is more than 0 , it will log a warning and continue to attempt to download without queuing data fee payments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ at anytime.
   *
 
 ### Removed
-  *
+  * Removed payment queueing and therefore ability for clients to make data payments
   *
 
 ## [0.14.1] - 2017-07-07
--- a/lbrynet/core/PTCWallet.py
+++ b/lbrynet/core/PTCWallet.py
@@ -12,7 +12,6 @@ from lbrynet.pointtraderclient import pointtraderclient
 from twisted.internet import defer, threads
 from zope.interface import implements
 from twisted.python.failure import Failure
-from lbrynet.core.Wallet import ReservedPoints
 
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


lyoshenka:
does point c) above mean that no data fees are paid at all? so if someone  asks for data fees, we just don't pay them? or are they being paid elsewhere?

lyoshenka:
looks like the answer is yes. i know we talked about how this fixes more things than it breaks, but it still feels crappy to do this.

can you write up what's required to make data payment work for real? at least that way we'll know how much more work we need to do there

kauffj:
Agreed with @lyoshenka above. This seems like a step backwards.

kaykurokawa:
Well the premise is that data payment works as it is before this PR, and I don't really think that it was working in the first place. 

a) Needs test for data payment, as of now there are zero tests in unit tests or lbry-in-a-box integration tests. Which means, that basically, it doesn't work. 

b) Lbryum needs a way to reserve credits for future use (point reservation) and to clear them out periodically. The code will be similar to what was removed here.  

c) If you don't pay the data fee, nothing happens anyways right now. There is some peer penalty code but its unfinished, never really been paid any attention to.

d) Similar to c, If you are receiving data fees, there is not really any to audit how much you received, or figure out  who sent what.

eukreign:
Can this be closed? It's probably too out of date to ever be merged?

tzarebczan:
I assume the new wallet doesn't do any queuing. I think it was enabled to support data blob payments initially. Something we'll need to revisit in the future.

Issue #710, opened 2017-06-19
Daemon start fails if port is already bound 


tzarebczan:
## The Issue
If there are other services running on the typical daemon port (and possibly on 3333/4444), the daemon will fail to load. 
### Steps to reproduce
1. Run claymore miner
2. Run lbry. 
3. Receieve error

### Expected behaviour
Port should be incrementally revised until a free one is found. User should be made aware of any non default ports. (I.e if it's the daemon port, the api call will be different, if it's the sharing port, port forwarding may need to be adjusted) 

### Actual behaviour
Daemon fails. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
Here is a log of the issue occurring: https://github.com/lbryio/lbry/files/1103576/lbrynet.3.log.txt

kaykurokawa:
Ports to bind to should never automatically change to some other random port. This is poor practice because people need to write firewall rules which requires fixed port numbers. It also makes testing the service on the port impossible because it could be bound to some random port that the client has no knowledge of. 

It definitely should give proper feedback if it failed to bind to a port, and probably include instructions on how to change the port through the config file.

tzarebczan:
@hackrush01 It should alert the user that the port is already bound and explain how to change the port.
 
@kaykurokawa  / @jackrobison How would this port change be done with the app?

kauffj:
It doesn't even need to explain how to change the port at this point. I suppose it could spit out a link to a FAQ page or similar.

At this time, I'm not concerned with getting port changing inside of the app. All I'd like to see for now is a clean exit and user feedback alerting about the issue.

lyoshenka:
Fixed by https://github.com/lbryio/lbry/pull/749

PR #749, opened 2017-07-01
Daemon connects even if port 3333 is bound.

hackrush01:
Daemon incrementally searches for open ports until it is found(queries
10 ports ahead of peer_port).
Then informs the user of the port that is used.

Fixex #710


kaykurokawa:
Don't think we can merge this. My response on  #710:

Ports to bind to should never automatically change to some other random port. This is poor practice because people need to write firewall rules which requires fixed port numbers. It also makes testing the service on the port impossible because it could be bound to some random port that the client has no knowledge of.

It definitely should give proper feedback if it failed to bind to a port, and probably include instructions on how to change the port through the config file.

hackrush01:
That could also be taken care of. Should I instead print out instructions do change the port?

hackrush01:
@kauffj @zestyr @jackrobison 
This PR builds on the PR by @zestyr and adds support for downloading from http and bittorent sources. It has some callbacks that need fixing, otherwise it is working. The lbryschema has to be updated and libtorrent has to be installed inside the python virtual env which is bit of a tough nut to crack.

lyoshenka:
This feels like two separate PRs. Can you break this up please?

hackrush01:
Done.

kauffj:
@hackrush01 rather than change the port, please update this to fail to start and ensure that the info returned by `status` indicates the reason why.

lyoshenka:
This is good, but linking people to lbry.io/faq is not ideal. Can you write a faq article to link people to? See https://github.com/lbryio/lbry.io/blob/master/content/faq/how-to-stop-lbry.md as an example of a file that gets transformed into a faq article.

hackrush01:
Yes, I will do that in a short while.

PR #755, opened 2017-07-03
Fix default directories

zestyr:
#705 had a bug, so @kaykurokawa reverted it.
The functional change of this PR is the same. I've made an attempt at improving the big if-else statement, but imo it's kinda messy and quite a bit longer than the other version. 
Let me know if the other version is better, or if there's another way to structure this.
This also includes the two functions @kaykurokawa requested in the other PR.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -2,11 +2,17 @@ import base58
 import json
 import logging
 import os
+import re
 import sys
 import yaml
 import envparse
-from appdirs import user_data_dir
+from appdirs import user_data_dir, user_config_dir
 from lbrynet.core import utils
+
+try:
+    from lbrynet.winhelpers.knownpaths import get_path, FOLDERID, UserHandle
+except (ImportError, ValueError):
+    pass
 
 log = logging.getLogger(__name__)
 
@@ -43,6 +49,7 @@ settings_encoders = {
     '.yml': yaml.safe_dump
 }
 
+
 def _win_path_to_bytes(path):
     """
     Encode Windows paths to string. appdirs.user_data_dir()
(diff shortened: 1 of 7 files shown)


kaykurokawa:
This PR https://github.com/lbryio/lbry/pull/746 that got  merged in master changes somethings in conf.py  so please look over that and rebase this. 

The test is failing (Click on Details next to the red X's . If you want to run the tests on your local machine run "PYTHONPATH=. trial tests" from the base folder of this repo ) because the above mentioned PR adds a new test for download_directory which got renamed to download_dir here.

zestyr:
I've already included the PR, but I forgot to fix the test. All tests pass now.

lbrynet/conf.py
```diff
@@ -58,29 +65,72 @@ def _win_path_to_bytes(path):
...
-    default_data_dir = _win_path_to_bytes(default_data_dir)
-    default_lbryum_dir = _win_path_to_bytes(default_lbryum_dir)
+if 'darwin' in sys.platform:
+    platform = DARWIN
+    dirs = _get_old_directories(DARWIN)
+elif 'win' in sys.platform:
+    platform = WINDOWS
+    if os.path.isdir(user_data_dir('lbryum', roaming=True)) and \
+       os.path.isdir(user_data_dir('lbrynet', roaming=True)):
+        dirs = _get_old_directories(WINDOWS)
```
kaykurokawa: use: 

"_get_old_directories(WINDOWS)['data']" etc.. instead  to get the old directories  . I think it probably should be a 'or' statement instead of 'and' (if either one of the old folder exists, use old folder locations)

lbrynet/conf.py
```diff
@@ -58,29 +65,72 @@ def _win_path_to_bytes(path):
...
     platform = LINUX
-    default_download_directory = os.path.join(os.path.expanduser('~'), 'Downloads')
-    default_data_dir = os.path.join(os.path.expanduser('~'), '.lbrynet')
-    default_lbryum_dir = os.path.join(os.path.expanduser('~'), '.lbryum')
+    if os.path.isdir(os.path.expanduser('~/.lbrynet')) and \
+       os.path.isdir(os.path.expanduser('~/.lbryum')):
+        dirs = _get_old_directories(LINUX)
+    else:
+        dirs = _get_new_directories(LINUX)
+
```
kaykurokawa: same thing here, use _get_old_directroies(LINUX)['data'] etc... and use 'or' instead of 'and'

lbrynet/conf.py
```diff
@@ -58,29 +65,72 @@ def _win_path_to_bytes(path):
...
+    elif platform == DARWIN:
+        dirs['data'] = user_data_dir('LBRY')
+        dirs['lbryum'] = os.path.expanduser('~/.lbryum')
+        dirs['download'] = os.path.expanduser('~/Downloads')
+    elif platform == LINUX:
+        dirs['data'] = os.path.expanduser('~/.lbrynet')
+        dirs['lbryum'] = os.path.expanduser('~/.lbryum')
+        dirs['download'] = os.path.expanduser('~/Downloads')
+    return dirs
+
```
kaykurokawa: add

else:
   raise ValueError('unknown platform value') 

incase platform is set to some unknown value

lbrynet/conf.py
```diff
@@ -58,29 +65,72 @@ def _win_path_to_bytes(path):
...
+            with open(os.path.join(user_config_dir(), 'user-dirs.dirs'), 'r') as xdg:
+                down_dir = re.search(r'XDG_DOWNLOAD_DIR=(.+)', xdg.read()).group(1)
+                down_dir = re.sub('\$HOME', os.getenv('HOME'), down_dir)
+                dirs['download'] = re.sub('\"', '', down_dir)
+        except EnvironmentError:
+            dirs['download'] = os.getenv('XDG_DOWNLOAD_DIR')
+
+        if not dirs['download']:
+            dirs['download'] = os.path.expanduser('~/Downloads')
+    return dirs
```
kaykurokawa: same thing here, add else statement for safety

lbrynet/conf.py
zestyr: So instead of 
`dirs = _get_old_directories(WINDOWS)`
You want 
```
dirs['data'] = _get_old_directories(WINDOWS)['data']
dirs['lbryum'] = _get_old_directories(WINDOWS)['lbryum']
dirs['download'] = _get_old_directories(WINDOWS)['download']
```
Or am I misunderstanding?

kaykurokawa: no instead of rewriting : 

user_data_dir('lbryum', roaming=True) on line 115, you write _get_old_directories(WINDOWS)['lbryum'] . This makes it easier to understand and you don't repeat yourself.

zestyr: oh I see, that makes sense

kaykurokawa:
Detection for old directories in windows is failing. user_data_dir('lbrynet', roaming=True) resolves to ...roaming directory/lbrynet/lbrynet (for author lbrynet and app lbrynet) . Where previously it was resolving to just ...roaming directory/lbrynet/ , same thing for the lbryum folder. 

I think you need to actually reuse the exact previous code (https://github.com/lbryio/lbry/blob/master/lbrynet/conf.py#L70, https://github.com/lbryio/lbry/blob/master/lbrynet/conf.py#L81) that defined the directories for both Windows and Linux in _get_old_directories(). Even for Linux, you made some minor adjustments which on first look should be the same (and it is on Ubuntu 16.04 I used for testing), but maybe there could be some obscure ways in which it resolve to different things. 

In the future, if you don't have access to an OS to test on, please ask other people to test it, especially if the PR directly deals with OS specific code.

zestyr:
Huh, I guess my testing wasn't thorough enough. I've been postponing setting up virtual machines but I'll do it asap so I can do proper testing.

kaykurokawa:
This works and LGTM now

PR #756, opened 2017-07-03
Fix error when resolving an integer

### LINKED ISSUE #588: Error resolving name that's an integer (thanks, dynamic typing)
Run this: `lbrynet-cli resolve_name name=1`

Get this error:

```
ERROR: 'int' object has no attribute 'startswith'

Help for resolve_name method:
    
    Resolve stream info from a LBRY name
    
    Args:
        'name': (str) name to look up, do not include lbry:// prefix
    Returns:
        (dict) Metadata dictionary from name claim, None if the name is not
                resolvable
    
Here's the traceback for the error you encountered:
  File "/home/grin/lbry/lbrynet/lbrynet_daemon/auth/server.py", line 214, in render
    return self._render(request)
  File "/home/grin/lbry/lbrynet/lbrynet_daemon/auth/server.py", line 327, in _render
    d = defer.maybeDeferred(function, **args_dict)
  File "/home/grin/.virtualenvs/lbry/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/grin/.virtualenvs/lbry/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1445, in unwindGenerator
    return _inlineCallbacks(None, gen, Deferred())
--- <exception caught here> ---
  File "/home/grin/.virtualenvs/lbry/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1299, in _inlineCallbacks
    result = g.send(result)
  File "/home/grin/lbry/lbrynet/lbrynet_daemon/Daemon.py", line 1343, in jsonrpc_resolve_name
    metadata = yield self._resolve_name(name, force_refresh=force)
  File "/home/grin/lbry/lbrynet/lbrynet_daemon/Daemon.py", line 761, in _resolve_name
    if name.startswith('lbry://'):
exceptions.AttributeError: 'int' object has no attribute 'startswith'
```

zestyr:
Fixes #588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ at anytime.
   * Fixed incorrect formatting of "amount" fields
   * Fixed handling of SIGINT, SIGTERM.
   * Fixed shutdown sequence
+  * Fix error when resolving an integer
+  *
 
 ### Deprecated
   * The API will no longer be served at the /lbryapi path. It will now be at the root.
@@ -80,7 +82,7 @@ at anytime.
 ### Fixed
  * Fixed timeout behaviour when calling API command get
  * Fixed https://github.com/lbryio/lbry/issues/765
- 
+
 ### Removed
   * Removed stream_info_cache.json from daemon.py
 
--- a/lbrynet/daemon/DaemonCLI.py
+++ b/lbrynet/daemon/DaemonCLI.py
@@ -133,6 +133,8 @@ def main():
 def guess_type(x, key=None):
     if not isinstance(x, (unicode, str)):
         return x
+    if key in ('uri', 'channel_name', 'name', 'file_name', 'download_directory'):
+        return x
     if x in ('true', 'True', 'TRUE'):
         return True
     if x in ('false', 'False', 'FALSE'):
@@ -143,8 +145,6 @@ def guess_type(x, key=None):
         except ValueError:
             # not a float
             pass
-    if key == "uri":
-        return x
     try:
         return int(x)
     except ValueError:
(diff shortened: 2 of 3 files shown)


kaykurokawa:
I think this is not the right approach. 
We just can't litter type casting code everywhere where we expect a certain type. 

I think since this is strictly a CLI issue, the fix has to go here: 
https://github.com/lbryio/lbry/blob/master/lbrynet/daemon/DaemonCLI.py#L23

There is a guess_type() function that is applied to all CLI input arguments but it should probably detect that if the argument is for something called "uri" , then it does not apply it.

zestyr:
I agree, though I talked to Jack about how I'd done this and he said it was probably fine. 
I thought the issue included the JSON API usage, since an integer URI would cause the same issue there. @lyoshenka what do you think? I

lyoshenka:
I agree with Kay that it should happen in one place and that we should not cast to string all over. It's possible that there is no single place where this can happen right now, which is a sign that our code should be refactored to create such a place.

One general solution I see is to automatically validate/convert all args to all jsonrpc functions. If the docstring says that an arg should be a `str`, then it should be cast to `str` before being passed to the function.

@kaykurokawa is this too much to fix this issue? do you see a smaller but still general solution? id like to avoid a solution where all params named "uri" are cast, because that's too magicy.

kaykurokawa:
The issue does not include JSON RPC usage because in JSON you specify strings by wrapping them in quotes. If a user inputs for example the number 1 without quotes as a uri through JSON RPC, then they've used the API wrong. 

I still stand by my original suggestion for the fix.

lyoshenka:
@kaykurokawa sounds good to me

@zestyr if you implement kay's fix, does that solve everything?

lyoshenka:
@zestyr status update?

zestyr:
I'll look into it shortly

kaykurokawa:
I also made it so that it will also not parse uri arguments as floats, or booleans. And also added other arguments (name, file_name, channel_name, download_directory) that should only be parsed as string in the CLI. 

rebased to master. 

LGTM

zestyr:
Looks good, fixed a little inconsistency regarding the array

kaykurokawa:
lgtm

PR #757, opened 2017-07-04
Remove old float formatting

zestyr:
Goes with lbryio/lbryum#127 and lbryio/lbryschema#12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ at anytime.
   *
 
 ### Fixed
-  *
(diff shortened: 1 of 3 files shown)


jackrobison:
I need to look at this closer (and the lbryum pr) and test it out. Let's get it on the schedule for next week.

jackrobison:
Merged here: https://github.com/lbryio/lbry/commit/15e44b0c9d9e4ad6ce4983ba989a86a05c63062e

PR #761, opened 2017-07-06
Updated max_key_fee_setting

hackrush01:
Now setting max_key_fee to 0 will mean that max_key_fee is unlimited.


hackrush01:
Goes with this PR [lbry-app/#324](https://github.com/lbryio/lbry-app/pull/324)

kaykurokawa:
User might actually want to set it to zero though ( if he only wants free content).
Setting it to negative is an option. 
Personally I'd prefer something less hacky like a new setting variable like disable_max_key_fee_check.

kauffj:
@kaykurokawa it looks like @hackrush01 updated this to work as requested and use an additional disable flag. Is this mergeable?

lbrynet/daemon/Downloader.py
```diff
@@ -97,7 +98,7 @@ def check_fee_and_convert(self, fee):
                                                                            fee.amount)
         if converted_fee_amount > self.wallet.get_balance():
             raise InsufficientFundsError('Unable to pay the key fee of %s' % converted_fee_amount)
-        if converted_fee_amount > max_key_fee_amount:
+        if converted_fee_amount > max_key_fee_amount and not disable_max_key_fee:
```
kaykurokawa: missing self. here

kaykurokawa:
I will recommend that you write unit test for this function https://github.com/lbryio/lbry/blob/master/lbrynet/daemon/Downloader.py#L94 to make sure your max key fee and disable key fee settings are working. 

Here is where the unit tests for the Downloader.py file goes: 
https://github.com/lbryio/lbry/blob/master/tests/unit/lbrynet_daemon/test_Downloader.py

You can run the twisted trial tests for lbry by running PYTHONPATH=. trial tests in the base folder ( or to run specific tests you can run something like PYTHONPATH=. trial tests/unit/lbrynet_daemon/test_Something.py )

hackrush01:
Ok, I'll look into it.

hackrush01:
@kaykurokawa I DM'ed you on slack, regarding this issue.

kaykurokawa:
I tested the disable_max_key_fee setting and works as expected. 
Let's remove the new unit tests you attempted for now (below line 186 on test_Downloader.py)  and merge this.
 
And then we can open up a new PR to finish writing the tests which I can help you out on.

hackrush01:
Ok, yeah. That would be great, I'll just make the changes and push it.

hackrush01:
@kaykurokawa Please check.

kaykurokawa:
will be merged from here https://github.com/lbryio/lbry/pull/781 , thanks

PR #763, opened 2017-07-07
Write "failed to get version from Git" notice to stderr, not stdout

alexliebowitz:
This means the stdout output is always pure JSON which allows the app (and other tools) to parse it.

Needed for https://github.com/lbryio/lbry-app/pull/335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ at anytime.
   *
 
 ### Changed
-  *
+  * Make `lbrynet-daemon --version` reliably machine readable (writes warning to stderr instead of stdout)
   *
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


jackrobison:
Can we drop this function altogether?

I think it would be better to have the `lbrynet_version` always be what's in `lbrynet.__version__`, and check the bundled version against what the `version` api command returns instead of parsing stderr

Additionally, when non-fatal errors like that happen, I think it should really be a `log.error` or `log.warning` instead of printing to stdout or stderr.

lyoshenka:
@alexliebowitz i dont think this is the right solution, for several reasons:

1 - we already have a standard logging process. use that instead of `sys.stderr.write`
2 - the real fix is to detect when the `git` command doesnt exist, and return `lbrynet_version`. that way this function never fails

kauffj:
Wasn't outcome of discussion today that this can be dropped?

alexliebowitz:
Yep, it was part of the original approach of checking the existing version using `lbrynet-daemon --version` which we dropped. Closing.

PR #775, opened 2017-07-11
claim_show without name also works with CLI now.

hackrush01:
Improves upon 2nd point of #644.
However if the user is running the command without name argument, then he/she has to specify `--` before txid, nout or claim_id as the arguments without `--` are positional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ at anytime.
   *
 
 ### Fixed
-  *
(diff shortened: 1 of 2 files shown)


hackrush01:
It returns an error message if the given name and the server returned
names are not same.

Fixes #644

kaykurokawa:
We are actually removing the name argument unfortuantely

see here: https://github.com/lbryio/lbry/pull/730

PR #783, opened 2017-07-18
Wallet encryption

jackrobison:
* Add `wallet_unlock`, a command available during startup to unlock an encrypted wallet

  * Block wallet startup on successfully being unlocked


lbrynet/daemon/Daemon.py
```diff
@@ -1218,6 +1220,34 @@ def jsonrpc_wallet_balance(self, address=None, include_unconfirmed=False):
...
     @defer.inlineCallbacks
+    def jsonrpc_wallet_unlock(self, password):
+        """
+        Unlock an encrypted wallet
+
+        Usage:
+            wallet_unlock (<password>)
+
+        Returns:
+            True if wallet is unlocked, otherwise False
```
zestyr: For consistency's sake, `(bool) true if ...`

jackrobison: fixed this

jackrobison:
This will check the OS keychain for a lbryum password if the wallet is encrypted and the pw isn't provided. This means that if a user has an encrypted wallet and they have the password saved to their keychain, the daemon will start right up without blocking on pw entry. If the password isn't in the keychain, daemon startup gets blocked until it has been provided with the `wallet_unlock` api command.

There aren't yet any commands to update an existing wallet encryption pw, remove the pw, or add one to an unencrypted wallet (but the functionality is in lbryum).

lbrynet/core/Wallet.py
```diff
@@ -1168,6 +1172,32 @@ def check_started():
...
+
+        def unlock(password):
+            if self._cmd_runner and self._cmd_runner.locked:
+                try:
+                    self._cmd_runner.unlock_wallet(password)
+                except InvalidPassword:
+                    log.warning("Incorrect password")
+                    check_locked()
+                    raise InvalidPassword
+            if self._cmd_runner.locked:
```
lyoshenka: why is this a separate if statement from the previous if statement? what if `self._cmd_runner` is None?

lyoshenka: maybe you meant to tab this in one more level?

lbrynet/core/Wallet.py
```diff
@@ -1168,6 +1172,32 @@ def check_started():
...
+                raise Exception("Failed to unlock wallet")
+            self.wallet_unlocked_d.callback(True)
+            log.info("Unlocked the wallet!")
+
+        def check_locked():
+            if self._cmd_runner.locked:
```
lyoshenka: what if cmd_runner is none?

requirements.txt
```diff
@@ -14,7 +14,7 @@ jsonrpc==1.2
 jsonrpclib==0.1.7
 jsonschema==2.5.1
 git+https://github.com/lbryio/lbryschema.git@v0.0.9rc1#egg=lbryschema
-git+https://github.com/lbryio/lbryum.git@v3.1.4rc3#egg=lbryum
+git+https://github.com/lbryio/lbryum.git@wallet-encryption#egg=lbryum
```
lyoshenka: dont forget to revert before merge

lbrynet/core/Wallet.py
```diff
@@ -1455,6 +1492,24 @@ def _save_wallet(self, val=None):
         self.wallet.storage.write()
         return defer.succeed(val)
 
+    def decrypt_wallet(self):
+        if not self.wallet.use_encryption:
+            return True
```
lyoshenka: not sure i like this. `decrypt_wallet` shouldn't say it decrypted something when it didnt

lbrynet/daemon/Daemon.py
```diff
@@ -988,6 +991,10 @@ def jsonrpc_status(self, session_status=False, dht_status=False):
         local_height = self.session.wallet.network.get_local_height() if has_wallet else 0
         remote_height = self.session.wallet.network.get_server_height() if has_wallet else 0
         best_hash = (yield self.session.wallet.get_best_blockhash()) if has_wallet else None
+        if has_wallet and self.session.wallet.wallet:
+            wallet_is_encrypted = self.session.wallet.wallet.use_encryption
+        else:
+            wallet_is_encrypted = False
```
lyoshenka: these 4 lines can just be 

    wallet_is_encrypted = has_wallet and self.session.wallet.wallet and 
        self.session.wallet.wallet.use_encryption

lyoshenka: also, isnt `session.wallet.wallet` weird? why the double `wallet`?

lbrynet/daemon/Daemon.py
```diff
@@ -1199,6 +1207,67 @@ def jsonrpc_wallet_balance(self, address=None, include_unconfirmed=False):
...
+        Unlock an encrypted wallet
+
+        Usage:
+            wallet_unlock (<password>)
+
+        Returns:
+            (bool) true if wallet is unlocked, otherwise false
+        """
+
+        if not self.session.wallet._cmd_runner.locked:
```
lyoshenka: isn't it bad to access "private" members?

also, isn't `_cmd_runner` sometimes `None`?

lbrynet/tests/unit/core/test_Wallet.py
```diff
@@ -195,3 +208,59 @@ def update_balance():
             'test', "f43dc06256a69988bdbea09a58c80493ba15dcfa", 4))
         self.assertFailure(d, InsufficientFundsError)
         return d
+
```
kaykurokawa: These tests look like they belong in lbryum since no lbrynet objects are used here 
I think moving this to new file under lbryum/tests like test_encryption.py would be good.

lbrynet/conf.py
```diff
@@ -278,6 +278,7 @@ def _convert_value(value):
     'peer_search_timeout': (int, 3),
     'use_auth_http': (bool, False),
     'use_upnp': (bool, True),
+    'use_keyring': (bool, False),
```
kaykurokawa: This configuration does not propagate to the lbryum configuration

lbrynet/tests/unit/core/test_Wallet.py
akinwale: @kaykurokawa I have refactored the tests to make use of common setUp and tearDown code. I think there's a benefit of having some of these tests here to just to make sure that they work properly when called from the daemon. We can add more tests to the lbryum suites.

I think we should create additional issues to address creating improved accuracy tests in the lbrynet test suites as well as the lbryum test suites.

kaykurokawa:
I'll look this over by tomorrow

kaykurokawa:
Attempting to run wallet_encrypt with use_keyring = False results in this error:

    result = g.send(result)
  File "/usr/local/lib/python2.7/dist-packages/lbrynet-0.19.0rc15-py2.7.egg/lbrynet/daemon/Daemon.py", line 1307, in jsonrpc_wallet_encrypt
    self.session.wallet.encrypt_wallet(new_password)
  File "/usr/local/lib/python2.7/dist-packages/lbrynet-0.19.0rc15-py2.7.egg/lbrynet/core/Wallet.py", line 1577, in encrypt_wallet
    self._cmd_runner.update_password(new_password, update_keyring)
  File "/usr/local/lib/python2.7/dist-packages/lbryum-3.2.0rc4-py2.7.egg/lbryum/commands.py", line 179, in update_password
    raise ValueError("no keyring is available to update")
exceptions.ValueError: no keyring is available to update

I think this is a problem on the lbryum side in the command update_password() 
The argument update_keyring is by default True, but should probablly be adjusted based on what the conf is set to.

kaykurokawa:
Some more things: 

We need to adjust API command settings_set to be able change use_keyring conf so the app has a way of adjusting it. Which brings up some more issue:

Should changing the use_keyring conf set the keyring password? Seems like it should

If  user can turn keyring on and off, there is a potential that the wrong password gets in the keyring (user changes password when keyring is off, turns it back on again). If the keyring password is wrong, then lbrynet needs to catch this problem and allow the user to recover from this problem (needs to at least error out with an information error message).

akinwale:
@kaykurokawa The encrypt_wallet method in the lbrynet side also has an update_keyring parameter which is set to True by default. I guess it would make sense to change it here. See https://github.com/lbryio/lbry/pull/783/files#diff-eafee006192cca2ffbca5ba91dc83eeeR1572

jackrobison:
@akinwale we'll also need to make sure `keyring` is getting packed in the build

kaykurokawa:
The app will also likely need some way of determining whether it should prompt the user for a password or not (whether its prompting to add a password for a wallet that's unencrypted , or prompting for a password on an encrypted wallet without a keyring) 

Probably we need to add a new entry in the API command status that tells you whether a wallet is encrypted or not.

kauffj:
Having spent all of 30 seconds thinking about this, I don't think the app cares whether this is returned by `status` or `settings_get`

akinwale:
@kaykurokawa @kauffj wallet_is_encrypted is currently returned in the status call.

> We need to adjust API command settings_set to be able change use_keyring conf so the app has a way of adjusting it. Which brings up some more issue:
> 
> Should changing the use_keyring conf set the keyring password? Seems like it should
Do we add another parameter to settings_set in order to accept the password? I'm guessing that also means that we would need to verify the password parameter is always present if use_keyring is set to true. We would also need to accept an old password (if set) before letting the user set a new password. I think this require some more thought and maybe we should create a new issue for it?

kaykurokawa:
>@kaykurokawa @kauffj wallet_is_encrypted is currently returned for the status call.

ah, ok i missed that. 

>Do we add another parameter to settings_set in order to accept the password? I'm guessing that also means that we would need to verify the password parameter is always present if use_keyring is set to true. We would also need to accept an old password (if set) before letting the user set a new password. I think this require some more thought and maybe we should create a new issue for it?

Yes the issue is a bit complicated, because there's different cases here:

wallet is unencrypted, use_keyring is set to True : nothing happens in this case I guess, or maybe we don't allow this to happen? 
wallet is unencrypted, use_keyring is set to False: nothing happens in this case I guess
wallet is encrypted, use_keyring is set to True : password gets sent to the keyring 
wallet is encrypted, use_keyring is set to False : do we delete the password from keyring here? not sure 

I think adf0201 does not seem quite right to fix the exception I mentioned "exceptions.ValueError: no keyring is available to update". Now the behavior won't be right when you are using the keyring and you want to update the password, because update_keyring is now False. The root of the problem is here in lbryum: 

https://github.com/lbryio/lbryum/blob/master/lbryum/commands.py#L169

The function update_password(self, new_password, update_keyring=True) takes an update_keyring argument which is set to default True but the decision to update_keyring should be dependent on whether Command._keyring exists nor not ( if self._keyring is not None: update_keyring = True)

akinwale:
> I think adf0201 does not seem quite right to fix the exception I mentioned "exceptions.ValueError: no keyring is available to update". Now the behavior won't be right when you are using the keyring and you want to update the password, because update_keyring is now False. The root of the problem is here in lbryum:
> 
> https://github.com/lbryio/lbryum/blob/master/lbryum/commands.py#L169
> 
> The function update_password(self, new_password, update_keyring=True) takes an update_keyring argument which is set to default True but the decision to update_keyring should be dependent on whether Command._keyring exists nor not ( if self._keyring is not None: update_keyring = True)

The commit should fix the error that you encountered in your previous test (from the lbry side), since the `update_keyring` parameter in the lbry daemon previously defaulted to True, and the same parameter is passed on to the command runner's `update_password` method. I also tested this just to be sure it works.

The reason I raise the `ValueError` in lbryum is to handle cases where the parameter is set to `True`, but a valid keyring backend may not have been installed on the system (resulting in `self._keyring` being `None`). I believe it was expected for a keyring to always be present before the modifications to make it optional, and I think @jackrobison could provide some insight into what the original reason for having this parameter is.

PR #789, opened 2017-07-19
Added validation for currencies.

### LINKED ISSUE #778: When setting max_key_fee using API, currency is not getting validated.
Currently any currency can be set such as "INR", "EUR" etc and it will not be validated. However the expected behavior is the currency should be any of LBC, USD or BTC only.

hackrush01:
Now while setting currency via API/CLI it is checked that the user has
not entered any currency other that LBC, BTC or USD.

Fixes #778


kaykurokawa:
this LGTM  ,
added some tests and rebased to master here https://github.com/lbryio/lbry/pull/802

PR #801, opened 2017-07-24
Resolving a channel returns 'channel_claim_count'

hackrush01:
Since I couldn't get my lbryum server up for testing this, it is a dry
fix based on the review.

Goes with lbryio/lbryum#138


lbrynet/daemon/Daemon.py
```diff
@@ -2002,7 +2002,7 @@ def jsonrpc_claim_list(self, name):
     @defer.inlineCallbacks
     def jsonrpc_claim_list_by_channel(self, page=0, page_size=10, uri=None, uris=[]):
         """
-        Get paginated claims in a channel specified by a channel uri
+        Get claims in a channel specified by a channel uri
```
jackrobison: Since the results are still paginated, this shouldn't be changed

hackrush01: Had it in my local repo, forgot to push :-P

tzarebczan:
@hackrush01 can we get this fixed?

PR #806, opened 2017-07-26
[WIP] BlobManager database migration for future announce rework 

kaykurokawa:
This is a database migration of BlobManager's database blobs.db

We create two new columns should_announce, and last_announce_time. should_announce is for marking some sd hashes and head blobs (first blob in a stream) as True so they can be distinguished from blobs that shouldn't be announced. 

last_announce_time is for marking a blob's last announce time so they can be troubleshooted better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ at anytime.
   *
 
 ### Added
-  *
+  * Added additional columns in BlobManager database for future blob announcement change
   *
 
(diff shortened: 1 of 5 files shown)


lyoshenka:
@kaykurokawa if this needs review, please assign reviewers

kaykurokawa:
closing for https://github.com/lbryio/lbry/pull/824

PR #807, opened 2017-07-27
Remove TempBlobManager

kaykurokawa:
Remove the abomination that is TempBlobManager. Base BlobManager class is also removed since there is now only DiskBlobManager.  

The class is used only as a replacement for DiskBlobManager when running various functional tests. The problem is that we are not testing DiskBlobManager which is the actual class that gets used in the client and needs to be tested.  This change is necessary because I will be making various changes to DiskBlobManager for the blob announcement rework and I will need tests run on the actual class. 

To get the tests working with DiskBlobManager, we use tempfile.mkdtemp() anywhere we can to create temporary working directories that DiskBlobManager and other classes need.  I created mk_db_and_blob_dir() and rm_db_and_blob_dir() under tests.util to do this.


tests/util.py
```diff
@@ -2,14 +2,24 @@
...
 DEFAULT_ISO_TIME = time.mktime(DEFAULT_TIMESTAMP.timetuple())
 
 
+def mk_db_and_blob_dir():
+    db_dir = tempfile.mkdtemp()
+    blob_dir = tempfile.mkdtemp()
+    return db_dir, blob_dir
+
+def rm_db_and_blob_dir(db_dir, blob_dir):
+    shutil.rmtree(db_dir, ignore_errors=True)
```
jackrobison: Why ignore errors? What errors get ignored?

jackrobison: Does this work on windows?

jackrobison:
A couple questions, otherwise LGTM

tests/util.py
kaykurokawa: there is a race condition that gets triggered on windows , but it could probablly happen on other OS (BlobFile class creates a temporary file which it deletes, rmtree will try to delete that file but it will have already been deleted)  
https://bugs.python.org/issue29699

PR #814, opened 2017-07-31
[WIP] add blob_announce ability to specify lists of blobs

kaykurokawa:
need to announce multiple blobs at once on reflector since they don't keep files


kaykurokawa:
only for temporary use closing now

PR #818, opened 2017-08-02
add claim_send_tip

jackrobison:


kauffj:
I also propose that `send_amount_to_address` be renamed to `wallet_send` and that this be called `wallet_send_tip` with a signature of `wallet_send_tip(amount, claim_id)` to match `wallet_send`.

kauffj:
Other alternatives:
- `wallet_spend` and `wallet_spend_tip`
- just `wallet_tip`

hackrush01:
There are two `_tip_claim` functions in wallet. Is it intentional?

PR #819, opened 2017-08-02
Add option for clients to seek head blob first when downloading 

kaykurokawa:
This adds an option in conf for clients to download by prioritizing the seeking of the "head blob" (the very first blob in the stream) before they seek the next unavailable blob. 

It is turned off by default in this PR so will not change how downloading works unless configured to be turned on manually.  Should be turned on when reflector switch to announcing head blob only.

### DIFF
--- a/lbrynet/core/client/ConnectionManager.py
+++ b/lbrynet/core/client/ConnectionManager.py
@@ -25,10 +25,12 @@ class ConnectionManager(object):
 
     def __init__(self, downloader, rate_limiter,
                  primary_request_creators, secondary_request_creators):
+
         self.downloader = downloader
         self.rate_limiter = rate_limiter
         self._primary_request_creators = primary_request_creators
         self._secondary_request_creators = secondary_request_creators
+        self.seek_head_blob_first = conf.settings['seek_head_blob_first']
         self._peer_connections = {}  # {Peer: PeerConnectionHandler}
         self._connections_closing = {}  # {Peer: deferred (fired when the connection is closed)}
         self._next_manage_call = None
@@ -150,10 +152,7 @@ class ConnectionManager(object):
             log.debug("%s have %d connections, looking for %d",
                         self._get_log_name(), len(self._peer_connections),
                         conf.settings['max_connections_per_stream'])
-            ordered_request_creators = self._rank_request_creator_connections()
-            peers = yield self._get_new_peers(ordered_request_creators)
-            new_conns = conf.settings['max_connections_per_stream'] - len(self._peer_connections)
-            peers = self._pick_best_peers(peers, new_conns)
+            peers = yield self._get_new_peers()
             for peer in peers:
                 self._connect_to_peer(peer)
         self._manage_deferred.callback(None)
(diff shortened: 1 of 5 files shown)


tzarebczan:
In what cases should this option be turned on? How would it be turned on manually?

kaykurokawa:
"seek head blob first" download method should not be turned on until our reflector servers has switched to announcing head blob only.

It can be turned on and off through the configuration file (the default configuration is set to off)

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
-                if request_creator in p.request_creators])
-
-        return sorted(self._primary_request_creators, key=count_peers)
+    def return_shuffled_peers_not_connected_to(self, peers, new_conns_needed):
+        if peers is None:
+            # can happen if there is some error in the lookup
+            return []
+        out = [peer for peer in peers if peer not in self._peer_connections]
+        random.shuffle(out)
+        out = out[0:new_conns_needed]
```
jackrobison: This can just be `return out[:new_conns_needed]`

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
-        def count_peers(request_creator):
-            return len([
-                p for p in self._peer_connections.itervalues()
-                if request_creator in p.request_creators])
-
-        return sorted(self._primary_request_creators, key=count_peers)
+    def return_shuffled_peers_not_connected_to(self, peers, new_conns_needed):
+        if peers is None:
+            # can happen if there is some error in the lookup
+            return []
```
jackrobison: Should this be an exception? A warning?

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
-        if not new_peers:
-            new_peers = yield self._get_new_peers(request_creators[1:])
-        defer.returnValue(new_peers)
-
-    def _pick_best_peers(self, peers, num_peers_to_pick):
-        # TODO: Eventually rank them based on past performance/reputation. For now
-        # TODO: just pick the first to which we don't have an open connection
+
+        # find peers for the head blob if configured to do so
+        if self.seek_head_blob_first is True:
```
jackrobison: `if self.seek_head_blob_first:`

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
+        out = [peer for peer in peers if peer not in self._peer_connections]
+        random.shuffle(out)
+        out = out[0:new_conns_needed]
+        return out
 
     @defer.inlineCallbacks
-    def _get_new_peers(self, request_creators):
+    def _get_new_peers(self):
+        new_conns_needed = self.max_connections_per_stream - len(self._peer_connections)
+        if new_conns_needed < 1:
```
jackrobison: `if not new_conns_needed:`

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
+        # find peers for the head blob if configured to do so
+        if self.seek_head_blob_first is True:
+            peers = yield request_creator.get_new_peers_for_head_blob()
+            peers = self.return_shuffled_peers_not_connected_to(peers, new_conns_needed)
+        else:
+            peers = []
+
+        # we didn't find any new peers on the head blob,
+        # we have to look for the first unavailable blob
+        if len(peers) == 0:
```
jackrobison: `if not peers:`

lbrynet/core/client/ConnectionManager.py
```diff
@@ -146,57 +150,58 @@ def _send_secondary_requests(self, peer, protocol):
...
-        likes
-        """
-        def count_peers(request_creator):
-            return len([
-                p for p in self._peer_connections.itervalues()
-                if request_creator in p.request_creators])
-
-        return sorted(self._primary_request_creators, key=count_peers)
+    def return_shuffled_peers_not_connected_to(self, peers, new_conns_needed):
+        if peers is None:
```
jackrobison: Should this be `if not peers:`? When is `peers` None?

lbrynet/core/client/ConnectionManager.py
lyoshenka: @jackrobison what if there's a bug and `new_conns_needed` ends up being -1 somehow? :smiling_imp:

lbrynet/core/client/ConnectionManager.py
kaykurokawa: this right here: 
https://github.com/lbryio/lbry/blob/master/lbrynet/core/client/BlobRequester.py#L118
It looks like you could get that when there is no more blobs for it to download... not sure.  I'll look through the code again to see if we can improve this part.

kaykurokawa:
rebased to master
made pep8 fixes as requested by jack 
Will need to check  BlobRequest._find_peers_for_hash() returning None

kaykurokawa:
Corrected and clarified in docstrings the returning of None when looking for peers. 

BlobRequestr._get_hash_for_peer_search() , function to get the next unavailable hash, could return None if there was nothing left to download. Previously, the None would propagate and functions would have its own way of handling None. Now we deal with None immediately in BlobRequester. get_new_peers_for_next_unavailable() and removed all other handling of None in other places.

lyoshenka:
@kaykurokawa whats the next step on this?

kaykurokawa:
@jackrobison please take a look at 734c5a1

PR #820, opened 2017-08-02
Daemon fixes

lyoshenka:
This does several things (bad practice, I know :cry: )

- fix handling of SIGTERM/SIGINT so we no longer get the "cant stop non-running reactor" bug
- fix daemon shutdown. It always shuts down by using `reactor.fireSystemEvent("shutdown")`
- dropped some old API endpoints: `/view`, `/upload`
- API is now accessible at `localhost:5279` in addition to the old `localhost:5279/lbryapi`. The latter is now deprecated, and will be dropped eventually.
- Fixed logging messages to say what class is shutting down, rather than printing the class object (which is meaningless)
- Pyinstaller now uses the same console_scripts that setup.py uses. It's not 100% unified, but we don't need a separate script for pyinstaller packaging.


lbrynet/daemon/DaemonServer.py
```diff
@@ -54,10 +54,9 @@ def start(self, use_auth):
 
     @defer.inlineCallbacks
     def stop(self):
-        if self._api is not None:
-            yield self._api._shutdown()
-        if self.server_port is not None:
```
jackrobison: I think you should keep `server_port.stopListening()`, maybe in `_shutdown` instead

lbrynet/daemon/Daemon.py
```diff
@@ -385,7 +386,15 @@ def _clean_up_temp_files(self):
             except OSError:
                 pass
 
+    @staticmethod
+    def _already_shutting_down(sig_num, frame):
+        log.info("Already shutting down")
+
     def _shutdown(self):
+        # ignore INT/TERM signals once shutdown has started
```
jackrobison: Why were these were necessary? Is `_shutdown` being called more than once?

lyoshenka: If you press Ctrl-C while daemon is shutting down, you'll get an error message where it tries to shut down the reactor twice. This prevents that and lets you know the daemon's working on it.

lbrynet/daemon/DaemonServer.py
lyoshenka: This happens in [Daemon._stop_server()](https://github.com/lbryio/lbry/blob/b464853ede24cb03043b921c58dbef1e4d9a3192/lbrynet/daemon/Daemon.py#L351) which is called in [Daemon._shutdown()](https://github.com/lbryio/lbry/blob/b464853ede24cb03043b921c58dbef1e4d9a3192/lbrynet/daemon/Daemon.py#L406)

lyoshenka:
Merged in 350386ba

PR #824, opened 2017-08-04
Add option for seeders to announce head blob only

kaykurokawa:
This PR adds the option to announce by head blob only. Head blob is the first blob in a stream. 

We create two new columns in the blob manager database: should_announce, and last_announce_time. should_announce is for marking some sd hashes and head blobs (first blob in a stream) as True so they can be distinguished from blobs that shouldn't be announced. last_announce_time is for marking a blob's last announce time so they can be troubleshooted better, this column's usage has not been implemented and is left as null  for future work however. 

Anything that creates blobs will be responsible for marking them as should_announce = True / False depending on if its a head blob, or sd blob (this is partially why commit bdccc35 removes the responsibility of HashBlobCreator to know about BlobManager. Since HashBlobCreator doesn't know if it is a sd blob or a head blob, but the classes creating and using HashBlobCreator does, we want those classes to mark blobs as should_announce=True/False through the BlobManager) 

We add a new configuration in conf "announce_head_blob_only" which when True, means that only the head blob and sd blob will be announced. When False, all blobs will be announced. This PR sets it to False by default so this PR will not change the announcement method. 

This PR will not affect how reflector announces blobs, since we are replacing reflector with the prism module and we need to change the prism module to support the annoucement of head blob only.

### DIFF
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -142,7 +142,6 @@ class DiskBlobManager(DHTHashSupplier):
                                 "    blob_length integer, " +
                                 "    last_verified_time real, " +
                                 "    next_announce_time real, " +
-                                "    last_announce_time real, " +
                                 "    should_announce integer)")
 
 
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -71,7 +71,7 @@ class DiskBlobManager(DHTHashSupplier):
         raise Exception("Hash announcer not set")
 
     @defer.inlineCallbacks
-    def blob_completed(self, blob, next_announce_time=None, should_announce=False):
+    def blob_completed(self, blob, next_announce_time=None, should_announce=True):
         if next_announce_time is None:
             next_announce_time = self.get_next_announce_time()
         yield self._add_completed_blob(blob.blob_hash, blob.length,
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -128,7 +128,10 @@ class DiskBlobManager(DHTHashSupplier):
                                 "    blob_hash text primary key, " +
                                 "    blob_length integer, " +
                                 "    last_verified_time real, " +
-                                "    next_announce_time real)")
+                                "    next_announce_time real, " +
+                                "    last_announce_time real, " +
+                                "    should_announce integer)")
+
 
             transaction.execute("create table if not exists download (" +
                                 "    id integer primary key autoincrement, " +
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -47,7 +47,7 @@ class DiskBlobManager(DHTHashSupplier):
         return self._make_new_blob(blob_hash, length)
 
     def get_blob_creator(self):
-        return self.blob_creator_type(self, self.blob_dir)
+        return self.blob_creator_type(self.blob_dir)
 
     def _make_new_blob(self, blob_hash, length=None):
         log.debug('Making a new blob for %s', blob_hash)
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -5,6 +5,7 @@ import sqlite3
 
 from twisted.internet import threads, defer, reactor
 from twisted.enterprise import adbapi
+from lbrynet import conf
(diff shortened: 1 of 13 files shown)


lbrynet/core/BlobManager.py
```diff
@@ -59,29 +69,32 @@ def _immediate_announce(self, blob_hashes):
...
             next_announce_time = self.get_next_announce_time()
-        d = self._add_completed_blob(blob.blob_hash, blob.length, next_announce_time)
-        d.addCallback(lambda _: self._immediate_announce([blob.blob_hash]))
-        return d
+        yield self._add_completed_blob(blob.blob_hash, blob.length,
+                                       next_announce_time, should_announce)
+        # we announce all blobs immediately, if announce_head_blob_only is False
+        # otherwise, announce only if marked as should_announce
+        if not self.announce_head_blobs_only or should_announce:
+            yield self._immediate_announce([blob.blob_hash])
```
jackrobison: https://github.com/lbryio/lbry/commit/09d336bd0ccd6fd049b38e4deb35c4284492a673

lbrynet/core/BlobManager.py
```diff
@@ -127,7 +140,10 @@ def create_tables(transaction):
                                 "    blob_hash text primary key, " +
                                 "    blob_length integer, " +
                                 "    last_verified_time real, " +
-                                "    next_announce_time real)")
+                                "    next_announce_time real, " +
+                                "    last_announce_time real, " +
+                                "    should_announce integer)")
```
jackrobison: this should be a bool

lbrynet/core/BlobManager.py
```diff
@@ -127,7 +140,10 @@ def create_tables(transaction):
                                 "    blob_hash text primary key, " +
                                 "    blob_length integer, " +
                                 "    last_verified_time real, " +
-                                "    next_announce_time real)")
+                                "    next_announce_time real, " +
+                                "    last_announce_time real, " +
```
lyoshenka: Do you have a specific use for `last_announce_time` in mind? If not, I'd prefer not creating it prematurely.

lyoshenka: You don't even set the value anywhere.

lbrynet/core/BlobManager.py
kaykurokawa: there's no bool in sqlite

lbrynet/core/BlobManager.py
kaykurokawa: I've removed this

kaykurokawa:
Rebased to master, made sure 09d336b was properly merged as requested by jack
Removed last_announce_time as requested by grin

As discussion with Jack: 

I said that we should create another flag for should_announce to indicate "unknown whether to announce or not", but I think we don't need to. I changed should_announce to be True by default. So blobs created without setting should_announce flag (this is done by reflector currently) will be announced when announce_head_blob_only is True. If announce_head_blob_only is False, we ignore the should_announce flag so everything is announced regardless of flag.

lyoshenka:
I think the right way to do this is instead of `should_announce`, we have a column like `blob_type`. Then it can have values `sd_blob`, `head_content_blob`, or `content_blob`. Then the code can decide what to do with the blob.

This way, if you change your announce settings, you don't have to reset all the blobs in the db.

kauffj:
Can we just store the ordinal position of the blob rather than enshrining head_blob?

I consider all additions/instances of head blob to be technical debt.

kaykurokawa:
>I think the right way to do this is instead of should_announce, we have a column like blob_type. Then it can have values sd_blob, head_content_blob, or content_blob. Then the code can decide what to do with the blob.

So BlobManager is the class that was changed in this PR and it only deals with blobs, and has no idea what a stream is. It is also the class responsible for announcing blobs, since up until now the blob announcement does not depend on stream information.   

EncryptedFileMetadataManager is the class which deals with streams and in here we have actually information regarding what position in a stream a blob is (0 means its a head blob) and what the sd_hash is.  So in here, we actually have all the information to decide what to announce. 

It would have been nice to move blob announcement into EncryptedFileMetadataManger, since it already can figure out what kind of blob a blob is, and we wouldn't need to add this new should_announce field. However, if we do that, than we lose the capability to announce Blobs that doesn't belong to a stream at all (or we need to create a field in the BlobManager that marks a blob as not belonging to a stream, but then we've just created another field so the tecnical debt created is equivalent).

Issue #817, opened 2017-08-02
Limit max number of simultaneous connections


lyoshenka:
As discussed in https://github.com/lbryio/lbry/issues/810, the daemon creates too many connections when downloading content. Often these connections are to the same server. I looked at speech at one point and it had about 468 connections to the same reflector server (not sure if those connections are downloads or uploads). This doesn't scale, and it can hit the max open file limit on some systems. 

Two thoughts:

1. Can the daemon use a single connection per server?
2. Can we limit how many simultaneous connections the daemon makes in total?

kauffj:
Is it possible this reflects a deeper bug? I think the majority of content accessed on spee.ch should either already be on spee.ch or was submitted directly to spee.ch. Is spee.ch really needing to download such significant amounts of content? Or is this from images submitted to spee.ch getting uploaded elsewhere?

kaykurokawa:
Upon daemon startup, it tries to reflect all the files it has so for each file, it will create a reflector connection for each one simultaneously, which is not good if there are a lot of files. 

It also tries to reflect every hour, and if it takes more than 1 hour to reflect everything you will start to accumulate a backlog of reflect commands. 

relevant file: https://github.com/lbryio/lbry/blob/master/lbrynet/file_manager/EncryptedFileManager.py

billbitt:
FYI after @lyoshenka raised the limit I got the 'too many files open' error again today. log and file list below
[21745.docx.txt](https://github.com/lbryio/lbry/files/1198644/21745.docx.txt)
[lbrynet-8-3-17.docx.txt](https://github.com/lbryio/lbry/files/1198643/lbrynet-8-3-17.docx.txt)

kaykurokawa:
Firstly I think there should be an option to turn the scheduled and startup re-reflecting completely off , but keep the initial reflecting on for publishes. I think this will be an easy fix to your problems.  

Secondly there should be a fix for how it auto reflects (limit number of files to reflect at a time).

kaykurokawa:
once this is merged https://github.com/lbryio/lbry/pull/826 , set 'auto_re_reflect' in your config file to False

alyssaoc:
@tzarebczan and @billbitt can you check how many open connections spee.ch still has? Is this an open/closed issue?

tzarebczan:
[...]

PR #826, opened 2017-08-04
Add ability to control re-reflects through config

kaykurokawa:
File manager will attempt to automatically re-reflect files on startup and every hour. 

Make this configurable through config file. "auto_re_reflect" will turn this feature on/off , "auto_re_reflect_interval" will control how often it attempts to re-reflect. Auto re-reflect is by default on, and interval is every hour (same as before)

Should solve some issues for speech https://github.com/lbryio/lbry/issues/817 by turning auto_re_reflect off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ at anytime.
   *
 
 ### Added
-  *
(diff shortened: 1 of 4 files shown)


jackrobison:
LGTM

Issue #795, opened 2017-07-22
Better Claim_abandon support, cannot abandon in a couple scenarios


tzarebczan:
## The Issue
Added support (claim_new_support via api) to someone else's claim twice and it looks like the claim_id is the same on both entries. 
    {
      "address": "bXJjCQxEedecAdrWarfzMbeuX7ZFdDJU4r",
      "amount": 1.0,
      "blocks_to_expiration": 243208,
      "category": "support",
      "claim_id": "b64155f2d44553871dfdb1785aa49d6d3972cd84",
      "confirmations": 19766,
      "expiration_height": 454392,
      "expired": false,
      "height": 191418,
      "is_pending": false,
      "is_spent": false,
      "name": "three",
      "nout": 0,
      "txid": "02c6d313e8cdcce96382ed4450205180ca63d7d424e734e72c3dcb44350133e0"
    },
    {
      "address": "bCrbM9X7shQhxsaxKHBk7xAFCPZMR8h3gd",
      "amount": 249.0,
      "blocks_to_expiration": 243208,
      "category": "support",
      "claim_id": "b64155f2d44553871dfdb1785aa49d6d3972cd84",
      "confirmations": 19766,
      "expiration_height": 454392,
      "expired": false,
      "height": 191418,
      "is_pending": false,
      "is_spent": false,
      "name": "three",
      "nout": 1,
      "txid": "67b078d4b19e90a078441bdf8191d1960126b45b3d66b07927b78200c008202e"
    },

Which causes claim_abandon to fail since it only takes a claim ID.  

The same scenario happens when you add support to your own claim. 

Additional claim_abandon support will be needed for tipping: https://github.com/lbryio/lbry-app/issues/355 

### Steps to reproduce
1. Add claim support to the same claim twice
2. Try to abandon claim
3. Fails - "more than one claim that matches"

### Expected behaviour
1) Abandon all claims with that claim id
or
2) Each support claim should have its own claimid
or
3) allow txid as part of args (optional), so you can abandon each

### Actual behaviour
cannot abandon

## System Configuration

 

- LBRY Daemon version: 0.14.2RC7
- LBRY App version: 0.13.0
- LBRY Installation ID:
- Operating system: Windows

kaykurokawa:
This is actually expected behavior when abandoning by claim_id if there is more than one match. 

[...]

PR #829, opened 2017-08-08
Expose abandon by txid/nout in Daemon

kaykurokawa:
Enable abandoning by txid/nout (feature already available in lbryum) , necessary for abandoning when you make supports for your own claim , since they share the same claim_id. Fixes https://github.com/lbryio/lbry/issues/795

Removed some unnecessary error handling for abandon in the Daemon. Exceptions are thrown in Wallet and will be propagated so the catching is unnecessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ at anytime.
   * Added `claim_send_tip`, a command to tip the owner of a claim via a support transaction
   * Added `reflector` keyword parameter to `file_reflect` command
   * Added configuration options for auto re-reflect
+  * Added option to abandon by txid/nout
 
 ### Fixed
   *
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -951,11 +951,12 @@ class Wallet(object):
         defer.returnValue(claim)
 
     @defer.inlineCallbacks
-    def abandon_claim(self, claim_id):
-        claim_out = yield self._abandon_claim(claim_id)
+    def abandon_claim(self, claim_id, txid, nout):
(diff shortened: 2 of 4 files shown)


jackrobison:
I think we should also break out the `return_address` parameter, [since by default it'll be a source of  address clutter](https://github.com/lbryio/lbryum/blob/6aba97e61a234ccbc588a5dbfdbc472ca84c6f8a/lbryum/commands.py#L1921)

Otherwise LGTM

kaykurokawa:
I have a test for it in lbry in a box, https://github.com/lbryio/lbry-in-a-box/pull/13 which can be merged.
I can do return_address in a seperate PR

okwme:
i've updated `lbrynet-daemon` but my `lbrynet-cli` still tells me `claim_abandon` usage is:
````
            claim_abandon (<claim_id> | --claim_id=<claim_id>)
````

**Comment by tzarebczan** [comment]
@okwme is your lbrynet-cli also updated to at least 0.15.2 (https://github.com/lbryio/lbry/releases/tag/v0.15.2) ?

I ran claim_abandon and I get: 
```
c:\Program Files (x86)\LBRY\resources\app\dist>lbrynet-cli claim_abandon
ERROR: Must specify claim_id, or txid and nout

Help for claim_abandon method:

        Abandon a name and reclaim credits from the claim

        Usage:
            claim_abandon [<claim_id> | --claim_id=<claim_id>]
                          [<txid> | --txid=<txid>] [<nout> | --nout=<nout>]

        Return:
            (dict) Dictionary containing result of the claim
            {
                txid : (str) txid of resulting transaction
                fee : (float) fee paid for the transaction
            }
```

PR #835, opened 2017-08-11
Fix improper use of floats

kaykurokawa:
The daemon, when processing JSON RPC calls, needs to treat float fields as decimal.Decimal. 

It also needs to be able to return decimal.Decimal's as JSON without losing precision.

We switch to use simplejson library instead of json as it naively supports dumps() for Decimal types.

### DIFF
--- a/lbrynet/daemon/auth/client.py
+++ b/lbrynet/daemon/auth/client.py
@@ -3,7 +3,7 @@ import logging
 import requests
 import os
 import base64
-import json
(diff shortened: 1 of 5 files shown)


kaykurokawa:
This requires lbryschema changes that doesn't exist here

eukreign:
We're going to switch to using dewies/satoshis in the API so this is no longer needed.

PR #858, opened 2017-08-18
Android platform detection and androidhelpers module

akinwale:
* Added Android platform detection.
* Added `androidhelpers` module for determining base paths. The methods are implemented in the Android python service.


lbrynet/androidhelpers/__init__.py
```diff
@@ -0,0 +1 @@
+import paths
```
jackrobison: I don't think this is necessary

jackrobison:
One small thing, otherwise LGTM

PR #865, opened 2017-08-21
add exclude_claimtrie to wallet_balance

jackrobison:


kauffj:
Isn't the claim tree excluded by default currently? If so, is there a reason to change that behavior?

lbrynet/daemon/Daemon.py
```diff
@@ -1188,26 +1189,38 @@ def jsonrpc_commands(self):
...
             -u         :  Include unconfirmed
+            -c         :  Include claimtrie transactions
 
         Returns:
             (float) amount of lbry credits in wallet
         """
+
         if address is None:
-            return self._render_response(float(self.session.wallet.get_balance()))
+            result = float(self.session.wallet.get_balance())
```
kaykurokawa: looks like include_unconfirmed and exclude_claimtrie does nothing here

tzarebczan:
@eukreign can we close this as it will be taken into account in the new wallet?

eukreign:
Yeah, there is a new command for this and it lets you specify number of confirmations, whether to include/exclude reserved UTXOs (TX created but not in mempool yet) and any TX that are claim related:

https://github.com/lbryio/lbry/blob/lbryum-refactor/lbrynet/daemon/Daemon.py#L3185-L3186

PR #867, opened 2017-08-22
Add is_address RPC method

alexliebowitz:
Queries lbryum to ask if an address is valid. Needed for this lbry-app PR: https://github.com/lbryio/lbry-app/pull/489

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -2128,6 +2128,22 @@ class Daemon(AuthJSONRPCServer):
         return d
 
     @AuthJSONRPCServer.auth_required
+    def jsonrpc_is_address(self, address):
+        """
+        Checks if the given string is a valid wallet address.
+
+        Usage:
+            is_address (<address> | --address=<address>)
+
+        Returns:
+            (bool) true, if address is valid
+        """
+
+        d = self.session.wallet.address_is_mine(address)
+        d.addCallback(lambda is_mine: self._render_response(is_mine))
+        return d
+
+    @AuthJSONRPCServer.auth_required
     def jsonrpc_wallet_is_address_mine(self, address):
         """
         Checks if an address is associated with the current wallet.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ at anytime.
   *
 
 ### Added
-  *
+  * is_address RPC method (checks if address is valid using lbryum)
   *
 
 ### Removed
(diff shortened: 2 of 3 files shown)


lbrynet/daemon/Daemon.py
```diff
@@ -2128,6 +2128,22 @@ def jsonrpc_transaction_show(self, txid):
...
+        Checks if the given string is a valid wallet address.
+
+        Usage:
+            is_address (<address> | --address=<address>)
+
+        Returns:
+            (bool) true, if address is valid
+        """
+
+        d = self.session.wallet.address_is_mine(address)
```
kaykurokawa: Don't you mean wallet.is_address() here?

kaykurokawa:
I would prefer that we implemented this on the app side, using something like bitcoinjs-lib 

https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/address.js#L9

our address headers (variable labeled as version in the code above) are 85 or 122 in decimal

kaykurokawa:
another library option here https://github.com/cryptocoinjs/coinstring

lyoshenka:
Agreed with Kay. There's no reason to query the daemon for this.

lbrynet/daemon/Daemon.py
alexliebowitz: Duh, thanks. I started by copy and pasting the code from jsonrpc_address_is_mine since it's almost the exactly the same thing, but I didn't notice that I failed to change that part because all of the test addresses that I used happened to be my own.

alexliebowitz:
Alright. Yeah, I was right on the border on this one. I feel like the big idea of lbrynet is that it's effectively a library that gives you just about everything you need to build a LBRY app (even things you could theoretically build yourself). On the other hand, this is so trivial that it seems overkill to have to query the demon just to validate an address.

alexliebowitz:
Alright. Yeah, I was right on the border on this one. I feel like the big idea of lbrynet is that it's effectively a library that gives you just about everything you need to build a LBRY app (even things you could theoretically build yourself). On the other hand, this is so trivial that it seems overkill to have to query the demon just to validate an address.

kauffj:
After discussing with @alexliebowitz it does sound like it's potentially non-trivial for every use-case to re-validate? If so, I'm inclined to agree that every application utilizing LBRY ought to not have to re-implement this.

If we don't want to expose a validation call, maybe the daemon cleanly return invalid address rejections?

lyoshenka:
clean error on invalid address makes sense

lyoshenka:
Next steps: 
 - lbry daemon errors if address is invalid (not a priority now)
 - lbry app checks address using regex or real base58 check

PR #869, opened 2017-08-24
fix setup.py

ofek:
`.decode('utf-8')` is unnecessary and breaks on Python 3


jackrobison:
I think this is necessary: https://github.com/lbryio/lbry/commit/4e765b265f7e91bc9b48dffbadf07f63ac004af8

ofek:
@jackrobison Ok then, fixed.

lyoshenka:
@jackrobison bump

PR #872, opened 2017-08-25
Reflector rework (for head blob announce)

kaykurokawa:
When running lbrynet as a reflector cluster, it previously did not use the stream info manager at all when receiving streams from clients. Thus a reflector cluster had no knowledge of streams and would  not be able to use head blob announce. 

This adds stream info manager to the reflector server, and uses this to mark blobs as "should_announce" or not. test_reflectory.py has been modified to check if the stream info is handled properly and blobs are marked as "should_announce" if appropriate.

### DIFF
--- a/lbrynet/reflector/server/server.py
+++ b/lbrynet/reflector/server/server.py
@@ -5,6 +5,8 @@ from twisted.internet import error, defer
 from twisted.internet.protocol import Protocol, ServerFactory
 from lbrynet.core.utils import is_valid_blobhash
 from lbrynet.core.Error import DownloadCanceledError, InvalidBlobHashError
+from lbrynet.core.StreamDescriptor import BlobStreamDescriptorReader
+from lbrynet.lbry_file.StreamDescriptor import save_sd_info
 from lbrynet.reflector.common import REFLECTOR_V1, REFLECTOR_V2
 from lbrynet.reflector.common import ReflectorRequestError, ReflectorClientVersionError
 
@@ -30,6 +32,7 @@ class ReflectorServer(Protocol):
         log.debug('Connection made to %s', peer_info)
         self.peer = self.factory.peer_manager.get_peer(peer_info.host, peer_info.port)
         self.blob_manager = self.factory.blob_manager
+        self.stream_info_manager = self.factory.stream_info_manager
         self.protocol_version = self.factory.protocol_version
         self.received_handshake = False
         self.peer_version = None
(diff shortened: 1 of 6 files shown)


lbrynet/reflector/server/server.py
```diff
@@ -63,7 +66,23 @@ def clean_up_failed_upload(self, err, blob):
...
+        should_announce = False
+        if response_key == RECEIVED_SD_BLOB:
+            sd_info = yield BlobStreamDescriptorReader(blob).get_info()
+            yield save_sd_info(self.stream_info_manager, sd_info)
+            yield self.stream_info_manager.save_sd_blob_hash_to_stream(sd_info['stream_hash'],
+                                                                       blob.blob_hash)
+            should_announce = True
+
+        else:
+            stream_hash = yield self.stream_info_manager.get_stream_of_blob(blob.blob_hash)
```
jackrobison: Why announce if the blob isn't an sd blob? If the sd blob hasn't completed, then stream hash would be none and the blob would have `should_announce` as false. If the sd blob had completed, then it and each subsequent stream blob received would have `should_announce` is true. If that's the desired behavior, then it should probably also go through any data blobs received before the sd blob completed and mark should_announce as true for them too.

kaykurokawa: should_announce is set to True if its a SD blob or if its a head blob (blob_num == 0) 

Since the first line of the function sets should_announce to False, all other blobs should have should_announce == False 

I don't think its possible to receive blobs from clients before receiving the sd blobs when receiving streams.

jackrobison: If the head blob was sent via reflector v1 before the sd blob, then only the sd blob would be announced. If the sd blob was also sent via reflector v1 it wouldn't be announced either, since the response key would be RECEIVED_BLOB.

These could be fixed by checking that the sd blob is known by the stream info manager in `handle_descriptor_request` and by scheduling an immediate_announce callLater for the head blob if it already exists when `_on_completed_blob` is called for a sd blob.

kaykurokawa:
Rebased to master 
Added a unit test in test_reflector to check blob v1 protocol does affect stream information

kaykurokawa:
I think this is good to merge, 

but noting here that there should probably a future change to Hash announcer for it to not announce invalid blobs, instead of doing it in set_should_announce()

kaykurokawa:
Adding this on top https://github.com/lbryio/lbry/pull/910 , so this single PR is all that's required for head blob announce

PR #875, opened 2017-08-28
Updating documentation for clarity

wallermadev:
Updating `claim_new_support` documentation to add a bit more clarity to the name field.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -188,7 +188,7 @@ Returns
 Support a name claim
 
 Args:
-    'name': (str) name
+    'name': (str) Name of claim
     'claim_id': (str) claim ID of claim to support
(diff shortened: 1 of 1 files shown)


MSFTserver:
upon entering a different name for this command it generates a new url and replaces the old one within the publishes section of the app and doesn't know what to so with itself, the name I think should be allocated from the id I would presume?
[.GIF of Issue](https://files.slack.com/files-pri/T0AFFTU95-F6UFHL6Q3/5.gif)

jackrobison:
@lyoshenka @wallermadev 

This needs to be in the docstring, these changes in the markdown file will be overwritten the next time mkdocs gets run.

Issue #890, opened 2017-09-07
Google BTC to USD feed failing


tzarebczan:
## The Issue
The Google BTC to USD feed is failing because http://finance.google.com/finance/info?q=CURRENCY:BTC doesn't work anymore.  https://www.google.com/finance?q=CURRENCY:BTC does. In the short term, this link needs to be fixed. In the long term, we need to have redundant APIs for all conversion calls. 

Daemon throws: `2017-09-07 11:46:05,375 WARNING  lbrynet.daemon.ExchangeRateManager:69: There was a problem updating USDBTC exchange rate information from Coinbase via Google finance`

Code:
https://github.com/lbryio/lbry/blob/aef31991f9bf2751d5d980b0a2074caf19af1b14/lbrynet/daemon/ExchangeRateManager.py

### Expected behaviour
Price API should work or fall back to backup(s)

### Actual behaviour
Price API call fails

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

jackrobison:
A patch release for 0.15.1 (broken latest release) is [here](https://github.com/lbryio/lbry/tree/v0.15.2)

The fix on master is [here](https://github.com/lbryio/lbry/commit/1f4dfe1f933e5da2b677c26e9e81c04eb4ee187a)

PR #891, opened 2017-09-07
Quick Fix Google Finance BTC Feed

MSFTserver:
fix for issue https://github.com/lbryio/lbry/issues/890
seems https://www.google.com/finance/info?q=CURRENCY:BTC doesnt exist anymore and instead dropped the /info from the link and the url works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ at anytime.
 
 ### Fixed
   * Fixed uncaught error when shutting down after a failed daemon startup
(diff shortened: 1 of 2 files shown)


jackrobison:
Thanks, although unfortunately they also changed the response format. Rather than wrestling with this I just dropped the usage of google finance in favor of the lbry.io api. 

Here's the [patch](https://github.com/lbryio/lbry/commit/12e938aa17f2557d4c7f7f4ae41964f5c6d71b14) for 0.15.2

PR #897, opened 2017-09-11
[Testing] use test reflector instead of reflector.lbry.io

kaykurokawa:
Use a temporary test reflector (with head blob announce enabled) for clients to upload things to.


lyoshenka:
@kaykurokawa why is this in "needs review"?

PR #910, opened 2017-09-20
Turn announce by head blob "on" by default

kaykurokawa:
This sets "announce_head_blobs_only to True in conf.py

### DIFF
--- a/lbrynet/core/BlobManager.py
+++ b/lbrynet/core/BlobManager.py
@@ -226,6 +226,14 @@ class DiskBlobManager(DHTHashSupplier):
         return d
 
     @rerun_if_locked
+    @defer.inlineCallbacks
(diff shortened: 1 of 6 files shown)


kaykurokawa:
made it rely on https://github.com/lbryio/lbry/pull/872

lbrynet/core/BlobManager.py
```diff
@@ -226,6 +226,14 @@ def _get_all_blob_hashes(self):
         return d
 
     @rerun_if_locked
+    @defer.inlineCallbacks
+    def _get_all_should_announce_blob_hashes(self):
```
jackrobison: I think that for the tests you should mock time.time() and use the real `_get_blobs_to_announce` call, since this function isn't the same as the one that's used for real.

Issue #908, opened 2017-09-20
Daemon refuses to start if DNS cannot resolve lbry.io domain


lyoshenka:
The error message says it cannot connect to the internet, but the truth is that it cannot connect to lbry.io because the DNS resolve failed. Internet connectivity check needs to be more robust.

```
$ /opt/lbry/lbrynet-daemon --version
{
  "build": "release",
  "installation_id": "5GfsGqN89ysnaQiorZZNg3P3vDbc5WWJ75FbNDEiCoZXpzE4beznwMTHbNPYLww8gA",
  "lbrynet_version": "0.16.0",
  "lbryschema_version": "0.0.11",
  "lbryum_version": "3.1.7",
  "os_release": "4.4.0-93-generic",
  "os_system": "Linux",
  "platform": "Linux-4.4.0-93-generic-x86_64-with-Ubuntu-16.04-xenial",
  "processor": "x86_64",
  "python_version": "2.7.12"
}
$ /opt/lbry/lbrynet-daemon
2017-09-20 13:40:34,369 INFO     lbrynet.daemon.DaemonControl:70: Starting lbrynet-daemon from command line
2017-09-20 13:40:34,373 INFO     lbrynet.core.utils:108: Failed to connect to lbry.io:80. Maybe the internet connection is not working
Traceback (most recent call last):
  File "site-packages/lbrynet/core/utils.py", line 101, in check_connection
gaierror: [Errno -2] Name or service not known
2017-09-20 13:40:34,374 INFO     lbrynet.daemon.DaemonControl:77: Not connected to internet, unable to start
```

rkrp:
I would like to take this up! Would appreciate any pointers.

tzarebczan:
@rkrp looks like we already have a PR for this, but it wasn't merged yet.

jackrobison:
Merged #912 [here](https://github.com/lbryio/lbry/commit/b9e03a7f134bcc1704052a525426da28550a92d3)

PR #912, opened 2017-09-20
Modified check_connection() to try bypassing DNS before giving up.

mirgee:
Addresses #908.

### DIFF
--- a/lbrynet/core/utils.py
+++ b/lbrynet/core/utils.py
@@ -94,14 +94,29 @@ def obfuscate(plain):
     return base64.b64encode(plain).encode('rot13')
 
 
-def check_connection(server="lbry.io", port=80):
+def check_connection(server="lbry.io", port=80, timeout=2, bypass_dns=False):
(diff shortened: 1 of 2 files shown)


tzarebczan:
@mirgee thanks for the PR. We'll have someone review it. In the meantime, check out https://lbry.io/faq/tips !

lbrynet/core/utils.py
```diff
@@ -94,14 +94,29 @@ def obfuscate(plain):
...
 
-def check_connection(server="lbry.io", port=80):
+def check_connection(server="lbry.io", port=80, timeout=2, bypass_dns=False):
     """Attempts to open a socket to server:port and returns True if successful."""
+    log.debug('Checking connection to %s:%s', server, port)
     try:
-        log.debug('Checking connection to %s:%s', server, port)
-        host = socket.gethostbyname(server)
-        s = socket.create_connection((host, port), 2)
+        if not bypass_dns:
```
jackrobison: I don't think there should be a `bypass_dns` parameter since the exception is handled regardless

Issue #901, opened 2017-09-15
`publish` returns useless error


lyoshenka:
Run `lbrynet-cli publish stuff 1`. The daemon responds with `ERROR: 'source'`. That's it. What does that mean? What am I supposed to do about that?

mirgee:
It seems that this call should make stream type name claim which is required to contain 'source' in its metadata at several places. Can anyone please explain the difference between certificate and stream type name claims?

jackrobison:
@mirgee 

https://github.com/lbryio/lbryschema/blob/master/docs/schema.md

https://github.com/lbryio/lbryschema/blob/master/docs/signing.md

lyoshenka:
Fixed in https://github.com/lbryio/lbry/issues/913

PR #913, opened 2017-09-21
better publish errors

jackrobison:


lbrynet/daemon/Daemon.py
```diff
@@ -1841,8 +1842,25 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
...
+            if not os.path.isfile(file_path):
+                raise Exception("invalid file path to publish")
+            claim_copy['stream']['source'] = {
+                'version': '_0_0_1',
+                'sourceType': 'lbry_sd_hash',
+                'source': '0' * 96,
+                'contentType': ''
+            }
+        else:
+            raise Exception("no source provided to publish")
```
mirgee: The documentation says "If no path is given but a metadata dict is the source from the given metadata will be used". This doesn't seem to be intended behavior here. Maybe update the doc, or change the logic?

mirgee:
Shouldn't this be linked to #901?

lyoshenka:
please rebase and merge @jackrobison

lbrynet/daemon/Daemon.py
jackrobison: updated the docstring

Issue #923, opened 2017-09-26
`ConnectBindError: Too many open files` on daemon startup


lyoshenka:
`ERROR    lbrynet.reflector.client.client:365: Reflector connection failed: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectBindError'>: Couldn't bind: 24: Too many open files.`

Kay says its due to reflector/client.py not closing blob files once it finishes sending it

PR #924, opened 2017-09-27
Fix blob reader closing

kaykurokawa:
Fix for https://github.com/lbryio/lbry/issues/923 and also some continuing work to improve the blob classes (https://github.com/lbryio/lbry/pull/898)

Blob file opened for reading ( BlobFile.open_for_reading() ) was not being closed after file transfer , thus during client reflection, too many blob files would be opened. Commit a4ea49c fixes this. 

This kind of mistake could be easily made since blob files opened for reading had to be closed by calling BlobFile.close_read_handle(handle) . We can now just call handle.close() to close the blob files.

### DIFF
--- a/lbrynet/blob/blob_file.py
+++ b/lbrynet/blob/blob_file.py
@@ -75,19 +75,17 @@ class BlobFile(object):
         """
         open blob for reading
 
-        returns a file handle that can be read() from.
-        once finished with the file handle, user must call close_read_handle()
-        otherwise blob cannot be deleted.
+        returns a file like object that can be read() from, and closed() when
+        finished
         """
         if self._verified is True:
-            file_handle = None
             try:
-                file_handle = open(self.file_path, 'rb')
+                reader = HashBlobReader(self.file_path, self.reader_finished)
                 self.readers += 1
-                return file_handle
+                return reader
             except IOError:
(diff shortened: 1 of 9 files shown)


lbrynet/blob/blob_file.py
```diff
@@ -150,12 +144,16 @@ def is_downloading(self):
...
+        This function is only used in StreamBlobDecryptor
+        and should be deprecated in favor of open_for_reading()
+        """
         def close_self(*args):
             self.close_read_handle(file_handle)
             return args[0]
 
         file_sender = FileSender()
-        reader = HashBlobReader(write_func)
+        reader = HashBlobReader_v0(write_func)
```
jackrobison: Since this was the only usage of `HashBlobReader`, there doesn't need to be a new `HashBlobReader_v0` class.

kaykurokawa: It's possible, FYI the new class is HashBlobReader , and the old class was renamed from HashBlobReadr to HashBlobReader_v0 ,

HashBlobReader_v0 provides a read interface to StreamBlobDecryptor in a producer/consumer pattern.  But StreamBlobDecryptor can be reworked to use a read /write handle interface.

Issue #854, opened 2017-08-15
Solve/design the general approach for making daemon calls asynchronous


tzarebczan:
@tzarebczan commented on [Tue Aug 15 2017](https://github.com/lbryio/lbry-app/issues/475)

 

## The Issue
As a Publisher, when I upload files to LBRY, the current process is fairly uninformative after the Publish button is clicked. There is no indication that the file is being processed or when it's finished uploading (besides the URI being available).  This is a bigger issue with larger files as they take some time to upload.  

Feel free to add/comment. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

## Screenshots

---

@kauffj commented on [Tue Aug 15 2017](https://github.com/lbryio/lbry-app/issues/475#issuecomment-322553552)

This isn't an app issue. If you want, you can re-open on lbryio/lbry as "Make publish and get calls asynchronous"

lyoshenka:
Next step: figure out if Websockets is the right solution to this

lyoshenka:
Related: http://wamp-proto.org/

kauffj:
@lyoshenka I think this possibly conflates two different issues. We don't need sockets or the ability for the daemon to push data to apps to solve this (though this is a great idea!).

Currently, existing apps _already_ regularly poll `get` and `file_list` to determine state. This change is simply to have `get` and `publish` return immediately, with some kind of field indicating their status ("pending" or whatever).

lyoshenka:
Thought about this some. I think there are times when we want it to return right away, and there are times when we want it to return when its done. I think the latter is more common, so we should add a `--return-immediately` flag.

eukreign:
`publish` is significantly faster with the new wallet, closing this as the original issue may no longer be valid.

tzarebczan:
I don't think we want to close this just becuase publish reruns faster, there are still valid reasons for this requirement. If a file is large, we would want a way to return a loading status.

This ticket also included a similar feature for the get command. 

@lyoshenka / @jack thoughts?

eukreign:
Would it suffice to say that adding a websockets API would solve this problem? I made an epic: https://github.com/lbryio/lbry/issues/1649

kauffj:
@eukreign yes, we can use that moving forward

PR #925, opened 2017-09-27
Improve download errors

jackrobison:


lbrynet/daemon/Daemon.py
```diff
@@ -1435,18 +1471,21 @@ def jsonrpc_resolve(self, force=False, uri=None, uris=[]):
...
     @AuthJSONRPCServer.auth_required
     @defer.inlineCallbacks
-    def jsonrpc_get(self, uri, file_name=None, timeout=None):
+    def jsonrpc_get(self, uri, file_name=None, timeout=None, delete_on_timeout=True):
         """
         Download stream from a LBRY name.
 
         Usage:
             get <uri> [<file_name> | --file_name=<file_name>] [<timeout> | --timeout=<timeout>]
+                      [<delete_on_timeout> | --delete_on_timeout=<delete_on_timeout>]
```
lyoshenka: whats the point of this flag? does anyone actually want to do this? cant they just use `file_delete`?

tests/unit/lbrynet_daemon/test_Downloader.py
```diff
@@ -115,31 +111,30 @@ def download_sd_blob(self):
...
-        name='test'
-        stream_info = None
-        start = getstream.start(stream_info,name)
-        self.clock.advance(1)
-        self.clock.advance(1)
-        with self.assertRaises(DownloadTimeoutError):
-            yield start
-        self.assertTrue(getstream.downloader.stop_called)
-        self.assertTrue(getstream.pay_key_fee_called)
+    # @defer.inlineCallbacks
```
lyoshenka: why did you get rid of these tests? i feel like we want more tests, not fewer

lbrynet/daemon/Daemon.py
kauffj: Agreed, this seems backwards. We're supposed to be moving towards an asynchronous get (#854), not adding more synchronicity.

tests/unit/lbrynet_daemon/test_Downloader.py
kaykurokawa: i'm gonna take a closer looker at these, i had a quick conversation with jack about it

lbrynet/daemon/Downloader.py
```diff
@@ -150,6 +152,12 @@ def finish(self, results, name):
         self._check_status(status)
         defer.returnValue(self.download_path)
```
kaykurokawa: I think this fail() function just needs to stop the looping call and reraise the err, 
It's attached as an errback to self.downloader.start(), (downloader is an ManagedEncryptedFileDownloader, it doesn't raise DownloadDataTimeout)  , I think you probablly mistook this as being attached as an errback to Downloader.start()

lbrynet/daemon/Downloader.py
```diff
@@ -213,9 +213,8 @@ def start(self, stream_info, name):
 
         try:
             yield self.data_downloading_deferred
-        except Exception as err:
-            self.downloader.stop()
```
kaykurokawa: So where is this being stopped now?

jackrobison: https://github.com/lbryio/lbry/pull/925/files#diff-309f9696e2149ec1146e8674e77e4e49R666

kaykurokawa: nvm I see, I think this is probablly a good idea , to remove the management of downloader from  outside of this class

lbrynet/daemon/Daemon.py
```diff
@@ -621,17 +654,20 @@ def _download_name(self, name, claim_dict, claim_id, timeout=None, file_name=Non
                                                file_name)
             try:
                 lbry_file, finished_deferred = yield self.streams[claim_id].start(claim_dict, name)
-                finished_deferred.addCallback(
-                    lambda _: self.analytics_manager.send_download_finished(download_id,
-                                                                            name,
-                                                                            claim_dict))
+                finished_deferred.addCallback(lambda _: self._download_finished(download_id, name,
+                                                                                claim_dict))
                 result = yield self._get_lbry_file_dict(lbry_file, full_status=True)
```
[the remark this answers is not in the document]
kaykurokawa: Also , we need to stop self.streams[claim_id].downloader ( this is same as variable lbry_file ) regardless of what the Exception is here

lbrynet/daemon/Downloader.py
```diff
@@ -62,9 +61,7 @@ def download_path(self):
         return os.path.join(self.download_directory, self.downloader.file_name)
 
     def _check_status(self, status):
-        stop_condition = (status.num_completed > 0 or
-                status.running_status == ManagedEncryptedFileDownloader.STATUS_STOPPED)
-        if stop_condition and not self.data_downloading_deferred.called:
```
kaykurokawa: This is related to the last unit test you commented out. 

I think the problem here was that if the user manually stops the download before the first blob is finished.  But I think looking a the daemon, that should probably not be possible for them to do now. Was there a reason you changed this?

kaykurokawa:
If fixed one of the test you commented out, 

I think the other test you commented out can be removed, wanna hear jack's thought on the last comment i made.

lbrynet/daemon/Downloader.py
jackrobison: >Was there a reason you changed this?

It caused a false positive in the stop condition

>should probably not be possible for them to do now

It can happen if the daemon gets shut down, which should now be handled by `cancel`

lbrynet/daemon/Downloader.py
jackrobison: changed it to this

PR #931, opened 2017-09-29
Fixes to Blob (unify way to read from BlobFile, move conf.py BLOB_SIZE  )

kaykurokawa:
8e70c0e
for HashBlobReader and HashBobWriter, add destructors that closes read/write handles if not closed already. Since this should not happen if the class is used properly,  log a Warn if destructor is called without handles being closed ( Hopefully we catch any missing close() calls this way)

eb40ffa
Have HashBlobReader take read handle instead of a file path as argument. If open() throws an exception, we don't need to worry about HashBlobReader being ever created ( thus no need to handle cases of HashBlobReader.close() without a read handle)   

4327f9d
906e8b0
10c4551
Removed StreamBlobDecryptor's way of reading from BlobFile by using consumer/producer pattern. Now we have StreamBlobDecryptor just utilize BlobFile.open_for_reading() to get a read handle, and use a deferToThread for decrypting from the read handle on a seperate thread. 

8f5166a
Moved the BLOB_SIZE fixed setting from conf.py, to MAX_BLOB_SIZE in blob/blob_file.py . This way prism can import lbrynet.blob classes without needing to set up the config. I think this is also just a better place for this constant as changing it pretty much breaks everything.

### DIFF
--- a/lbrynet/blob/blob_file.py
+++ b/lbrynet/blob/blob_file.py
@@ -3,15 +3,14 @@ import os
 from twisted.internet import defer, threads
 from twisted.web.client import FileBodyProducer
 from twisted.python.failure import Failure
-from lbrynet import conf
 from lbrynet.core.Error import DownloadCanceledError, InvalidDataError, InvalidBlobHashError
 from lbrynet.core.utils import is_valid_blobhash
 from lbrynet.blob.writer import HashBlobWriter
 from lbrynet.blob.reader import HashBlobReader
 
-
 log = logging.getLogger(__name__)
 
+MAX_BLOB_SIZE = 2 * 2 ** 20
 
 class BlobFile(object):
     """
@@ -124,7 +123,7 @@ class BlobFile(object):
     def set_length(self, length):
         if self.length is not None and length == self.length:
             return True
-        if self.length is None and 0 <= length <= conf.settings['BLOB_SIZE']:
+        if self.length is None and 0 <= length <= MAX_BLOB_SIZE:
             self.length = length
             return True
         log.warning("Got an invalid length. Previous length: %s, Invalid length: %s",
--- a/lbrynet/blob/blob_file.py
+++ b/lbrynet/blob/blob_file.py
@@ -78,7 +78,8 @@ class BlobFile(object):
         finished
         """
         if self._verified is True:
-            reader = HashBlobReader(self.file_path, self.reader_finished)
+            f = open(self.file_path, 'rb')
+            reader = HashBlobReader(f, self.reader_finished)
             self.readers += 1
             return reader
         return None
--- a/lbrynet/blob/blob_file.py
+++ b/lbrynet/blob/blob_file.py
@@ -1,14 +1,13 @@
 import logging
 import os
 from twisted.internet import defer, threads
-from twisted.protocols.basic import FileSender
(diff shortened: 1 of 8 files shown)


kaykurokawa:
rebased to master

kaykurokawa:
rebased again to master

lbrynet/cryptstream/CryptBlob.py
```diff
@@ -67,13 +70,17 @@ def finish_decrypt():
...
+        def decrypt_bytes():
+            data = read_handle.read()
             self.buff += data
             self.len_read += len(data)
             write_bytes()
+            finish_decrypt()
 
-        d = self.blob.read(decrypt_bytes)
-        d.addCallback(lambda _: finish_decrypt())
+        d = threads.deferToThread(decrypt_bytes)
```
jackrobison: Why deferToThread?

kaykurokawa: Before this PR StreamBlobDecryptor used FileSender() and the producer/consumer pattern to decrypt blobs in an asynchronous manner. 

If we dont't deferToThread, this will block the reactor till a standard read()  and write() to a normal file handle is finished ( the write() is called by write_bytes() see lbry_file/client/EncryptedFileDownloader.py EncryptedFileSaver._get_write_func() )

lbrynet/blob/reader.py
```diff
@@ -1,50 +1,22 @@
...
-    def unregisterProducer(self):
-        pass
-
-    def write(self, data):
-        from twisted.internet import reactor
-
-        self.write_func(data)
-        if self.streaming is False:
-            reactor.callLater(0, self.producer.resumeProducing)
-
```
kaykurokawa: This class here sets up the producer/consumer pattern

lbrynet/blob/blob_file.py
```diff
@@ -142,33 +141,6 @@ def is_downloading(self):
...
-            self.close_read_handle(file_handle)
-            return args[0]
-
-        file_sender = FileSender()
-        reader = HashBlobReader_v0(write_func)
-        file_handle = self.open_for_reading()
-        if file_handle is not None:
-            d = file_sender.beginFileTransfer(file_handle, reader)
-            d.addCallback(close_self)
-        else:
```
kaykurokawa: you can see here that it uses FileSender() so that StreamBlobDecryptor can decrypt blobs asynchronously

lbrynet/cryptstream/CryptBlob.py
jackrobison: I think it would be better to pass the read_handle to a FileBodyProducer and call decrypt_bytes on the result of that instead of doing the read() in the decrypt_bytes function

kaykurokawa:
I changed it to use the FileBodyProducer instead of using deferToThread. I think that is the more twisted way to do it,  but it will block on the write, which probably isn't that big of a deal. 

I think the next step is to write the consumer, which will prevent blocking on the write. This will require some refactoring of the StreamBlobDecryptor. 

Rebased to master.

lyoshenka:
@kaykurokawa is this done?

jackrobison:
LGTM

kaykurokawa:
rebased to master and merging

PR #932, opened 2017-10-02
Reflector client bug 

kaykurokawa:
Fix inconsistency in how EncryptedFileReflectorClient.blob_hashes_to_send is used, preventing proper reflecting.

Some functions put blob hashes in EncryptedFileReflectorClient.blob_hahes_to_send , others put BlobFile's in them. EncryptedFileReflectorClient.open_blob_for_reading() expected BlobFiles so if it received blob hashes, it errored out. 

I fixed it so that EncryptedFileReflectorClient.blob_hashes_to_send always contains blob hashes. And EncryptedFileReflectorClient.open_blob_for_reading() would always be fed BlobFiles.

This looks like a bug that's been here forever.. It looks like it should have prevented any partial reflects from working out properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ at anytime.
   * Fixed redundant blob requests to a peer
   * Fixed https://github.com/lbryio/lbry/issues/923
   * Fixed concurrent reflects opening too many files
+  * Fixed cases when reflecting would fail on error conditions
 
 ### Deprecated
   * Deprecated `blob_announce_all` JSONRPC command. Use `blob_announce` instead.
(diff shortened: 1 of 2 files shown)


kaykurokawa:
Before, these cases would cause failures because of appending blob hashes to EncryptedFileReflectorClient.blob_hashes_to_send

https://github.com/lbryio/lbry/blob/master/lbrynet/reflector/client/client.py#L256

https://github.com/lbryio/lbry/blob/master/lbrynet/reflector/client/client.py#L292

https://github.com/lbryio/lbry/blob/master/lbrynet/reflector/client/client.py#L231

PR #935, opened 2017-10-04
DHT fixes

jackrobison:


lbrynet/dht/delay.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Delay(object):
+    maxToSendDelay = 10 ** -3  # 0.05
+    minToSendDelay = 10 ** -5  # 0.01
+
+    def __init__(self, start=0):
+        self._next = start
+
+    # TODO: explain why this logic is like it is. And add tests that
```
lyoshenka: can you do this now, while this is fresh in your mind

lyoshenka: at least the explaining part, but tests would be great as well

lbrynet/dht/node.py
```diff
@@ -425,14 +421,22 @@ def parseResults(nodes):
...
-    def ping(self):
+    def ping(self, **kwargs):
         """ Used to verify contact between two Kademlia nodes
 
         @rtype: str
         """
         return 'pong'
 
     @rpcmethod
+    def echo(self, value, **kwargs):
```
lyoshenka: why do we need this?

jackrobison: I'd initially fixed it because it was necessary to get testRPCRequestArgs to pass. Since it's not one of the 4 rpc commands in the kademlia spec I removed it and updated the test to use `ping`.

lbrynet/dht/delay.py
jackrobison: I just moved it to its own file, it haven't deciphered it myself yet.

lbrynet/core/Session.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, blob_data_payment_rate, db_dir=None,
 
         self.rate_limiter = rate_limiter
 
-        self.external_ip = '127.0.0.1'
+        self.external_ip = None
```
kaykurokawa: You said this happens when UPNP is not enabled? 
Maybe log a warning here? 
Also does this still work if someone wants to test the DHT on their local network? 
( I should probablly run lbry-in-a-box testing for this)

lbrynet/dht/protocol.py
```diff
@@ -24,6 +24,14 @@
 
 log = logging.getLogger(__name__)
 
+REMOTE_EXCEPTIONS = {
+    "exceptions.%s" % e: getattr(exceptions, e) for e in dir(exceptions) if not e.startswith("_")
+}
+
```
kaykurokawa: This a list of built in exceptions no? 
Kind of confusing , may be rename it to BUILTIN_EXCEPTIONS

lbrynet/core/Session.py
```diff
@@ -241,6 +242,11 @@ def threaded_try_upnp():
 
         def upnp_failed(err):
             log.warning("UPnP failed. Reason: %s", err.getErrorMessage())
+            with TXRequestsSession() as session:
+                r = session.get('https://api.ipify.org', format='json')
+                log.info("got external ip from ipify.org")
+                self.external_ip = r.json()['ip']
+                log.info("Set external ip: %s", self.external_ip)
             return False
```
kaykurokawa: Does it make sense to do this? If UPNP failed, what good is setting an external ip that you can't listen to?

kaykurokawa:
I added some comments (marked as outdated for some reason), 
I'll try running this patch in lbry-in-a-box

lbrynet/core/Session.py
```diff
@@ -98,112 +93,109 @@ def __init__(self, blob_data_payment_rate, db_dir=None,
...
 
+        hosts = []
+        for host, port in self.known_dht_nodes:
+            h = yield reactor.resolve(host)
+            hosts.append((h, port))
+
+        yield self.dht_node.joinNetwork(tuple(hosts))
+        self.peer_finder.run_manage_loop()
+        self.hash_announcer.run_manage_loop()
+
```
kaykurokawa: Can't do self.hash_announcer.run_manage_loop() without running self.blob_manager.setup() first since hash announcer depends on blob manager.

kaykurokawa:
Looks like there's some failures in the lbry-in-a-box regtest where it checks the DHT for existence of published hashes. Not sure the cause.

lbrynet/core/Session.py
lyoshenka: How can we make this dependency more real? How can we get the daemon to error when the dependency is not  met?

PR #940, opened 2017-10-09
Android tests refactoring

akinwale:
* Refactored asserts due to the `PYTHONOPTIMIZE` flag set for the Android build.
* Return empty string as the default value if the `__doc__` attribute is `None` in cases where the docstring isn't present (also due to the `PYTHONOPTIMIZE` flag)


akinwale:
@jackrobison Updated raised errors and tests as per your comments in Slack.

lbrynet/daemon/ExchangeRateManager.py
```diff
@@ -17,8 +17,10 @@
 
 class ExchangeRate(object):
     def __init__(self, market, spot, ts):
-        assert int(time.time()) - ts < 600
-        assert spot > 0
+        if not int(time.time()) - ts < 600:
+            raise AssertionError()
+        if not spot > 0:
+            raise AssertionError()
         self.currency_pair = (market[0:3], market[3:6])
```
kaykurokawa: might as well make these of type ValueError  for clarity (first is because timestamp is too dated, second is because spot must be more than 0)

kaykurokawa:
Is there a reason android must be compiled with PYTHONOPTIMIZE flag?

akinwale:
> Is there a reason android must be compiled with PYTHONOPTIMIZE flag?

I'm not sure what the exact reason is as it is a setting from the default kivy / python for android app launcher build. I tried to remove it but that resulted in errors launching the Python modules when trying to run the app.

jackrobison:
LGTM

PR #945, opened 2017-10-11
video content duration support

KamilPiekutowski:
Bellow is the schema that it's suppose to work with.
New publish item gets a duration represented by milliseconds with integer.
It requires ffmpeg to be intalled on the system.
If anyone has a better suggestion how retrieve mp4 duration using python module only,
I am open to change  it.
I can also add duration if audio is being published, but i need a list of formats that lbry will handle.

Bare in mind that this is a first attempt, so probably more will get either added or changed.

[https://github.com/lbryio/lbryschema/pull/28](url)


lbrynet/daemon/MediaDescriptor.py
```diff
@@ -0,0 +1,19 @@
...
+class Duration(object):
+    def __init__(self, time):
+        (self.hours, self.minutes, self.seconds) = time
+    def all_in_millisec(self):
+        return (int((float(self.hours) * 3600) +
+               (float(self.minutes) * 60) + (float(self.seconds))) * 1000)
+
+def get_video_length(path):
+    process = subprocess.Popen(['/usr/bin/ffmpeg', '-i', path],
+              stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
kaykurokawa: this is not a portable way of getting the video duration (won't work on windows, won't work if you have ffmpeg installed elsewhere). 

It also only works with ffmpeg and does not handle cases when the file is some other file type.

lbrynet/daemon/testMedia.py
```diff
@@ -0,0 +1,3 @@
+from MediaDescriptor import get_video_length
+
```
kaykurokawa: what is this for?

lbrynet/daemon/MediaDescriptor.py
KamilPiekutowski: Did you read my comment? I mentioned it there. I was wondering if someone knows a good portable library that can take duration out of mp4. I also asked on slack. No one has yet pointed one out.

lbrynet/daemon/testMedia.py
KamilPiekutowski: that is is a big oops. It shouldn't be there. Sorry.

lbrynet/daemon/MediaDescriptor.py
jackrobison: Searching around, this looked promising: https://pypi.python.org/pypi/tinytag/

KamilPiekutowski:
oh great I will look at it tonight.

jackrobison:
The more I think about it I do think content duration should be checked in lbrynet-daemon, and not in the app. Please update to find the content duration using a python library - `lbrynet` should not have binary dependancies like ffmpeg. The `tinytag` module looks like it would work.

tzarebczan:
@KamilPiekutowski Did you still want to work on this based on Jack's comments?

PR #948, opened 2017-10-12
Update the year in the LICENSE

HugoHugo:
The current LICENSE contains copyright with the year 2016, so I propose to update the year to 2017.


tzarebczan:
Thanks for noticing it and fixing it, we love contributions! Check out https://lbry.io/faq/tips if you'd be interested in a reward!

Issue #905, opened 2017-09-19
v16: KeyError: 0 in get_head_blob_hash for old failed downloads


tzarebczan:
## The Issue
Noticed this issue while starting up the latest version of the daemon. I have an older install, so it looks like it tries to start some previously uncomplete/failed downloads while iterating through my claims. This could be part of a bigger issue, I'll see if I can get it to reproduce in other ways.  Couple examples where this happens are in the logs at the end. 

### Steps to reproduce
1. Start Daemon
2. Check logs for errors
3. See Key error after previously failed downloads try to start. 

### Expected behaviour
Error more gracefully?

### Actual behaviour
KeyError

## System Configuration

 

- LBRY Daemon version: 0.16RC1
- LBRY App version: 0.16RC1
- LBRY Installation ID:
- Operating system: Windows

## Anything Else

```
2017-09-19 09:47:51,891 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://battlefront-ii-storytrailer (5c3d3c) is finished
2017-09-19 09:47:51,904 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://hitmans-bodyguard-trailer (3303b0) is running
2017-09-19 09:47:51,914 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://thelovewitch-trailer (59edac) is finished
2017-09-19 09:47:51,964 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://bellflower-trailer (618fbb) is finished
2017-09-19 09:47:51,986 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://itsadisaster-threesome (395066) is finished
2017-09-19 09:47:51,987 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://jinglebellrocks-trailer (224aa1) is finished
2017-09-19 09:47:52,009 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://pub (57fdb5) is finished
2017-09-19 09:47:52,025 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://lbryscrreen1 (2dab0b) is finished
2017-09-19 09:47:52,072 INFO     lbrynet.file_manager.EncryptedFileDownloader:32: lbry://lbryscrreen1 (2dab0b) is finished
[...]

PR #949, opened 2017-10-12
Fix bug where last blob would not get saved in blob manager

kaykurokawa:
Found bug where if you were creating a stream, the last blob in a stream would not get saved to the blob manager (It was still being created, reflected, and everything else however). See the added test 574660c for the failure case. 

This means that if you created a stream ,  you would never announce the last blob in the stream because the announcement blobs were obtained from the blob manager. (However, this bug would not prevent reflection, or reflector from announcing the blob). 

Might be a potential fix for https://github.com/lbryio/lbry/issues/905

### DIFF
--- a/lbrynet/cryptstream/CryptStreamCreator.py
+++ b/lbrynet/cryptstream/CryptStreamCreator.py
@@ -89,7 +89,6 @@ class CryptStreamCreator(object):
         self._finalize()
         dl = defer.DeferredList(self.finished_deferreds)
         dl.addCallback(lambda _: self._finished())
-        dl.addErrback(self._error)
         return dl
 
     # TODO: move the stream creation process to its own thread and
@@ -144,9 +143,6 @@ class CryptStreamCreator(object):
 
     def _get_blob_maker(self, iv, blob_creator):
         return CryptStreamBlobMaker(self.key, iv, self.blob_count, blob_creator)
-
-    def _error(self, error):
-        log.error(error)
 
     def _finished(self):
         raise NotImplementedError()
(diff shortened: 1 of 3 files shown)


lbrynet/cryptstream/CryptStreamCreator.py
```diff
@@ -69,17 +69,23 @@ def unregisterProducer(self):
...
         """Stop creating the stream. Create the terminating zero-length blob."""
         log.debug("stop has been called for StreamCreator")
         self.stopped = True
         if self.current_blob is not None:
-            current_blob = self.current_blob
-            d = current_blob.close()
-            d.addCallback(self._blob_finished)
-            d.addErrback(self._error)
-            self.finished_deferreds.append(d)
-            self.current_blob = None
```
kaykurokawa: When calling stop, current_blob was getting closed but not being written tot he blob_manager as it would be in _write() function

lbrynet/cryptstream/CryptStreamCreator.py
```diff
@@ -69,17 +69,23 @@ def unregisterProducer(self):
...
+
     def stop(self):
         """Stop creating the stream. Create the terminating zero-length blob."""
         log.debug("stop has been called for StreamCreator")
         self.stopped = True
         if self.current_blob is not None:
-            current_blob = self.current_blob
-            d = current_blob.close()
-            d.addCallback(self._blob_finished)
-            d.addErrback(self._error)
```
jackrobison: Should this errback still be used?

kaykurokawa: no and should be removed

PR #957, opened 2017-10-16
Bump jsonschema version to 2.6.0

kaykurokawa:
latest jsonschema version is 2.6.0 https://pypi.python.org/pypi/jsonschema, 
Chanelog: https://github.com/Julian/jsonschema/blob/master/CHANGELOG.rst

Also for lbryschema and lbryum: 

https://github.com/lbryio/lbryschema/pull/30
https://github.com/lbryio/lbryum/pull/163

Note: the travis integration test here fails because it tries to start up the daemon but can't because lbryschema and lbryum has set requirements to tagged jsonschema version 2.5.1 , once above two is merged, test should pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ at anytime.
   *
 
 ### Changed
-  *
+  * Bumped jsonschema requirement to 2.6.0
   *
 
 ### Added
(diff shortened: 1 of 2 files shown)


jackrobison:
LGTM

PR #960, opened 2017-10-18
add wallet_prefill_addresses command

lyoshenka:
This is very useful when you're publishing multiple files at once (e.g. youtube sync)

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -626,7 +626,7 @@ class Wallet(object):
         @param reserved_points: ReservedPoints object previously returned by reserve_points
 
         @param amount: amount of points to actually send. must be less than or equal to the
-            amount reselved in reserved_points
+            amount reserved in reserved_points
 
         @return: Deferred which fires when the payment has been scheduled
         """
@@ -1321,6 +1321,23 @@ class LBRYumWallet(Wallet):
         else:
             return Decimal((float(c) + float(u) + float(x)) / COIN)
 
+    @defer.inlineCallbacks
+    def create_addresses_with_balance(self, num_addresses, amount, broadcast=True):
+        addresses = self.wallet.get_unused_addresses(account=None)
+        if len(addresses) > num_addresses:
+            addresses = addresses[:num_addresses]
+        elif len(addresses) < num_addresses:
+            for i in range(len(addresses), num_addresses):
+                address = self.wallet.create_new_address(account=None)
+                addresses.append(address)
+            yield self._save_wallet()
+
+        outputs = [[address, amount] for address in addresses]
+        # should this be defer_to_thread or defer_succeed ??
+        tx = yield self._run_cmd_as_defer_to_thread('paytomany', outputs)
+        if broadcast and tx['complete']:
+            yield self._broadcast_transaction(tx)
+        defer.returnValue(tx)
 
     # Return an address with no balance in it, if
     # there is none, create a brand new address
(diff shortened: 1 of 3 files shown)


lyoshenka:
Not sure if I did https://github.com/lbryio/lbry/pull/960/files#diff-eafee006192cca2ffbca5ba91dc83eeeR1336 right

lbrynet/core/Wallet.py
```diff
@@ -1321,6 +1321,23 @@ def get_address_balance(self, address, include_balance=False):
...
+    @defer.inlineCallbacks
+    def create_addresses_with_balance(self, num_addresses, amount, broadcast=True):
+        addresses = self.wallet.get_unused_addresses(account=None)
+        if len(addresses) > num_addresses:
+            addresses = addresses[:num_addresses]
+        elif len(addresses) < num_addresses:
+            for i in range(len(addresses), num_addresses):
+                address = self.wallet.create_new_address(account=None)
+                addresses.append(address)
+            yield self._save_wallet()
```
jackrobison: Can you add a TODO to remove `_save_wallet` from lbrynet? (lbryum should take care of it)

lbrynet/core/Wallet.py
```diff
@@ -1321,6 +1321,23 @@ def get_address_balance(self, address, include_balance=False):
...
+
+        outputs = [[address, amount] for address in addresses]
+        # should this be defer_to_thread or defer_succeed ??
+        tx = yield self._run_cmd_as_defer_to_thread('paytomany', outputs)
+        if broadcast and tx['complete']:
+            yield self._broadcast_transaction(tx)
```
jackrobison: you should keep the result here:
```
if broadcast and tx['complete']:
    txid = yield self._broadcast_transaction(tx)
    tx['txid'] = txid
```

lbrynet/daemon/Daemon.py
```diff
@@ -2376,6 +2376,33 @@ def jsonrpc_wallet_send(self, amount, address=None, claim_id=None):
...
+        """
+        Create new addresses, each containing `amount` credits
+
+        Usage:
+            wallet_prefill_addresses [--no_broadcast]
+                                     (<num_addresses> | --num_addresses=<num_addresses>)
+                                     (<amount> | --amount=<amount>)
+
+        Returns:
+            (bool) true if addresses were created successfully
```
jackrobison: This returns a dict, not a bool.

lyoshenka:
merged in beca640f896e32ce29bd97befb55f0bd774a149a

lbrynet/core/Wallet.py
```diff
@@ -1321,6 +1321,23 @@ def get_address_balance(self, address, include_balance=False):
...
+            addresses = addresses[:num_addresses]
+        elif len(addresses) < num_addresses:
+            for i in range(len(addresses), num_addresses):
+                address = self.wallet.create_new_address(account=None)
+                addresses.append(address)
+            yield self._save_wallet()
+
+        outputs = [[address, amount] for address in addresses]
+        # should this be defer_to_thread or defer_succeed ??
+        tx = yield self._run_cmd_as_defer_to_thread('paytomany', outputs)
```
kaykurokawa: this should be run with _run_cmd_as_defer_succeed

    # run commands as a defer.succeed,
    # lbryum commands should be run this way , unless if the command
    # only makes a lbrum server query, use _run_cmd_as_defer_to_thread()

lbrynet/daemon/Daemon.py
kaykurokawa: probablly should return a bool to be consistent with other wallet_ commands

PR #961, opened 2017-10-23
DHT and hash announcer inlinecallbacks refactor

jackrobison:
* Fixed handling error from dht clients with old ping method
 * Refactor several dht internals to inlineCallbacks
 * Retry announcing a blob up to three times if no peers store the announcement
 * Return storing peer information from the hash announcer after finishing an announce call
 * Convert `DHTHashAnnouncer` and `Node` manage functions to `LoopingCall`s.


lbrynet/core/server/DHTHashAnnouncer.py
```diff
@@ -45,60 +73,102 @@ def immediate_announce(self, blob_hashes):
...
+                log.debug("Stored %s to %i peers (hashes announced by this announcer: %i)",
+                          h.encode('hex')[:16],
+                          len(stored_to_nodes), len(progress))
+
+                yield announce(progress)
             else:
-                self._concurrent_announcers -= 1
+                with self._lock:
+                    self._concurrent_announcers -= 1
+            defer.returnValue(progress)
```
kaykurokawa: We can use a defer.DeferredSemaphore to simplify this and remove managing of self._concurrent_announcers 

See https://github.com/lbryio/lbry/blob/bbc960aaf95569d4cafcbae7e4b1c638973f3797/lbrynet/file_manager/EncryptedFileManager.py#L245 for example

kaykurokawa: Actually I think I need to think about this a bit, there might be a cleaner way of doing this using DeferredQueue's

lbrynet/dht/protocol.py
```diff
@@ -253,8 +253,18 @@ def datagramReceived(self, datagram, address):
                     else:
                         exception_type = UnknownRemoteException
                     remoteException = exception_type(message.response)
-                    log.error("Remote exception (%s): %s", address, remoteException)
-                    df.errback(remoteException)
+                    # this error is returned by nodes that can be contacted but have an old
+                    # and broken version of the ping command, if they return it the node can
+                    # be contacted, so we'll treat it as a successful ping
+                    old_ping_error = "ping() got an unexpected keyword argument '_rpcNodeContact'"
```
kaykurokawa: Put in comments here what version of the daemon has the "old and broken version of the ping command"  (16.4 right?) 
I don't think we want to keep this around forever

kaykurokawa:
All commits except 071e5de  looks fine to me. 

Can you break out  071e5de into a different PR ? I think we need some further work on the DHTHashAnnouncer. I've mostly got it figured out how to remove the excess code involved with using locks and concurrent announcers using DeferredQueues that I can add on top.

Furthermore, it seems like the blob announce time has increased in magnitude and variance with the recent fixes we've made to the DHT, so it seems like we will need a fix to the logic in how the blobs are scheduled to be announced by DHTHashAnnouncer and BlobManager.

kaykurokawa:
Got this error when announcing blobs: 

 
Unhandled error in Deferred:
2017-11-06 13:19:21,946 CRITICAL twisted:154: Unhandled error in Deferred:

2017-11-06 13:19:21,947 CRITICAL twisted:154: 
UnboundLocalError: local variable 'result' referenced before assignment
2017-11-06 13:19:21,950 DEBUG    lbrynet.dht.node:303: Stored f5c3497729548688 to 0 of 0 attempted peers
2017-11-06 13:19:21,950 WARNING  lbrynet.core.server.DHTHashAnnouncer:122: No nodes stored f5c34977295486888eb816dff21d29dc49907f35644f348d723b418d80ea49711058b91e0ba417be4565ed37968f8af5
2017-11-06 13:19:21,951 INFO     lbrynet.core.server.DHTHashAnnouncer:158: announce_to:[Failure instance: Traceback: <type 'exceptions.UnboundLocalError'>: local variable 'result' referenced before assignment
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:498:errback
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:565:_startRunCallbacks
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:651:_runCallbacks
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:1355:gotResult
--- <exception caught here> ---
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:1297:_inlineCallbacks
/usr/local/lib/python2.7/dist-packages/twisted/python/failure.py:389:throwExceptionIntoGenerator
/usr/local/lib/python2.7/dist-packages/lbrynet-0.17.1rc1-py2.7.egg/lbrynet/core/server/DHTHashAnnouncer.py:140:announce
/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py:1297:_inlineCallbacks
/usr/local/lib/python2.7/dist-packages/twisted/python/failure.py:389:throwExceptionIntoGenerator
/usr/local/lib/python2.7/dist-packages/lbrynet-0.17.1rc1-py2.7.egg/lbrynet/core/server/DHTHashAnnouncer.py:131:do_store
]
2017-11-06 13:19:21,952 ERROR    lbrynet.daemon.auth.server:398: Failed to process blob_announce: iteration over non-sequence
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/usr/local/lib/python2.7/dist-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/usr/local/lib/python2.7/dist-packages/lbrynet-0.17.1rc1-py2.7.egg/lbrynet/daemon/Daemon.py", line 2520, in jsonrpc_blob_announce
    yield self.session.blob_manager.immediate_announce_all_blobs()
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1299, in _inlineCallbacks
    result = g.send(result)    
  File "/usr/local/lib/python2.7/dist-packages/lbrynet-0.17.1rc1-py2.7.egg/lbrynet/core/server/DHTHashAnnouncer.py", line 159, in _announce_hashes
    stored_to.update(announced_to)
TypeError: iteration over non-sequence

kaykurokawa:
I  am rebasing , fixing the issue I described above, and make some additional refactoring

kaykurokawa:
df9b4e8 - it is required to adjust the blob announce duration, in order to keep the announce queue from piling up with blobs 

30d96bf - return the blob announce duration in API command status with dht_status flag for monitoring purposes

23d1a53 - minor refactor to fix the fact we don't ever have more than one suppliers

f2b5d91 - cleaner to keep track of retries within the function than setting up a unbounded dictionary

ea5df14 - Since we don't use call later's in DHTHashAnnouncer, we can remove the clock from unit tests

9fca71c - Fix the exception I reported above 


I think this patch should be tested on a reflector host and the blob announces monitored to see if its working. 

This patch does not fix the fact that the DTHHashAnnouncer still needs structural refactoring to remove the maintenance of lock and concurrent announcers. Current major problems of this class:the number of concurrent announcers is not bounded because every time you call _announce_hashes() , it launches X amount of concurrent announcers.

kaykurokawa:
Seems to be working fine on "cluster-host-new2" after several days. Announce time per blob is 6.9 seconds and the announce queue is not piling up.

kaykurokawa:
I'm actually having problems finding blobs announced by "cluster-host-new2" with this PR. Investigation is needed.

kaykurokawa:
Seeing a lot of these types of messages: 

2017-11-15 18:16:02,507 DEBUG    lbrynet.dht.node:264: Timeout while storing blob_hash dfda24be0cdd01cf at 
2017-11-15 18:16:03,189 DEBUG    lbrynet.dht.node:302: Stored e2c1c7b17e3f2dbb to 0 of 6 attempted peers
2017-11-15 18:16:03,190 DEBUG    lbrynet.core.server.DHTHashAnnouncer:141: Stored 6532633163376231 to 6 peers (hashes announced by this announcer:

lbrynet/dht/protocol.py
ordex: just a thought here: what happens if a rpc method is invoked on a node that does not implement that method at all?

Imagine in 0.21 a new 'XXX' command gets implemented.
Then a 0.21 node tries to call 'XXX' on an older one. What will the latter say? 
Will it send an error message or will the node ignore it gracefully? Will the caller node properly handle the error?

(sorry if this problem was already discussed/solved elsewhere, but it just popped up while reading this chunk)

jackrobison: It'll raise an attribute error and return it as an error response (which will re-raise the attribute error for the calling client)

PR #962, opened 2017-10-23
Persist DHT node datastore and restore contacts on startup

jackrobison:


lbrynet/dht/node.py
```diff
@@ -51,7 +52,7 @@ class Node(object):
 
     def __init__(self, node_id=None, udpPort=4000, dataStore=None,
                  routingTableClass=None, networkProtocol=None,
-                 externalIP=None):
+                 externalIP=None, dataDirectory=None, fileName=None):
         """
         @param dataStore: The data store to use. This must be class inheriting
                           from the C{DataStore} interface (or providing the
```
kaykurokawa: need better docstring for dataDirectory (describe what it does when None, etc..)  , need docstring for fileName

lbrynet/dht/datastore.py
```diff
@@ -13,11 +22,22 @@ def __init__(self):
         # Dictionary format:
         # { <key>: (<value>, <lastPublished>, <originallyPublished> <originalPublisherID>) }
         self._dict = {}
+        self._node_state = {}
+
```
kaykurokawa: add docstring for self._node_state

lbrynet/dht/datastore.py
```diff
@@ -46,7 +66,64 @@ def getPeersForBlob(self, key):
...
+    def Save(self, contacts):
+        pass
+
+
+class JSONFileDataStore(DictDataStore):
+    def __init__(self, node, path, file_name=None):
+        DictDataStore.__init__(self)
+        self._node = node
+        self._path = os.path.join(path, file_name or "dht_datastore.json")
```
kaykurokawa: Doesn't look like JSONFileDataStore uses self._node
If it does, you have a circular dependency

tzarebczan:
@jackrobison will this PR be revived or can it be closed?

Issue #892, opened 2017-09-07
Redundant API servers for currency conversion


tzarebczan:
## The Issue
Earlier, we ran into this issue: https://github.com/lbryio/lbry/issues/890 and in the past, we also had some currency conversion problems with the Bittrex API.  If LBRY cannot properly convert currencies, the app may not allow paid content to be purchased or it may lead to using cached conversions.  LBRY should support failover API servers in case the primary is not working properly in an effort to increase redundancy. 

Code for Exchange rate manager: https://github.com/lbryio/lbry/blob/aef31991f9bf2751d5d980b0a2074caf19af1b14/lbrynet/daemon/ExchangeRateManager.py

### Expected behaviour
if one API call fails, there should be at least one redundant backup

### Actual behaviour
if API is down, users cannot purchase paid content

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
Periodically I am seeing `2017-09-15 11:26:14,750 WARNING  lbrynet.daemon.ExchangeRateManager:69: There was a problem updating USDBTC exchange rate information from lbry.io` in my daemon, but don't think its breaking anything.

PR #915, opened 2017-09-21
Added redundant market feed (cryptonator).

mirgee:
Addressing #892.

PR #963, opened 2017-10-23
Added redundant market feed (continued)

kaykurokawa:
Reviewed and made some changes on top of https://github.com/lbryio/lbry/pull/915 by mirgee , which adds ExchangeRateManager capability to have fail over feeds so if one feed fails it can use another. Cryptonator was added as a back up feed.

I added also BittrexFeed as a back up feed, and re-added the unit test for it which was removed some patches ago. 

I also adjusted the detection of if a feed is detected to be online or not. mirgee set this variable in the _make_request function when requests.get() returned a response with status code that is not a success. However this doesn't handle the case when requests.get() throws an exception due to the end point not being available, or if parsing of the requests fail because they changed the API. We now set a feed to be offline in the errback so any exception while trying to get the feed will set the feed to be offline.

fixes https://github.com/lbryio/lbry/issues/892

### DIFF
--- a/lbrynet/daemon/ExchangeRateManager.py
+++ b/lbrynet/daemon/ExchangeRateManager.py
@@ -71,7 +71,7 @@ class MarketFeed(object):
         return defer.succeed(from_amount / (1.0 - self.fee))
 
     def _save_price(self, price):
-        log.debug("Saving price update %f for %s" % (price, self.market))
+        log.debug("Saving price update %f for %s from %s" % (price, self.market, self.name))
         self.rate = ExchangeRate(self.market, price, int(time.time()))
 
     def _log_error(self, err):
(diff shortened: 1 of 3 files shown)


lbrynet/daemon/ExchangeRateManager.py
```diff
@@ -43,11 +42,16 @@ def __init__(self, market, name, url, params, fee):
...
         self.rate = None
         self._updater = LoopingCall(self._update_price)
+        self._online = True
 
     @property
     def rate_is_initialized(self):
         return self.rate is not None
 
+    @property
+    def is_online(self):
```
lyoshenka: why not make this a function? whats the benefit of hiding the fact that a function is being called? i see this `@property` stuff everywhere in python, but i dont understand it at all

kaykurokawa: i don't see a good reason for using @property in either is_online() or rate_is_initialized() so I will remove them

kaykurokawa:
Removed @property from is_online() and rate_is_initialized()
Rebased to master

lyoshenka:
@kaykurokawa looks good to merge?

kaykurokawa:
Rebased to master and merging

PR #965, opened 2017-10-23
DHT connectivity checks

jackrobison:
* Added `pingback` dht method to test connectivity
  * Filter dht queries from unreachable peers
  * Check dht peers requesting to store can be contacted before accepting
  * Use `pingback` to check connectivity on startup, don't try to send `store` queries if this fails
  * Improved logging of dht errors and non-typical behavior


jackrobison:
Closing this since the merged dht changes better accomplish the goal of making sure all peers in the routing table are reachable and fresh.

PR #966, opened 2017-10-23
add peer_ping, improve peer_list and blob_announce

jackrobison:
* Added `peer_ping` daemon command
  * Addded `blob_hashes` parameter to `blob_announce`
* Return announce info dictionary of `{<blob_hash>: [<node_id>, ...]}` results from `blob_announce`
  * Return a list of dictionaries from `peer_list`, add `node_id` to peer info


lbrynet/conf.py
```diff
@@ -41,6 +41,7 @@
 MB = 2 ** 20
 
 DEFAULT_DHT_NODES = [
+    ('jack.lbry.tech', 4444),
```
lyoshenka: dont forget to remove

lbrynet/dht/distance.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Frequently we re-use one of the points so as an optimization
+    we pre-calculate the long value of that point.
+    """
+
+    def __init__(self, key):
+        self.key = key
+        self.val_key_one = long(key.encode('hex'), 16)
+
+    def __call__(self, key_two):
```
lyoshenka: this feels like magic

PR #968, opened 2017-10-25
Don't filter out local node ID when returning peer list

ordex:
If a node is returning a peer list for a given blob hash
(being this been requested via CLI or via DHT) and it is
part of the resulting peer list, it will filter itself out
before returning the list.

This makes the results across the DHT inconsistent as
different nodes won't include themselves when
responding a findValue/findNode query.

Remove such filtering so that the local node ID is always
included when needed.

### DIFF
--- a/lbrynet/dht/node.py
+++ b/lbrynet/dht/node.py
@@ -216,14 +216,13 @@ class Node(object):
         if result:
             if blob_hash in result:
                 for peer in result[blob_hash]:
-                    if self.node_id != peer[6:]:
-                        host = ".".join([str(ord(d)) for d in peer[:4]])
-                        if host == "127.0.0.1" and "from_peer" in result \
-                                and result["from_peer"] != "self":
-                            host = result["from_peer"]
-                        port, = struct.unpack('>H', peer[4:6])
-                        if (host, port) not in expanded_peers:
-                            expanded_peers.append((host, port))
+                    host = ".".join([str(ord(d)) for d in peer[:4]])
+                    if host == "127.0.0.1" and "from_peer" in result \
+                            and result["from_peer"] != "self":
+                        host = result["from_peer"]
+                    port, = struct.unpack('>H', peer[4:6])
+                    if (host, port) not in expanded_peers:
+                        expanded_peers.append((host, port))
         defer.returnValue(expanded_peers)
 
     def get_most_popular_hashes(self, num_to_return):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ at anytime.
   * Fixed lbryid length validation
   * Fixed an old print statement that polluted logs
   * Fixed rpc id length for dht requests
+  * Fixed local node ID reporting in peer list
 
 ### Deprecated
   *


kaykurokawa:
I think this change makes sense. 

But changes need to be made to the lbrynet classes using getPeersForBlob() to handle cases where it receives itself as a blob source. Antonio, either jack or I will handle that part since we are more familiar with that part. 

Also this would be for a new PR but we should clearly have functions for parsing and serializing the value of the key,value pair stored in the DHT.

ordex:
> But changes need to be made to the lbrynet classes using getPeersForBlob() to handle cases where it receives itself as a blob source, otherwise it may try to download blobs from itself.

@kaykurokawa how do we currently handle the case of several nodes behind the same NAT? Are peers able to distinguish this case? I believe that only the external IP is announced, right? If so, a node won't be able to understand if the peer list contains itself or another node behind the same router (thus sharing the public IP)

jackrobison:
@ordex 

>how do we currently handle the case of several nodes behind the same NAT?

With multiple external ports. A "contact triple" is a (<node id>, <ip address>, <port>) tuple.

tzarebczan:
A couple versions ago I was able to download from other PCs on my network, but I haven't seen that happening since the recent DHT changes. For instance, I'm having trouble finding a peer for a blob right now, announced it from other PC, still can't find it. A 3rd PC was able to download fine from other hosts....please make these issues go away :p

tzarebczan:
Also, now that I think about it, this should be a setting. I'd like to be able to test content availability/downloads from multiple machines at home. If I download from other peers behind my NAT, it's not a good test.

lyoshenka:
@ordex i removed jack and kay as reviewers for now, since it looks like they left comments above. please add them when you're ready for them to look at this again.

lbrynet/dht/node.py
```diff
@@ -216,14 +216,13 @@ def getPeersForBlob(self, blob_hash):
...
-                    if self.node_id != peer[6:]:
-                        host = ".".join([str(ord(d)) for d in peer[:4]])
-                        if host == "127.0.0.1" and "from_peer" in result \
-                                and result["from_peer"] != "self":
-                            host = result["from_peer"]
-                        port, = struct.unpack('>H', peer[4:6])
-                        if (host, port) not in expanded_peers:
-                            expanded_peers.append((host, port))
+                    host = ".".join([str(ord(d)) for d in peer[:4]])
+                    if host == "127.0.0.1" and "from_peer" in result \
```
kaykurokawa: what is this if statement doing here?

ordex: I think it means: if this peer_list has been received from node X (where X != me) and I have '127.0.0.1' in the list, then substitute 127.0.0.1 address with X's address. Basically the code assumes that X had written 127.0.0.1 in the peer_list to represent itself, but to 'me' 127.0.0.1 does not make sense, therefore I use the real X's address.

To my understanding nobody is expected to store 127.0.0.1 in the DHT, therefore this should never happen. My guess is that this chunk was here to work around some bug.

does it make sense?

kauffj: Isn't all of `127.x.x.x` reserved for loopback (or other internal usage)? 

So wouldn't it make sense to prevent all 127 addresses from ever being stored in the DHT?

ordex: Right. Theoretically all private addresses could be prevented from being stored in the DHT (127.0.0.0/8, 10.0.0.0/8, 172.16.0.0/12 and 192.168.0.0/16). Although there might be some cases where announcing private addresses may make sense, I believe it does not when nodes are connecting to the "global"/"public" DHT.

Therefore it should be safe to reject the store command and print some warning.

This is probably worth another patch.

kaykurokawa:
Will continue this here : https://github.com/lbryio/lbry/pull/975

PR #969, opened 2017-10-25
Extend close nodes list by right amount

ordex:
When a k-bucket does not contain enough close nodes, the
DHT will look into neighbouring k-buckets in order to entend
the list of returned nodes.

However, the list should not be extended beyond its maximum size.


jackrobison:
Added https://github.com/lbryio/lbry/commit/e89f3db31bb576dd3915fece38434279c8e2037b

This makes sure the result has at most `min(count, k)` items. Otherwise the length-checking logic could result in extension of the results for both the lower and higher buckets, whether or not enough results had already been found.

PR #972, opened 2017-10-26
change stream_cost_estimate "name" param to "uri"

treethought:
When making a request with the stream_cost_estimate method, an error is thrown if "name" is provided instead of "uri" as a parameter.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -633,7 +633,7 @@ Returns:
 Get estimated cost for a lbry stream
 
 Args:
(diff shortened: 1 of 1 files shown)


kaykurokawa:
That doc is autogenerated by https://github.com/lbryio/lbry/blob/master/scripts/gen_api_docs.py but looks like the script never got run on release?

kauffj:
**grumbles**

lyoshenka:
updated

PR #973, opened 2017-10-30
Session: simplify UPnP setup and avoid race condition

ordex:
Checking for a mapping availability and setting it up in two
different moments may lead to race conditions as other procsses may
register the same mapping before lbrynet does.

Moreover, the return value of upnp.addportmapping() was not checked,
thus leading lbrynet to believe that the mapping was succesful even if
it was not.

Starting with miniupnpc 2.0.x, we have support for the
upnp.addanyportmapping() primitive which instructs the server to reserve
the first available external port if the requested one is already taken.

If no port is avaialable the upnp call will fail.

This new primitive therefore avoids the need to "test and set" the
mapping thus eliminating the racy behaviour.

Simplify the UPnP setup by using upnp.addanyportmapping() and check
its return value to ensure everything went fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ at anytime.
   *
 
 ### Fixed
-  *
+  * Fixed race condition during UPnP setup between 'check' and 'set' of the new mapping
   *
 
 ### Deprecated
@@ -25,7 +25,8 @@ at anytime.
   *
 
 ### Added
-  *
+  * Added error detection in UPnP setup
+  * Updated miniupnpc from 1.9 to 2.0.2 (provides upnp.addanyportmapping)
   *
 
 ### Removed
--- a/lbrynet/core/Session.py
+++ b/lbrynet/core/Session.py
@@ -193,27 +193,33 @@ class Session(object):
 
         log.debug("In _try_upnp")
 
-        def get_free_port(upnp, port, protocol):
-            # returns an existing mapping if it exists
+        def get_mapped_port(upnp, port, protocol):
+            # return the external port mapped to this host:port/protocol, if any
             mapping = upnp.getspecificportmapping(port, protocol)
-            if not mapping:
-                return port
-            if upnp.lanaddr == mapping[0]:
-                return mapping
-            return get_free_port(upnp, port + 1, protocol)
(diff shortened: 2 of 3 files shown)


lyoshenka:
How does this interact with the port setting in the config file? If the user selected a port to use, it should always use that port and error if that port is not available. A lot of our servers have firewalls that block all ports except those that we allow, so we need to listen on one port and one port only.

Here's my proposal for the overall behavior:
- port defaults to being unset in the configuration
- if port is not set, bind any available port. error if no port could be bound
- if port is set, bind only that port. error if it cannot be bound

In addition, it would be nice to test if the bound port is actually accessible externally. Maybe we have a server set up that does @jackrobison's ping-handshake thing and returns whether it worked? Then we can warn (or even error) if it failed.

kauffj:
@lyoshenka it's preferable if we can work out a method to test external accessibility without a centralized server.

lyoshenka:
Sure. The ping handshake jack made can be done with any server. So the daemon can do it with the bootstrap nodes, or with any node at all. Just have to make sure this isn't a ddos vector (asking every node to ping a single node).

ordex:
@lyoshenka thanks for your comments!

> How does this interact with the port setting in the config file?

This interaction is the same as it was before this patch, because none of this logic is being altered.

> If the user selected a port to use, it should always use that port and error if that port is not available. A  lot of our servers have firewalls that block all ports except those that we allow, so we need to listen on one port and one port only.

Yeah, I totally agree here.
Should we better open a ticket and work on that in a separate PR, so to keep this change as it is?
Or do prefer to kill this PR and directly implement the suggested logic?

ordex:
Just checking the surrounding code: it seems that a node will still announce its originally configured port, even though the upnp setup has selected a different one.

ordex:
At this point I suggest we drop this PR. I'll then come up with something that takes care of all the comments above.

ordex:
If I am not wrong, the DHT protocol already does some kind of self lookup right after boot in order to populate its routing table (and to inject itself in the routing table of the others). If for a given a amount of time after startup we don't hear from anybody else with a public IP, then we know we are isolated.

I think this might be a simpler approach as we would not need to implement a new method.

lyoshenka:
@ordex sounds good

PR #975, opened 2017-10-30
Don't filter out local node ID when returning peer list (continued) 

kaykurokawa:
Continuing from https://github.com/lbryio/lbry/pull/968

Removed the "if" statement in dht.node.getPeersForBlob(), where it did some strange workaround for filtering localhosts from the results. 

I added argument "filter_self" to DHTPeerFinder.find_peers_for_blob() where it will filter itself from peer results if True, this option is needed when downloading blobs since you don't want to download blobs from yourself. In other use cases, such as API command peer_list, filter_self is defaulted to False so you will be able to find yourself. Before https://github.com/lbryio/lbry/pull/968, the filtering of itself was done in the DHT code and would actually affect availability. Now it is done purely outside of the DHT code so it has no impact on availability.

In order for DHTPeerFinder to know whether a peer was itself or not, I needed to add the "peerPort" as member variable that is initialized by the dht.node class. Thus when announcing blobs using dht.node.announceHaveBlob(), port is no longer needed as an argument. I think this is a more logical way of doing it since externalIP is already a member variable of the dht.node class and there is no use case where we need to announce blobs on different ports.

### DIFF
--- a/lbrynet/core/client/DHTPeerFinder.py
+++ b/lbrynet/core/client/DHTPeerFinder.py
@@ -15,6 +15,10 @@ class DHTPeerFinder(object):
     implements(IPeerFinder)
 
     def __init__(self, dht_node, peer_manager):
+        """
+        dht_node - an instance of dht.Node class
+        peer_manager - an instance of PeerManager class
+        """
         self.dht_node = dht_node
         self.peer_manager = peer_manager
         self.peers = []
@@ -34,7 +38,17 @@ class DHTPeerFinder(object):
         pass
 
     @defer.inlineCallbacks
-    def find_peers_for_blob(self, blob_hash, timeout=None):
+    def find_peers_for_blob(self, blob_hash, timeout=None, filter_self=False):
+        """
+        Find peers for blob in the DHT
+        blob_hash (str): blob hash to look for
+        timeout (int): seconds to timeout after
+        filter_self (bool): if True, and if a peer for a blob is itself, filter it
+                from the result
+
+        Returns:
+        list of peers for the blob
+        """
         def _trigger_timeout():
             if not finished_deferred.called:
                 log.debug("Peer search for %s timed out", short_hash(blob_hash))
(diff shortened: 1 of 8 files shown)


lbrynet/dht/node.py
```diff
@@ -216,14 +219,10 @@ def getPeersForBlob(self, blob_hash):
         if result:
             if blob_hash in result:
                 for peer in result[blob_hash]:
-                    if self.node_id != peer[6:]:
-                        host = ".".join([str(ord(d)) for d in peer[:4]])
-                        if host == "127.0.0.1" and "from_peer" in result \
-                                and result["from_peer"] != "self":
```
jackrobison: This is coming from https://github.com/lbryio/lbry/blob/0b771c16ba058d3919b2bac482f4d725473a72fa/lbrynet/dht/node.py#L369

kaykurokawa: see discussion at end here: #968

kaykurokawa:
After discussion with jack , it looks like his "from_peer" field never gets used except in the if statement we just removed. I added a commit 90060e9 that completely removes the addition of this "from_peer" field. 

Also rebased to master.

lbrynet/core/client/DHTPeerFinder.py
```diff
@@ -34,7 +38,17 @@ def _manage_peers(self):
...
 
     @defer.inlineCallbacks
-    def find_peers_for_blob(self, blob_hash, timeout=None):
+    def find_peers_for_blob(self, blob_hash, timeout=None, filter_self=False):
+        """
+        Find peers for blob in the DHT
+        blob_hash (str): blob hash to look for
+        timeout (int): seconds to timeout after
+        filter_self (bool): if True, and if a peer for a blob is itself, filter it
+                from the result
```
ordex: (assume that I have not dug yet in the downloader code) I am wondering if knowing that `self` already has the blob could be a useful information to avoid downloading it again. Or is this case handled differently by the downloader (i.e. the downloader checks the blob directory directly before issuing a request)?

kaykurokawa: There is a BlobManager class which knows all the blobs  it has. In theory , when the daemon is downloading it should not try to download blobs it has announced. But I think in practice, this can happen if for example you announce a blob, and than you delete the blob in blob manager and try to download the same blob again.

kaykurokawa:
Rebased to master, 
this is LGTM

lbrynet/core/client/DHTPeerFinder.py
ordex: OK, it makes sense. Thanks for clarifying!

Issue #971, opened 2017-10-25
Network connectivity broken on dual-stack hosts


ordex:
## The Issue
On a dual-stack host lbrynet-daemon may pick up the local IPv6 as address to listen on. When this happens it will try to use upnp to create a mapping between the external IPv4 and the internal IPv6. Obviously this can't work. The result is that lbrynet-daemon won't be able to connect to any other peer.

When it happens, it is possible to see something similar in the console log:

>2017-10-25 18:54:35,965 INFO     lbrynet.core.Session:222: Set UPnP redirect 2001:470:19:1:7f75:dd68:ae8:f65e:3333 (TCP) to 192.168.1.192:3333
>2017-10-25 18:54:35,980 INFO     lbrynet.core.Session:222: Set UPnP redirect 2001:470:19:1:7f75:dd68:ae8:f65e:4444 (UDP) to 192.168.1.192:4444

Subsequently, normal dht operations will result in the following message:

>2017-10-25 18:51:16,719 WARNING  lbrynet.dht.node:596: This node doesnt know any other nodes

or a crash in twisted:

>Unhandled error in Deferred:
>2017-10-25 18:55:08,838 CRITICAL twisted:154: Unhandled error in Deferred:

>2017-10-25 18:55:08,838 CRITICAL twisted:154: 

> Traceback (most recent call last):
>   File "/home/ordex/exp/lbry/lbry-venv/lib/python2.7/site-packages/twisted/internet/defer.py", line 651, in _runCallbacks
>     current.result = callback(current.result, *args, **kw)
>   File "/home/ordex/exp/lbry/lbrynet/dht/node.py", line 287, in requestPeers
>     self.store(blob_hash, value, self_store=True, originalPublisherID=self.node_id)
>   File "/home/ordex/exp/lbry/lbrynet/dht/node.py", line 472, in store
>     compact_ip = contact.compact_ip()
>   File "/home/ordex/exp/lbry/lbrynet/dht/contact.py", line 33, in compact_ip
>     lambda buff, x: buff + bytearray([int(x)]), self.address.split('.'), bytearray())
>   File "/home/ordex/exp/lbry/lbrynet/dht/contact.py", line 33, in <lambda>
>     lambda buff, x: buff + bytearray([int(x)]), self.address.split('.'), bytearray())
> ValueError: invalid literal for int() with base 10: '2001:470:19:1:7f75:dd68:ae8:f65e'`
> 

### Steps to reproduce
[...]

PR #976, opened 2017-10-31
platform ip: ensure daemon retrieves IPv4 address only

ordex:
When contacting jsonip.com to retrieve the node external IP,
the connection might be established with IPv6 and thus return
an address belonging to that family.

This address is then used to initialize the external_ip member of
Daemon session, but unfortunately IPv6 is not yet handled well.

Using an IPv6 as external IP is currently breaking parts of the
Daemon resulting in no peer connectivity at all.

We should stick to IPv4 for time being.

http://jsonip.com/about says "ipv4-only":"https://ipv4.jsonip.com"

therefore, change the IP retrieval URL to ipv4.jsonip.com to
make sure the connection is established only using an IPv4 address.

Closes: https://github.com/lbryio/lbry/issues/971

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ at anytime.
   *
 
 ### Fixed
-  *
+  * Fixed external IP detection via jsonip.com (avoid detecting IPv6)
   *
 
 ### Deprecated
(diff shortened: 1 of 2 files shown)


lyoshenka:
Merged in d827b4d5af46b552395a2673f8c86d07958cc6de

PR #979, opened 2017-11-01
add wallet_list_unspent

lyoshenka:


lbrynet/daemon/Daemon.py
```diff
@@ -2404,6 +2404,42 @@ def jsonrpc_wallet_prefill_addresses(self, num_addresses, amount, no_broadcast=F
...
+                    "nout": (int) nout of the output
+                    "txid": (str) txid of the output
+                },
+                ...
+            ]
+        """
+
+        unspent = yield self.session.wallet.list_unspent()
+        for i, utxo in enumerate(unspent):
+            utxo['txid'] = utxo.pop('prevout_hash')
```
jackrobison: Please file an issue for these poorly named fields in lbryum - lbrynet shouldn't have to work around it.

Issue #951, opened 2017-10-12
Cannot stream/download content that was updated (if it was previously downloaded)


tzarebczan:
## The Issue
Today I noticed I'm having trouble streaming lbry://one - the app didn't recognize it as a previously downloaded file, but I know I had it in the past and it shows up in my Downloads section. If I try to stream it, it also fails - just get's stuck on the loading icon after geting a completed notification and the daemon shows - `lbrynet.daemon.Daemon:1571: Already have a file for one`. If I click Download, it shows connecting and then goes back to showing the download button - the daemon shows the same message about having the file. This claim was recently updated by Oakie so I checked the file_list and noticed that the outpoint is incorrect on the broken install and shows correctly on working install (not sure how one updated but not the other).  I'm hoping the streaming issue is also related to an incorrect endpoint, but not sure.  Both installs had the file downloaded well before the outpoint was changed. 

So as a more general question, can we check the functionality of file_list and when/how it is updated? 

Update: Now I'm not even sure how the laptop received the updated endpoint. My current testing shows this will happen anytime a claim is updated and the filename remains the same. This causes a mess for anyone that downloaded the content but then the claim is updated - cannot download the new content without deleting the file manually (using file_delete, not sure about deleting the file in the downloads directory yet - haven't tested). 
main PC (broken)
```
  {
    "claim_id": "008401a48cecef2a13cf5376b696169bb36b18c6",
    "completed": true,
    "download_directory": "D:\\LBRY Downloads Main PC",
    "download_path": "D:\\LBRY Downloads Main PC\\Happy Working Song - Enchanted - Holly Oakey-lZ5ZL0VSlC4.mp4",
    "file_name": "Happy Working Song - Enchanted - Holly Oakey-lZ5ZL0VSlC4.mp4",
    "has_signature": false,
    "key": "abf7fb30e4bb9d683ffa635617ee0974",
    "message": null,
    "metadata": {
      "claimType": "streamType",
      "stream": {
        "metadata": {
          "author": "",
          "description": "All proceeds go to holly for buying toys, i will post the video with those toys on Xmas day",
          "language": "en",
          "license": "All rights reserved.",
          "licenseUrl": "",
          "nsfw": false,
          "preview": "",
          "thumbnail": "http://www.thetoydiscounter.com/happy.jpg",
          "title": "Holly singing The Happy Working Song",
          "version": "_0_1_0"
        },
        "source": {
          "contentType": "video/mp4",
[...]

Issue #981, opened 2017-11-02
Do not associate claim information with files 


kaykurokawa:
When calling API command file_list , various claim information is attached onto the file from the wallet

see Daemon._get_lbry_file_dict() : https://github.com/lbryio/lbry/blob/master/lbrynet/daemon/Daemon.py#L863

and ManagedEncryptedFileDownloader.load_file_attributes()
https://github.com/lbryio/lbry/blob/master/lbrynet/file_manager/EncryptedFileDownloader.py#L111

This is wrong, because the claim information at the time of download can be completely different from the claim information at the time of checking the claim information. Resulting in confusion and bugs when claims get updated and abandoned. Further more, files can and do exist without having any claim information attached to them at all (for example , files that have been uploaded to reflector. Calling file_list on reflector takes forever because it has to fetch all the claim information from lbryum-server since it doesn't have the information in the wallet cache). 

This has resulted in numerous bugs in the app side and daemon side, when claims get abandoned and updated ( see:  https://github.com/lbryio/lbry/issues/944, https://github.com/lbryio/lbry-app/issues/694 ) 

We need to completely remove claim information from the file_list and also file_delete API command. Below is the exact API change that will happen: 

Individual files in the output of API command file_list will no longer contain fields "name", "outpoint", "claim_id", "metadata", "channel_name", "has_signature", "signature_is_valid" 

 "claim_id", "outpoint", "name" is no longer a valid argument to search for files for API command file_list, file_delete, and file_reflect

kauffj:
:+1: from me. This will be relatively trivial to adjust app-side.

jackrobison:
merged here https://github.com/lbryio/lbry/commit/188822ef201d38b511472d01a1a7e370029e070a

PR #983, opened 2017-11-03
Remove claim information from files

kaykurokawa:
This removes claim information attached to files, and implements the fix for https://github.com/lbryio/lbry/issues/981

9502b29 - remove the claim information from file related API daemon commands

4aa3cf7 - reflector would use lbry_file.name, for logging purposes, use lbry_file.file_name

1527304 - fix test for above change

d17d410 -when calling API command get, it would check to see if it had the file by looking up the claim_id. Instead, look up by the sd_hash, also fixes https://github.com/lbryio/lbry/issues/951

1bd620a - when calling API command get, it would check to see if there was an ongoing download by looking up the claim_id, also look up by sd_hash, also fixes https://github.com/lbryio/lbry/issues/951

I assigned @kauffj as a reviewer here, he does not have to review this but he needs to give the OK to merge when the app-side work is complete.

NOTE: This PR changes the API  (should  I run the api script in the PR?) , and will break the app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ at anytime.
 ### Fixed
   * Fixed unnecessarily verbose exchange rate error (https://github.com/lbryio/lbry/issues/984)
   * Merged two sepereate dht test folders into one
-  *
 
 ### Deprecated
   * `channel_list_mine`, replaced with `channel_list`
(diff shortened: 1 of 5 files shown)


kauffj:
@kaykurokawa you can merge this at any time. Even if app changes aren't done yet, it just means the app won't bump the daemon version until they are.

Issue #511, opened 2017-03-06
Connection manager unit test fails


kaykurokawa:
test_parallel_connection() on connection manager unit test fails occasionally in a random manner : https://github.com/lbryio/lbry/blob/master/tests/unit/core/client/test_ConnectionManager.py#L192

tzarebczan:
Fixed by https://github.com/lbryio/lbry/pull/987

PR #987, opened 2017-11-04
Fix failing ConnectionManager unit test

ian-h-chamberlain:
Addresses #511. Previously, the test yielded to obtain the result of `connection_was_made`, but the assertions were made in sequential order. Sometimes the result of both connections would be returned before resuming the test, failing the assertions. Instead, this adds a callback to verify the result of each connection, then yields to get both Deferred results.


tzarebczan:
@ian-h-chamberlain, thanks for the contribution! Please check out https://lbry.io/faq/tips

Issue #984, opened 2017-11-03
Cryptonator market feed throwing errors


tzarebczan:
## The Issue
See below for error from cryptonator feed added recently:

```
2017-11-03 15:17:01,342 WARNING  lbrynet.daemon.ExchangeRateManager:74: There was a problem updating BTCLBC exchange rate information from cryptonator.com: [Failure instance: Traceback: <class 'lbrynet.core.Error.InvalidExchangeRateResponse'>: Failed to get exchange rate from cryptonator.com:invalid rate response : <!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
  <meta http-equiv="refresh" content="10;URL=/ciez2a">
</head>
<body>
  <script type="text/javascript">
    var head    = document.getElementsByTagName("head")[0];
    var script  = document.createElement("script");
    var done    = false; // Handle Script loading
                
    var url     = "https://thae3.variti.de/share/1HZCMx3ZlCg1?sid=391&scheme=https&host=api.cryptonator.com&uri=%2fapi%2fticker%2fbtc-lbc%3futm_referrer%3d&t=1509736621390&sad=v%2fGLhqBQ%3d%3d&uid=hXtm3GM5hxaovnpU&uct=1509736621389&kct=0&m=3&ver=5&jssc=0&v=4cHYl8s5_aEO1fUPg6Y-YA";
    if (window.callPhantom || window._phantom || window.__phantomas) {
      url = url + "&test=1HZCMx";
    } else {
      url = url + "&test=3ZlCg1";
    }
    script.src = url;
    script.onload = script.onreadystatechange = function() { // Attach handlers for all browsers
      if ( !done && (!this.readyState || this.readyState === "loaded" || this.readyState === "complete") ) {
        done = true;
        ipp.setCookie();
        location.href="https://api.cryptonator.com/api/ticker/btc-lbc?utm_referrer=";
        script.onload = script.onreadystatechange = null; // Handle memory leak in IE
      }
    };
    head.appendChild(script);           
  </script>
</body>
</html>


site-packages\twisted\internet\base.py:1208:mainLoop
site-packages\twisted\internet\base.py:801:runUntilCurrent
site-packages\twisted\internet\defer.py:457:callback
site-packages\twisted\internet\defer.py:565:_startRunCallbacks
--- <exception caught here> ---
site-packages\twisted\internet\defer.py:651:_runCallbacks
[...]

PR #993, opened 2017-11-10
less verbose exchange rate error

jackrobison:
fixes https://github.com/lbryio/lbry/issues/984


lyoshenka:
@jackrobison this doesnt need a PR. you can make this change right in master.

PR #998, opened 2017-11-15
Add lbry file manager to reflector server

kaykurokawa:
depends on https://github.com/lbryio/lbry/pull/983 . Reflector server now adds file to file manager. Needed to make a few structural changes to make this happen.

ManagedEncryptedFileDownloader had a function load_file_attribute() which is used to load claim information along with the sd_hash. Since #983 removes claim information from files, now all this function does is load the sd_hash. Move this task to the base class EncryptedFileDownloader in function set_stream_info() where it gets stream information (cleaner since we have less setup functions that needs to be run in order to load file information). 

EncryptedFileManager.add_lbry_file() has a mandatory payment_rate_manager argument which is annoying for users to specify (they need to have access to Session class, which EncryptedFileManager has already) Make it optional and use Session.payment_rate_manager.

Changes reflector tests to use file manager and check if files are loaded into file manager.

### DIFF
--- a/lbrynet/file_manager/EncryptedFileDownloader.py
+++ b/lbrynet/file_manager/EncryptedFileDownloader.py
@@ -7,7 +7,6 @@ from zope.interface import implements
 from twisted.internet import defer
 
 from lbrynet.core.client.StreamProgressManager import FullStreamProgressManager
-from lbrynet.core.Error import NoSuchStreamHash
 from lbrynet.core.utils import short_hash
 from lbrynet.core.StreamDescriptor import StreamMetadata
 from lbrynet.lbry_file.client.EncryptedFileDownloader import EncryptedFileSaver
@@ -35,18 +34,21 @@ class ManagedEncryptedFileDownloader(EncryptedFileSaver):
     STATUS_RUNNING = "running"
     STATUS_STOPPED = "stopped"
     STATUS_FINISHED = "finished"
-
+    """
+    These are started by EncryptedFileManager, aka, file_manager
+    """
     def __init__(self, rowid, stream_hash, peer_finder, rate_limiter,
                  blob_manager, stream_info_manager, lbry_file_manager,
                  payment_rate_manager, wallet, download_directory,
                  file_name=None):
+
         EncryptedFileSaver.__init__(self, stream_hash, peer_finder,
                                     rate_limiter, blob_manager,
                                     stream_info_manager,
                                     payment_rate_manager, wallet,
                                     download_directory,
                                     file_name)
-        self.sd_hash = None
+
         self.rowid = rowid
         self.lbry_file_manager = lbry_file_manager
         self._saving_status = False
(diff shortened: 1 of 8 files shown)


lbrynet/file_manager/EncryptedFileDownloader.py
```diff
@@ -35,18 +34,21 @@ class ManagedEncryptedFileDownloader(EncryptedFileSaver):
...
                  payment_rate_manager, wallet, download_directory,
                  file_name=None):
+
         EncryptedFileSaver.__init__(self, stream_hash, peer_finder,
                                     rate_limiter, blob_manager,
                                     stream_info_manager,
                                     payment_rate_manager, wallet,
                                     download_directory,
                                     file_name)
-        self.sd_hash = None
```
jackrobison: The attribute should be set here to `None` just so it's not a new attribute being added later on

lbrynet/reflector/server/server.py
```diff
@@ -107,6 +108,7 @@ def _on_completed_blob(self, blob, response_key):
             yield save_sd_info(self.stream_info_manager, sd_info)
             yield self.stream_info_manager.save_sd_blob_hash_to_stream(sd_info['stream_hash'],
                                                                        blob.blob_hash)
+            yield self.lbry_file_manager.add_lbry_file(sd_info['stream_hash'])
```
jackrobison: This should be non-blocking (otherwise it would block sending the `blob_completed` response on the file being started)

lbrynet/lbry_file/client/EncryptedFileDownloader.py
```diff
@@ -28,21 +29,22 @@ def __init__(self, stream_hash, peer_finder, rate_limiter, blob_manager,
         self.stream_info_manager = stream_info_manager
         self.suggested_file_name = None
         self._calculated_total_bytes = None
+        self.sd_hash = None
```
kaykurokawa: This was moved from ManagedEncryptedFileDownloader in lbrynet/file_manager/EncryptedFileDownloader.py which inherits from this class

lbrynet/file_manager/EncryptedFileDownloader.py
kaykurokawa: This attribute was moved to the class that it inherits from as part of this work
>
ManagedEncryptedFileDownloader had a function load_file_attribute() which is used to load claim information along with the sd_hash. Since #983 removes claim information from files, now all this function does is load the sd_hash. Move this task to the base class EncryptedFileDownloader in function set_stream_info() where it gets stream information (cleaner since we have less setup functions that needs to be run in order to load file information).

kaykurokawa:
>
ManagedEncryptedFileDownloader builds output files, which would double storage requirements for reflector servers

ManagedEncryptedFileDownloader tries to download blobs, reflector server probably shouldn't do that (certainly not immediately after receiving an sd blob at the beginning of a new upload)

I think (not 100% sure so checking on this) that this PR does not actually do this because it keeps the ManagedEncryptedFileDownloader in the "stopped" state (although the user could toggle the state to "running" using API command file_set_status).

kaykurokawa:
Fixed blocking on add_lbry_file() when reflector server is downloading blobs

Added some additional tests in test_reflector.py to check file status (make sure its in stopped state, make sure it has correct number of blobs)

PR #1000, opened 2017-11-16
[WIP] Add test for multiple nodes on localhost for DHT

kaykurokawa:
This test sets up multiple nodes on the localhost and tests DHT commands

### DIFF
--- a/lbrynet/tests/unit/dht/test_mutinode.py
+++ b/lbrynet/tests/unit/dht/test_mutinode.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+#
+# This library is free software, distributed under the terms of
+# the GNU Lesser General Public License Version 3, or any later version.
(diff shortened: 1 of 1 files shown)


lbrynet/tests/unit/dht/test_mutinode.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+#
+# This library is free software, distributed under the terms of
+# the GNU Lesser General Public License Version 3, or any later version.
+# See the COPYING file included in this archive
```
eukreign: This test is being licensed under GPL? But the rest of the project is MIT license.

ralmcruz: I would like to know that too. Why not licence the all thing under GPL?

Issue #977, opened 2017-10-31
store RPC call should save the external IP instead of the packet sender IP


ordex:
When receiving a `store` RCP call, a node currently uses the datagram sender IP address to compute the data to append to the peer list for the given hash.
This means that if a peer receives a `store` from another node in the same LAN, it will save the LAN IP in the DHT. This is not correct as this IP won't be usable by other nodes in the DHT that are not in the same LAN.

A node handling a `store` should always save the external IP of the announcing node.
This probably means changing the packet format and ensuring that the external IP is sent along with the `store` request.

ordex:
This issue is related to #974 as this is the logic for storing IPs in the DHT.

lyoshenka:
does #1007 fix this?

ordex:
@lyoshenka yes

alyssaoc:
@jackrobison does the new UPnP solve this?

jackrobison:
@alyssaoc no, but I'm not sure that this issue will actually happen anyway. The case that I imagine is broken is where a node is behind multiple layers of UPnP gateways, but that's an edge case probably not worrying about immediately. I'll test if the incoming ip of a node in our lan sending us a `store` is the external or lan ip when in mht this week.

lyoshenka:
why is this a blocker?

eukreign:
@jackrobison if this is a valid issue can we add integration tests for it?

eukreign:
oop, didn't mean to close

alyssaoc:
this is related to hole punching, which address and port you have vs what you expect to have

PR #1007, opened 2017-11-27
restructure UPnP code

ordex:
This PR restructures the UPnP code in order to accomplish the following:
* use the new `addanyportmapping()` function when creating mapping
* implement mapping logic specified in https://github.com/lbryio/lbry/pull/973#issuecomment-340514271
* add `external_peer_port` and `external_dht_node_port` fields to Session object. These fields will later be used to properly advertise the actual external ports allocated by UPnP (to be implemented in another PR)


lbrynet/conf.py
```diff
@@ -200,6 +200,9 @@ def _convert_value(value):
 TYPE_CLI = 'cli'
 TYPE_RUNTIME = 'runtime'
 
+DEFAULT_PEER_PORT = 3333
+DEFAULT_DHT_NODE_PORT = 4444
```
kauffj: :heart_eyes:

lbrynet/core/Session.py
```diff
@@ -119,6 +121,7 @@ def __init__(self, blob_data_payment_rate, db_dir=None,
         self.blob_tracker_class = blob_tracker_class or BlobAvailabilityTracker
 
         self.peer_port = peer_port
+        self.external_peer_port = peer_port
```
kaykurokawa: dht.node class be initialized with self.external_peer_port instead of self.peer_port for its peerPort argument.

kaykurokawa:
Work for either jack or me: 

Need to have some way in the daemon to obtain what Session.external_peer_port , Session.external_ip, and  Session.external_dht_node_port are set to (probably belongs in jsonrpc_status command )

lbrynet/conf.py
```diff
@@ -246,7 +251,7 @@ def _convert_value(value):
     'data_dir': (str, default_data_dir),
     'data_rate': (float, .0001),  # points/megabyte
     'delete_blobs_on_remove': (bool, True),
-    'dht_node_port': (int, 4444),
+    'dht_node_port' : (int, None),
     'download_directory': (str, default_download_dir),
```
kaykurokawa: Not sure if this is the best way to do this since ADJUSTABLE_SETTINGS is meant to store the default value.. so if you do the daemon API command  settings_get, you will get back that dht_node_port and peer_port is None if user haven't set them, which can cause confusion. 

Another problem is if the user wants to set the dht_node_port to some other non default value but also still wants to UPNP to map to any port, this is not possible.

I think alternatively, we can create another setting like 'upnp_any_port_mapping' that turns this feature on/off.

lbrynet/core/Session.py
ordex: Is this something you want me to change in this PR? I was planning to actually "use" the external port fields in a next PR.

lbrynet/conf.py
ordex: > Not sure if this is the best way to do this since ADJUSTABLE_SETTINGS is meant to store the default value.. so if you do the daemon API command settings_get, you will get back that dht_node_port and peer_port is None if user haven't set them, which can cause confusion.

I see - this wasn't clear from the code. I will fix that.

> Another problem is if the user wants to set the dht_node_port to some other non default value but also still wants to UPNP to map to any port, this is not possible.

Right, but apparently this was not wanted by design (check https://github.com/lbryio/lbry/pull/973#issuecomment-340514271).

> I think alternatively, we can create another setting like 'upnp_any_port_mapping' that turns this feature on/off. This solution seems cleaner to me.

Either strategy works for me. Maybe @lyoshenka has an opinion here?

ordex: >> I think alternatively, we can create another setting like 'upnp_any_port_mapping' that turns this feature on/off. This solution seems cleaner to me.

> Either strategy works for me. Maybe @lyoshenka has an opinion here?

by the way, my opinion about this point: if a user is configuring a specific port, his goal is probably to let that port be the one used "outside" (i.e. due to specific firewall rules) and therefore does not want the auto-selection to be in place.

I am sure there might be some use cases where having the possibility to set a specific port *and* having the automatic external port selection enabled might be helpful.
However, I think those cases are not worth the introduction of yet another config knob because it will increase the configuration complexity for very little gain.

lbrynet/core/Session.py
kaykurokawa: what was going to be in this PR? Other than this change?

lbrynet/conf.py
kaykurokawa: I get your above immediate above point. I don't have a strong preference for either way, but if we are going to go with the current implementation, than the dht_node_port in ADJUSTABLE_SETTINGS should be set to the default value and not None (and than we turn upnp any port map on/off depending on if the port is set to default value) OR, we need to create a mechanism in the conf.py that detects if some setting has been touched by the user or not because that doesn't exist right now.  Otherwise as I mentioned before, API command settings_get will return None for the port values which is not informative.

lbrynet/core/Session.py
ordex: Yes, right, but this is what is already happening right now in the code. It is not introduced by this PR.
But I can append some more commits. I just wanted to push some work out for review and this is happening already :)

lbrynet/conf.py
ordex: > OR, we need to create a mechanism in the conf.py that detects if some setting has been touched by the user or not because that doesn't exist right now.

This is probably a good the cleanest approach (setting None was just a way to work around that missing part).

kaykurokawa:
Two things needed to be done here by either jack or me: 

a) Expose the new external port variables in Session class to some daemon command 
b) have conf.py detect if a setting has been touched by the user 

Antonio will make some additional commits on this PR , that uses the external ports in the dht

ordex:
>  b) have conf.py detect if a setting has been touched by the user

I addressed b) with my two new commits as I was already digging around the conf.py code. Please let me know if that kind of approach is reasonable. Thanks!

ordex:
Something for @kaykurokawa : when the user specified a port in the config and this port is already mapped (i.e. by another host in the LAN), the upnp setup now fails and throws an exception. This exception is handled by sending a shutdown signal to the reactor.

Everything looks ok except for the Wallet which throws an exception. My guess is that the cleanup routine of the wallet, triggered by the shutdown, is trying to cleanup something that was not yet initialized at that point. It's not a major problem because the node is shutting down and it actually stops, but we might want to have a look at that.

lbrynet/conf.py
```diff
@@ -213,6 +216,8 @@ def _convert_value(value):
         'USD': {'type': 'fiat'},
     },
     'DB_REVISION_FILE_NAME': 'db_revision',
+    'DEFAULT_PEER_PORT': DEFAULT_PEER_PORT,
+    'DEFAULT_DHT_NODE_PORT': DEFAULT_DHT_NODE_PORT,
```
kaykurokawa: These FIXED_SETTINGS variables seem unused now

kaykurokawa:
Added a commit to retrieve the external IP and ports through API command "status -s"

lbrynet/conf.py
ordex: yeah, you are right. it can be removed

ordex:
* rebased on top of latest master
* merged patches doing and undoing the same code
* removed DEFAULT_*_PORTs from FIXED_SETTINGS since they are not used anymore

kaykurokawa:
Added a unit test for the is_default function , 
Is there further work to be done on this @ordex ?

ordex:
Thanks for the unit test @kaykurokawa .

There is no more work for this PR.

kaykurokawa:
@ordex will add more to this PR, to deal with external IP's

lbrynet/dht/node.py
```diff
@@ -475,7 +475,7 @@ def store(self, key, value, originalPublisherID=None, self_store=False, **kwargs
 
         if 'port' in value:
             port = int(value['port'])
-            if 0 <= port <= 65536:
+            if 0 <= port <= 65535:
```
lyoshenka: is 0 a valid port?

ordex: potentially yes, but it's reserved and can't really be used, so we could also exclude it from the check

ordex: the check has been changed

ordex:
@kaykurokawa with this updated PR I am changing the way message headers are constructed so they can contain the external IP/port that a receiver should store in the DHT (rather than using the UDP source). Any reasonable way to test this new logic? What would you recommend?

lbrynet/tests/unit/dht/test_messages.py
```diff
@@ -13,36 +13,44 @@
 class DefaultFormatTranslatorTest(unittest.TestCase):
     """ Test case for the default message translator """
     def setUp(self):
-        self.cases = ((RequestMessage('1' * 48, 'rpcMethod',
+        self.cases = ((RequestMessage('1' * 48, '127.0.0.1', 4444, 'rpcMethod',
```
ordex: @kaykurokawa do we already have some function to pack an IP address in binary form instead of sending the whole string? I guess we can just do that with some python primitive?

ordex: I am converting IP from str to binary form using:  ```[int(x) for x in message.nodeIP.split('.')]``` and then back with ```".".join([str(d) for d in msgPrimitive[self.headerNodeIP][:4]])```

ordex:
@kaykurokawa due to the fact that I am adding new fields to the messages headers (the IP is now advertised via header as well) I also introduced the compatibility version field.
We quickly discussed about this in the past. Its goal is to avoid that nodes "speaking" an old wire protocol can crash on unknown message formats.
We should test this carefully to make sure it works as expected. Tests are all passing so far.

ordex:
Given that with this PR the external IP is sent via message header, it gets useful to allow the user to set the IP manually instead of letting the node autodetect it. For this reason a new key 'external_ip' has been added to the configuration settings.
This is especially useful when testing a bunch of nodes on localhost: the IP should be manually set to 127.0.0.1 so that all the instances will be able to reach each other without passing through the public IP (which won't work when upnp is not used).

lbrynet/dht/msgformat.py
```diff
@@ -43,10 +44,13 @@ def toPrimitive(self, message):
...
     typeRequest, typeResponse, typeError = range(3)
-    headerType, headerMsgID, headerNodeID, headerNodeIP, headerNodePort = range(5)
-    headerPayload, headerArgs = range(5, 7)
+    headerCompVer, headerType, headerMsgID, headerNodeID, headerNodeIP, headerNodePort = range(6)
+    headerPayload, headerArgs = range(6, 8)
 
     def fromPrimitive(self, msgPrimitive):
+        if msgPrimitive[self.headerCompVer] != COMPAT_VERSION:
+            return
+
```
kaykurokawa: is "return" the proper thing to do here? 
Maybe we should be throwing a exception and catching it here:
https://github.com/lbryio/lbry/blob/master/lbrynet/dht/protocol.py#L206

kaykurokawa:
Tested with lbry-in-a-box, there were no errors there. 

So this is going to break backwards compatibility with the old nodes (which is fine), but I would like to know how the old nodes will break (what ERROR message should we expect to see on the old nodes so @tzarebczan will be able to diagnose users) , and whether the compatibility version changes will give you a useful error message (which it does not seem like it does). 

Otherwise I think it looks good to merge.

lbrynet/dht/msgformat.py
ordex: Yeah, that would be more inline with our coding style.

ordex:
@kaykurokawa I added a commit to convert the return to exception (in case of mismatching compat version).

I also fixed a small glitch I introduced during the last rebase in "Daemon: avoid exceptions on premature shutdown".

kaykurokawa:
This LGTM.

Should note that nodes with this PR is incompatible with the rest of the network (The below warning will be spammed in the logs: WARNING  lbrynet.dht.node:596: This node doesnt know any other nodes ),

So while it can be merged into master, it makes master not work with the rest of the network, unless a) our reflector hosts are updated and b) app-users get the PR.

lyoshenka:
this fixes #977 as well

lbrynet/daemon/Daemon.py
```diff
@@ -236,6 +239,8 @@ def _announce():
                 log.info("Started lbrynet-daemon")
                 log.info("%i blobs in manager", len(self.session.blob_manager.blobs))
 
+            if self.session.blob_manager is None:
+                return
```
jackrobison: this will throw a deferred generator exception (returning from inlinecallbacks)

lbrynet/daemon/Daemon.py
```diff
@@ -574,8 +579,13 @@ def create_session(wallet):
             )
             self.startup_status = STARTUP_STAGES[2]
 
+        def _handle_session_setup_error(err):
+            log.warning('Error while setting up session: %s', err.getErrorMessage())
```
jackrobison: this should be log.error

lbrynet/conf.py
```diff
@@ -24,6 +24,9 @@
 
 ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+# Compatibility Version of the wire protocol. Increasing this will break backward compatibility
+COMPAT_VERSION = 1
```
jackrobison: i think this would be better in lbrynet.dht.constants

lbrynet/daemon/Daemon.py
jackrobison: also, why is this necessary?

ordex: This change is part of this commit https://github.com/lbryio/lbry/pull/1007/commits/15fcd9a5f6df3aa1b534846c912003ccab2ff6f9
Basically I realized that when upnp fails and a shutdown is triggered in the middle of the startup, the blob_manager initialization is also stopped (as expected), but other routines still try to access it.
I added this check to prevent them from hitting the initialized object and explode.

lbrynet/daemon/Daemon.py
ordex: oky

lbrynet/conf.py
ordex: yeah, makes sense. I'll move it

lbrynet/conf.py
```diff
@@ -252,9 +258,10 @@ def _convert_value(value):
     'data_dir': (str, default_data_dir),
     'data_rate': (float, .0001),  # points/megabyte
     'delete_blobs_on_remove': (bool, True),
-    'dht_node_port': (int, 4444),
+    'dht_node_port' : (int, DEFAULT_DHT_NODE_PORT),
```
eukreign: Extraneous space between key string and colon.

lbrynet/conf.py
```diff
@@ -268,7 +275,7 @@ def _convert_value(value):
     'min_info_rate': (float, .02),  # points/1000 infos
     'min_valuable_hash_rate': (float, .05),  # points/1000 infos
     'min_valuable_info_rate': (float, .05),  # points/1000 infos
-    'peer_port': (int, 3333),
+    'peer_port' : (int, DEFAULT_PEER_PORT),
```
eukreign: Extraneous space between key string and colon.

lbrynet/core/Session.py
```diff
@@ -193,30 +196,56 @@ def _try_upnp(self):
...
+                                                      description, '')
+                else:
+                    # addportmapping() raises an exception on failure
+                    upnp.addportmapping(port, protocol, upnp.lanaddr, port, description, '')
+                    ext_port = port
+            except Exception as e:
+                log.warning('UPnP failed with error: %s', e.message)
+                return None
+
+            return ext_port
```
eukreign: A bit shorter and tidied up:
```
    def add_port_mapping(upnp, port, auto_port, protocol, description):
        log.debug('UPNP automatic port selection enabled: {}'.format(auto_port))
        try:
            if auto_port:
                return upnp.addanyportmapping(
                    port, protocol, upnp.lanaddr, port, description, ''
                )
            else:
                upnp.addportmapping(port, protocol, upnp.lanaddr, port, description, '')
                return port
        except Exception as e:
            log.warning('UPnP failed with error: %s', e.message)
```

ordex: I normally prefer to keep the returns out, but if like this style, I'll follow that.
For the function arguments, I really prefer to see them aligned to the opening parenthesis, unless we have a codestyle suggesting not to

lbrynet/conf.py
ordex: ops, thanks (for both)!

lbrynet/core/Session.py
eukreign: If the function has finished and you have the value to return it's good to just return the value. Someone reading your code will quickly see that after the appropriate value is computed nothing else happens. If you only put returns at the end of methods it makes it seem like other stuff happens. Also, it require temporary variables and just more lines of code. More lines of code means more technical debt.

I don't know if there is an official lbry style on this particular point. You can leave that part as you had it until we figure out if we need a consistent style. Personally, I find that lining up arguments with parenthesis leads to longer lines and big gaps of empty space below the function and to the left of the parenthesis. If you put arguments on a new line and indent 4 spaces you can fit more arguments, more concisely.

ordex: more lines of code are good if they increase readability. But I understand in this case I see there is no real gain.

Imho having the return at the end of the function with a specific variable makes you understand that, no matter what happens, we will return the content of that little box. How it is computed may be of interest or may not.

While having the returns into indented blocks forces you to dig in to understand what we are really returning. Well, I guess in python you'd need to do that anyway because there is not strict typing...so ext_port may contain any type of object..

I'll move the returns inside.

About the extra variable: I am used to compiled languages where you don't really care about it because it gets optimized out. Often in the linux kernel extra variables are used to avoid long conditions/expressions/lines and it works pretty well in my experience.
But I guess in python this does not happen as it is a scripted language with sequential parsing. I'll remove it.

Thanks for your feedback

ordex:
@jackrobison @eukreign I have rebased my branch on top of latest master and I addressed your comments. However Appveyor does not seem to be happy. I see it is still using miniupnpc 1.9 instead of 2.0.2 (I updated requirements.txt but I guess it is pulling the version from somewhere else)

ordex:
miniupnpc-2.0.2 seems to be ok now (after some mingw surgery), however Appveyor is still exploding

ordex:
rebased on top of latest master

jackrobison:
I'd prefer this pr to only have upnp changes and not change the DHT message format - as that makes merging and deploying it more complicated, and we can't get the benefits of the upnp changes until then. The dht message format changes can go in a separate pr and be merged along with the other breaking dht changes in our next sprint

ordex:
@jackrobison I think the changes are somehow intertwined, but I definitely understand your point.
I will dig into the commits a bit and see how to split them into 2 separate PRs.

ordex:
@jackrobison I think I have removed all the changes affecting the DHT wire format

lbrynet/core/Session.py
```diff
@@ -195,30 +198,54 @@ def _try_upnp(self):
...
-        def get_free_port(upnp, port, protocol):
-            # returns an existing mapping if it exists
+        def get_mapped_port(upnp, port, protocol):
+            # returns an existing mapped port if it exists
             mapping = upnp.getspecificportmapping(port, protocol)
-            if not mapping:
-                return port
-            if upnp.lanaddr == mapping[0]:
-                return mapping
-            return get_free_port(upnp, port + 1, protocol)
```
jackrobison: `get_mapped_port` doesn't handle there being a redirect for a different address in the lan (it will return None)

lbrynet/core/Session.py
```diff
@@ -195,30 +198,54 @@ def _try_upnp(self):
...
+            if mapping is not None and upnp.lanaddr == mapping[0]:
+                return mapping[1]
+            return None
+
+        def add_port_mapping(upnp, port, auto_port, protocol, description):
+            log.debug('UPNP automatic port selection enabled: ' + str(auto_port))
+            try:
+                if auto_port is True:
+                    # addanyportmapping() returns None on failure
+                    return upnp.addanyportmapping(port, protocol, upnp.lanaddr, port,
```
jackrobison: This fails with an ambiguous `UnknownError` if there is an existing redirect for another lan address, whereas `addportmapping` will raise `ConflictInMappingEntry`.

ordex: weird, `addanyportmapping` is supposed to assign the first available port starting from the one specified as argument. Basically it is doing what we were doing before in our routine, but in an atomic and non-racy way. This is the main reason why I introduced it. It should then return the assigned port or none if it couldn't find a port.
But it sounds like you observed a different behaviour?

jackrobison:
We'll get back to this after we release 0.19.0

jackrobison:
Closing this for now, can be re-opened if we want to cherry-pick anything out.

@eukreign is going to be updating `miniupnpc` to be current when he adds the windows build to travis

PR #1009, opened 2017-11-28
add claim_renew

jackrobison:


kaykurokawa:
Added auto_renew_claim_height_delta to config, 
Added auto renew feature on startup 
Made auto_renew_claim_height_delta settable from settings_set 

Still need to test a bit , also the docstring for settings_set is exceeding pyling line limit, need to figure out a way to avoid that without upsetting the docopt parser

kaykurokawa:
Fixed the pylint line limit on the docopt docstrings

kaykurokawa:
Bug fix and made sure that the return dictionary for claim_renew is consistent, whether you renew by height or by outpoint. 

This is now ready for jack to review again and merge.

lbrynet/daemon/Daemon.py
```diff
@@ -293,6 +296,23 @@ def _update_connection_status(self):
         if not self.connected_to_internet:
             self.connection_status_code = CONNECTION_STATUS_NETWORK
 
+    @defer.inlineCallbacks
+    def _auto_renew(self):
+        # automatically renew claims
+        # auto renew is turned off if 0 or some negative number
+        if self.auto_renew_claim_height_delta < 1:
+            return
```
jackrobison: this will raise a deferred generator error (should be `defer.returnValue(None)`)

lbrynet/daemon/Daemon.py
```diff
@@ -293,6 +296,23 @@ def _update_connection_status(self):
         if not self.connected_to_internet:
             self.connection_status_code = CONNECTION_STATUS_NETWORK
 
+    @defer.inlineCallbacks
+    def _auto_renew(self):
+        # automatically renew claims
+        # auto renew is turned off if 0 or some negative number
+        if self.auto_renew_claim_height_delta < 1:
```
jackrobison: This should also check that lbryum is caught up

kaykurokawa: We should be all caught up by the time it calls auto_renew() since it comes after setup of the Wallet

lbrynet/daemon/Daemon.py
kaykurokawa: fixed

PR #1012, opened 2017-11-29
add lbrynet-console

jackrobison:
screen shot 2017-12-05 at 11 44 17 am


kaykurokawa:
LGTM, 
An argument that shuts off its ability to startup its own daemon would be nice (since you can't immediately connect to the daemon on daemon startup and then  lbrynet-console will start up its own daemon)

PR #1025, opened 2017-12-12
refactor databases

jackrobison:


lyoshenka:
@jackrobison who needs to review this? i glanced at it briefly and i still think more columns should be `not null`

jackrobison:
@lyoshenka please check out https://github.com/lbryio/lbry/pull/1025/commits/fd0a7c3b387cf8de82cfd3fc22d361d57ab6573c , I rebased in some changes

lbrynet/database/migrator/migrate5to6.py
```diff
@@ -0,0 +1,164 @@
...
+                sd_hash char(96) not null,
+                stream_key text not null,
+                stream_name text not null,
+                suggested_filename text not null,
+                foreign key(sd_hash) references blob(blob_hash)
+            );
+            
+            create table stream_blob (
+                stream_hash char(96) not null,
+                blob_hash char(96),
```
lyoshenka: my note disappeared somehow. `blob_hash` should be not null. a stream_blob's purpose is to connect a stream and a blob. if there's no blob, there's no stream_blob

kauffj:
I think `file_list` should probably retain all of: `outpoint`, `claim_id`, `channel_claim_id` (doesn't exist as first class field currently and is instead currently buried 4 fields deep in an inconsistent way), and `metadata` (but this could presumably be what is currently `file.metadata.stream.metadata`).

I don't see how an application using LBRY can re-render to a user what they downloaded without these fields, or an efficient way to select them again.

lbrynet/core/BlobManager.py
```diff
@@ -1,21 +1,17 @@
...
 from lbrynet.blob.creator import BlobFileCreator
 from lbrynet.core.server.DHTHashAnnouncer import DHTHashSupplier
-from lbrynet.core.sqlite_helpers import rerun_if_locked
 
 log = logging.getLogger(__name__)
 
 
 class DiskBlobManager(DHTHashSupplier):
-    def __init__(self, hash_announcer, blob_dir, db_dir):
+    def __init__(self, hash_announcer, blob_dir, storage):
```
lyoshenka: is it possible to require a class or interface for arguments? i feel like `storage` should be an interface with functions like "addBlobs", "getBlobs", etc. Then we can implement multiple types of storage (e.g. `sqlite` for production, `in-memory` for testing) and it's clear what the contract is.

A simple way of doing this now without too much refactoring is to create an interface that has all the functions that are already in use. Then we can pare it down later.

lbrynet/database/migrator/migrate5to6.py
lyoshenka: @jackrobison this still applies

lbrynet/file_manager/EncryptedFileCreator.py
```diff
@@ -70,8 +63,27 @@ def _make_stream_hash(self):
...
+        for blob_info in self.blob_infos:
+            blob = {}
+            if blob_info.length != 0:
+                blob['blob_hash'] = blob_info.blob_hash
+            blob['blob_num'] = blob_info.blob_num
+            blob['iv'] = blob_info.iv
+            blob['length'] = blob_info.length
+            formatted_blobs.append(blob)
+        sd_blob['blobs'] = formatted_blobs
+        return defer.succeed(sd_blob)
```
eukreign: I'm having a bit of déjà vu here, is this function defined already somewhere else? `lbrynet/core/StreamDescriptor.py` They may do something slightly different but the function names are identical and they appear to be very similar internally.

lbrynet/database/migrator/migrate5to6.py
jackrobison: this is to handle stream terminators

.pylintrc
```diff
@@ -290,7 +290,7 @@ spelling-store-unknown-words=no
 [FORMAT]
 
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=120
```
eukreign: I've been thinking that maybe 120 is too much. In cases where you want to have two files side by side, even in PyCharm at slightly enlarged font (like mine), I can't comfortably fit two 120 lines side by side. 100 fits though... what do you think? Also, Python PEP8 suggests 100 as the max: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

lbrynet/tests/unit/lbryfilemanager/test_EncryptedFileCreator.py
```diff
@@ -21,43 +22,54 @@ def iv_generator():
...
-        session.blob_manager = self.blob_manager
-        yield session.blob_manager.setup()
-        session.db_dir = self.tmp_db_dir
         manager = mock.Mock(spec=EncryptedFileManager.EncryptedFileManager)()
         handle = mocks.GenFile(3*MB, '1')
         key = '2'*AES.block_size
-        out = yield EncryptedFileCreator.create_lbry_file(
-            session, manager, filename, handle, key, iv_generator())
+        out = yield EncryptedFileCreator.create_lbry_file(self.session, manager, filename, handle,
+                                                          key, iv_generator())
```
eukreign: This seems like a reverse formatting operation back to the pan handle :-D

lbrynet/database/storage.py
```diff
@@ -0,0 +1,457 @@
...
+            
+            create table if not exists stream_blob (
+                stream_hash char(96) not null,
+                blob_hash char(96),
+                position integer not null,
+                iv char(32) not null,
+                primary key (stream_hash, blob_hash),
+                foreign key(stream_hash) references stream(stream_hash),
+                foreign key (blob_hash) references blob(blob_hash)
+            );
```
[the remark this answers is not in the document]
jackrobison: I'll add a commit updating the db migrator at the end of the pr. Since moving the create tables statement out to a class attribute of `SQLiteStorage`, the migrator can use it instead of having a  duplicate create tables query (which is prone to being out of sync)

lbrynet/core/StreamDescriptor.py
```diff
@@ -239,6 +242,165 @@ def get_options():
...
+
+        if include_blobs is True:
+            d = storage.get_blobs_for_stream(stream_hash)
+        else:
+            d = defer.succeed([])
+        d.addCallback(format_blobs)
+        return d
+
+    d.addCallback(format_info)
+    return d
```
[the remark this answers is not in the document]
jackrobison: https://github.com/lbryio/lbry/blob/be57cc199f42c8f12b0d39f2c7eca81c36d1b2c4/lbrynet/core/StreamDescriptor.py#L271

lbrynet/core/BlobManager.py
```diff
@@ -75,7 +63,7 @@ def _immediate_announce(self, blob_hashes):
     def blob_completed(self, blob, next_announce_time=None, should_announce=True):
         if next_announce_time is None:
             next_announce_time = self.get_next_announce_time()
-        yield self._add_completed_blob(blob.blob_hash, blob.length,
+        yield self.storage.add_completed_blob(blob.blob_hash, blob.length,
                                        next_announce_time, should_announce)
```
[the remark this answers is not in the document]
jackrobison: :+1:

lbrynet/core/Session.py
```diff
@@ -313,12 +315,10 @@ def _setup_other_components(self):
...
-        d1 = self.blob_manager.setup()
-        d2 = self.wallet.start()
-
-        dl = defer.DeferredList([d1, d2], fireOnOneErrback=True, consumeErrors=True)
-        dl.addCallback(lambda _: self.blob_tracker.start())
-        return dl
+        d = self.storage.setup()
+        d.addCallback(lambda _: self.wallet.start())
+        d.addCallback(lambda _: self.blob_tracker.start())
+        return d
```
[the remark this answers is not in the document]
jackrobison: It shouldn't - setup on `storage` just creates the sqlite tables.

lbrynet/core/StreamDescriptor.py
```diff
@@ -239,6 +242,165 @@ def get_options():
...
+        d = defer.succeed(file_name)
+
+    def do_create(file_name):
+        descriptor_writer = PlainStreamDescriptorWriter(file_name)
+        d = get_sd_info(storage, stream_hash, True)
+        d.addCallback(descriptor_writer.create_descriptor)
+        return d
+
+    d.addCallback(do_create)
+    return d
```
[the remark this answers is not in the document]
jackrobison: done

lbrynet/core/StreamDescriptor.py
```diff
@@ -239,6 +242,165 @@ def get_options():
...
+                blob_hash = None
+            blob_num = b['blob_num']
+            iv = b['iv']
+            blob_hashsum = get_lbry_hash_obj()
+            if length != 0:
+                blob_hashsum.update(blob_hash)
+            blob_hashsum.update(str(blob_num))
+            blob_hashsum.update(iv)
+            blob_hashsum.update(str(length))
+            return blob_hashsum.digest()
```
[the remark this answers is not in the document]
jackrobison: done

lbrynet/database/migrator/migrate5to6.py
```diff
@@ -0,0 +1,158 @@
...
+        claim_info = name_metadata_cursor.execute("select claimId, name from claim_ids "
+                                     "where txid=? and n=?", (txid, nout)).fetchone()
+        if not claim_info:
+            log.warning("Missing claim id, cannot migrate stream %s", stream_hash)
+            del file_infos[stream_hash]
+            continue
+        claim_id, name = claim_info
+        file_infos[stream_hash]['claim_name'] = name
+        file_infos[stream_hash]['claim_id'] = claim_id
+        file_infos[stream_hash]['outpoint'] = "%s:%i" % (txid, nout)
```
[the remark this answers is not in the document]
jackrobison: converted these kind of things in the migrator script to use dict comprehensions

lbrynet/database/storage.py
```diff
@@ -0,0 +1,457 @@
...
+                                                          blob_hash))
+
+    def set_blob_status(self, blob_hash, status):
+        return self.db.runOperation("update blob set status=? where blob_hash=?", (status,
+                                                                                   blob_hash))
+
+    def get_blob_status(self, blob_hash):
+        d = self.db.runQuery("select status from blob where blob_hash=?", (blob_hash, ))
+        d.addCallback(lambda r: None if not r else r[0][0])
+        return d
```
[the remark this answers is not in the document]
jackrobison: done

lbrynet/database/storage.py
```diff
@@ -0,0 +1,457 @@
...
+        if status is None:
+            status = "pending"
+            yield self.db.runOperation("insert into blob values (?, ?, ?, ?, ?)",
+                                       (blob_hash, length, 0, 0, status))
+        defer.returnValue(status)
+
+    def should_announce(self, blob_hash):
+        d = self.db.runQuery("select should_announce from blob where blob_hash=?", (blob_hash, ))
+        d.addCallback(lambda r: None if not r else r[0][0])
+        return d
```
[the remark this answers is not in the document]
jackrobison: done

lbrynet/file_manager/EncryptedFileCreator.py
jackrobison: now the formatting is only in `lbrynet/core/StreamDescriptor.py`

.pylintrc
jackrobison: I like 120 but don't particularly prefer it to 100.

lbrynet/database/migrator/migrate5to6.py
```diff
@@ -0,0 +1,158 @@
...
+            );
+            
+            create table stream (
+                stream_hash char(96) not null primary key,
+                sd_hash char(96) not null,
+                stream_key text not null,
+                stream_name text not null,
+                suggested_filename text not null,
+                foreign key(sd_hash) references blob(blob_hash)
+            );
```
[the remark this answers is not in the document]
jackrobison: done

lbrynet/database/migrator/migrate5to6.py
```diff
@@ -0,0 +1,158 @@
...
+            
+            create table stream_blob (
+                stream_hash char(96) not null,
+                blob_hash char(96),
+                position integer not null,
+                iv char(32) not null,
+                primary key (stream_hash, blob_hash),
+                foreign key(stream_hash) references stream(stream_hash),
+                foreign key (blob_hash) references blob(blob_hash)
```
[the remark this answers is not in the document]
jackrobison: done

.pylintrc
eukreign: I like 120 in general, but on the rare occasion that you want to have two files open at the same time, it helps a lot to have slightly more reasonable lines.

Actually, since you use the PyCharm git tools, I'd think you would prefer the shorter line length too since PyCharm shows diffs side by side.

.gitignore
```diff
@@ -23,3 +23,4 @@
 
 # temporary files from the twisted.trial test runner
 _trial_temp/
+.tox/
```
eukreign: Can we keep this out of the .gitignore for now, unless it's causing problems. Once I'm done learning/testing lbryum I'll move onto lbry and will probably set it up the same way as lbryum (with codecove and tox with python2 and python3).

Tox makes it easy to test code under different versions of Python and also to make sure it's packaged properly (as a Python pip'able package). Which regular testing from source doesn't verify.

jackrobison: removed, I'd added it because I was messing around with a tox configuration

PR #1030, opened 2017-12-13
Fix slow file manager startup on spee.ch

jackrobison:
* Fixed the file manager startup being slow when handling thousands of files

### DIFF
--- a/lbrynet/file_manager/EncryptedFileDownloader.py
+++ b/lbrynet/file_manager/EncryptedFileDownloader.py
@@ -54,22 +54,16 @@ class ManagedEncryptedFileDownloader(EncryptedFileSaver):
     def saving_status(self):
         return self._saving_status
 
-    @defer.inlineCallbacks
-    def restore(self):
-
-        status = yield self.lbry_file_manager.get_lbry_file_status(self)
-        log_status(self.sd_hash, status)
-
+    def restore(self, status):
         if status == ManagedEncryptedFileDownloader.STATUS_RUNNING:
             # start returns self.finished_deferred
             # which fires when we've finished downloading the file
             # and we don't want to wait for the entire download
             self.start()
         elif status == ManagedEncryptedFileDownloader.STATUS_STOPPED:
-            defer.returnValue(False)
+            pass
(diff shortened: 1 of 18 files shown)


lbrynet/file_manager/EncryptedFileManager.py
```diff
@@ -86,53 +85,49 @@ def _add_to_sd_identifier(self):
...
+                )
+                try:
+                    # restore will raise an Exception if status is unknown
+                    lbry_file.restore(status)
+                except Exception:
+                    log.warning("Failed to start %i", rowid)
+                    continue
+                self.lbry_files.append(lbry_file)
+        log.info("Started %i lbry files", len(self.lbry_files))
         if self.auto_re_reflect is True:
```
kaykurokawa: I think  that start_lbry_file() function immediately below this should create ManagedEncryptedFileDownloader() in the same manner as _start_lbry_files(). Otherwise we have two completely different ways of creating ManagedEncryptedFileDownloader instances. 

This should mean that we should be able to delete some functions in the base classes of ManagedEncryptedFileDownloader like https://github.com/lbryio/lbry/blob/8eba9dc4b1b2a945fc66b7c5aab7354aad05ad50/lbrynet/lbry_file/client/EncryptedFileDownloader.py#L32 

that on its own pulls things from the stream_info_manager. 

Overall I think its good that the file manager is the one and only place responsible for pulling in data from stream_info_manager.

lbrynet/file_manager/EncryptedFileDownloader.py
```diff
@@ -28,50 +28,46 @@ def log_status(sd_hash, status):
...
                                     download_directory,
                                     file_name)
-        self.sd_hash = None
         self.rowid = rowid
         self.lbry_file_manager = lbry_file_manager
         self._saving_status = False
+        self.key = key
+        self.stream_name = stream_name
+        self.suggested_file_name = suggested_file_name
+        self.sd_hash = sd_hash
```
kaykurokawa: I think these variable need to be passed down to EncryptedFileSaver instead of being intialized here because they are managed there or in another base class CryptStreamDownloader .

PR #1034, opened 2017-12-17
Add stream and blob availability commands

jackrobison:


jackrobison:
Note that this doesn't redirect calls to the deprecated `get_availability` to the new `stream_availability`, I'll add that

lbrynet/core/Session.py
```diff
@@ -136,6 +137,7 @@ def __init__(self, blob_data_payment_rate, db_dir=None,
         self.payment_rate_manager = None
         self.payment_rate_manager_class = payment_rate_manager_class or NegotiatedPaymentRateManager
         self.is_generous = is_generous
+        self.single_peer_downloader = SinglePeerDownloader()
```
lyoshenka: why does a session always have a single-peer downloader? this functionality is unnecessary in 99% of cases. it should only be on when we're gonna use it

lbrynet/core/Session.py
```diff
@@ -318,6 +320,7 @@ def _setup_other_components(self):
 
         dl = defer.DeferredList([d1, d2], fireOnOneErrback=True, consumeErrors=True)
         dl.addCallback(lambda _: self.blob_tracker.start())
+        dl.addCallback(lambda _: self.single_peer_downloader.setup(self.wallet, self.blob_manager))
```
lyoshenka: same comment as above

lbrynet/core/Wallet.py
```diff
@@ -27,8 +27,10 @@
 from lbrynet.core.sqlite_helpers import rerun_if_locked
 from lbrynet.interfaces import IRequestCreator, IQueryHandlerFactory, IQueryHandler, IWallet
 from lbrynet.core.client.ClientRequest import ClientRequest
-from lbrynet.core.Error import RequestCanceledError, InsufficientFundsError, UnknownNameError
+from lbrynet.core.Error import InsufficientFundsError, UnknownNameError
 from lbrynet.core.Error import UnknownClaimID, UnknownURI, NegativeFundsError, UnknownOutpoint
+from lbrynet.core.Error import DownloadCanceledError, RequestCanceledError
```
lyoshenka: shouldn't  all these `from lbrynet.core.Error import ...` lines be one line? or is this about having lines that are too long?

lbrynet/daemon/Daemon.py
```diff
@@ -2920,79 +2920,146 @@ def jsonrpc_routing_table_get(self):
         return self._render_response(result)
 
     @defer.inlineCallbacks
-    def jsonrpc_get_availability(self, uri, sd_timeout=None, peer_timeout=None):
+    def jsonrpc_blob_availability(self, blob_hash, search_timeout=None, blob_timeout=None):
```
lyoshenka: so `get_availability` isn't just deprecated here, its removed, right?

lbrynet/core/Wallet.py
jackrobison: >is this about having lines that are too long?

right

Issue #1039, opened 2017-12-28
Add --conf_file flag to allow using other configuration files besides the default


lyoshenka:
This will also permit running multiple daemons on the same machine.

Some refactoring will also be required. The CLI arg defaults should not depend on conf file values. The default for everything should be None, since we're using a tiered conf system.

lyoshenka:
Related to #1040

PR #1043, opened 2017-12-29
Add conf file flag

ordex:
This PR adds the --conf CLI flag as requested in #1039 

The first commit in this PR also exists in the upnp-restructuring branch. I'll remove it form there if this PR gets merged first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ at anytime.
   * Add `peer_port` to settings configurable using `settings_set`
   * Added an option to disable max key fee check.
   * Add `wallet_unlock`, a command available during startup to unlock an encrypted wallet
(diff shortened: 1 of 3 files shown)


kaykurokawa:
I think conf_file should not be in ADJUSTBALE_SETTINGS, that's too weird (you would be able to specify in the config file where the config file is, so what happens to the other config variables in that config file?)

I believe, it should just be a one time use member variable in Config class, that you set on load_conf_file_settings() (i.e, there is Config.conf_file , and Config.load_conf_file_settings(conf_file=None) sets Config.conf_file and opens Config.conf_file if specified and uses default location when it is None)

ordex:
I like the idea of not having it in ADJUSTBALE_SETTINGS. I'll try to rework this change.

kaykurokawa:
LGTM

PR #1053, opened 2018-01-09
Fix to go with https://github.com/lbryio/lbryum/pull/195

kaykurokawa:
The paytomany argument has been reworked so some changes are required to the daemon Wallet code https://github.com/lbryio/lbryum/pull/195


kaykurokawa:
renamed paytomany to payto in lbryum, 
rebased to single commit and added changelog

Issue #1008, opened 2017-11-28
Resolve, claim_show, and claim_list_mine return different claim signatures


kauffj:
Specifically, the `category` field is missing from resolve, but there may be other differences.

#### resolve (missing category sub-field)
```
./lbrynet-cli resolve squirreltryingtohidenutinadog
{
  "squirreltryingtohidenutinadog": {
    "claim": {
      "address": "bWTLALD64DDHmg9DieNUWrkd6Uzh4XrRRp",
      "amount": 10.0,
      "claim_id": "240e7ce3ae3bd40c728412af9964270f1ee5b7e8",
      "claim_sequence": 1,
      "decoded_claim": true,
      "depth": 24942,
      "effective_amount": 10.0,
      "has_signature": false,
      "height": 254705,
      "hex": "080110011ad7010801128001080410011a20546865204265737420506c61636520546f204869646520596f7572204e7574732224537175697272656c20747269657320746f2068696465206e757420696e206120646f672e2a00320d5075626c696320446f6d61696e38004a1b687474703a2f2f737065652e63682f737175697272656c2e706e6752005a001a50080110011a3086d7df8f9b7fa0ef92177859bc20c72ca7d7116d76eefcd99e812e6a462639ea231f0fffa12fa5b7acc33f727df0c54c22186170706c69636174696f6e2f6f637465742d73747265616d",
      "name": "squirreltryingtohidenutinadog",
      "nout": 1,
      "supports": [],
      "txid": "44a4c20e4fe88a7d063703cdaa62c45fa08c7103230950d7501fb4418a32355d",
      "value": {
        "claimType": "streamType",
        "stream": {
          "metadata": {
            "author": "",
            "description": "Squirrel tries to hide nut in a dog.",
            "language": "en",
            "license": "Public Domain",
            "licenseUrl": "",
            "nsfw": false,
            "preview": "",
            "thumbnail": "http://spee.ch/squirrel.png",
            "title": "The Best Place To Hide Your Nuts",
            "version": "_0_1_0"
          },
          "source": {
            "contentType": "application/octet-stream",
            "source": "86d7df8f9b7fa0ef92177859bc20c72ca7d7116d76eefcd99e812e6a462639ea231f0fffa12fa5b7acc33f727df0c54c",
            "sourceType": "lbry_sd_hash",
            "version": "_0_0_1"
          },
          "version": "_0_0_1"
        },
        "version": "_0_0_1"
      }
    }
  }
}
```

[...]

Issue #1006, opened 2017-11-26
claim_list should return uniform data structure


kauffj:
`claim_list` currently returns a different data signature based on what type of claim it is (channel vs. stream).

`claim_xxx` API functions promise that they return claims. Claims must have a singular, uniform definition. Fields that are not shared across claim types should either be returned as null or these values should be in a sub-property (e.g. claims could contain `stream` and `channel` sub-properties that contain either null or a map of type-specific properties).

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

lyoshenka:
part of https://github.com/orgs/lbryio/projects/5

eukreign:
All of the local commands (which query sqlite db) return uniform data structure for tx and txos. Only exception right now is data coming from lbryumx. Re-assigning this to @shyba for further consideration.

shyba:
If I understood it correctly, this relates closely to #1483 but I think it has a broader scope. The issue with missing claim data is being handled at 1483, but maybe we can say this issue is to make the response of `claim_xxx` commands fixed on a template right?

tzarebczan:
It's consistent between channels/claims now, but not consistent with other api commands that hit the wallet server. This will be covered by https://github.com/lbryio/lbry-sdk/issues/2259

PR #1063, opened 2018-01-20
Added permanent_url attribute to api calls that return claim information

hackrush01:


kauffj:
@hackrush01 `permanent_url` should include channel if it is published under one

lbrynet/core/Wallet.py
```diff
@@ -740,6 +740,9 @@ def _decode_and_cache_claim_result(self, claim, update_caches):
             name = claim['name']
             claim['value'] = claim_dict
             claim['hex'] = decoded.serialized.encode('hex')
+
+            # lbry/issue#958
+            claim['permanent_url'] = claim["name"] + "#" + claim["claim_id"]
```
jackrobison: These should use [formatted strings](https://docs.python.org/2/library/stdtypes.html#string-formatting), like `           claim['permanent_url'] = "%s#%s" % (claim['name'], claim['claim_id')`

jackrobison: These should really go in lbryum ([here](https://github.com/lbryio/lbryum/blob/f7fd95f54687c46911471f3d805a155209701a53/lbryum/commands.py#L1093) for the function that formats the response for `getvalueforname` (winning resolution) and [here](https://github.com/lbryio/lbryum/blob/f7fd95f54687c46911471f3d805a155209701a53/lbryum/commands.py#L994) for the rest)

jackrobison: The channel name can also be included in the permanent url (as requested by @kauffj) in lbryum if the signature is valid.

kaykurokawa:
I agree that additional formatting to claim , as is done from in this PR from here in line 737 : https://github.com/lbryio/lbry/blob/0b46d9295792de75e168987096fc326fcc120ebf/lbrynet/core/Wallet.py#L737 to line 745 should moved to lbryum as jack stated. We don't want multiple places that adjust the formatting. 

Perhaps as @kauffj stated in a meeting , it would be more logical if the one place it did the formatting was in lbrynet daemon, but as it stands now, this is not possible because lbrynet daemon does not talk to lbryum server directly which it would need to do in order to do the formatting on its own. 
, 
Perhaps this can be in a new PR that you work on next since its unrelated to the issue at hand.

kauffj:
I have no objections to where this is placed.

hackrush01:
Agreed on keeping this logic in lbryum right now.
Suggestion: We are already planning to do something along these lines but I'll get it out there. We should have lbryum, lbryum-server (and possibly lbrycrd?) return the same signature for all claims, with _only_ the claim information, i.e. no satellite information such as permanent URL etc. and that signature should be present in the lbryschema, similar to ClaimDict. All satellite information should be added to daemon or app level as required. This way we can also ensure standardized signatures across different repos(step towards #1008, #1006)

lbrynet/core/Wallet.py
eukreign: For even better forward compatibility (python 3) use the `.format()` string method:
```
claim['permanent_url'] = "{}#{}".format(claim["name"], claim["claim_id"])
```

lbrynet/core/Wallet.py
```diff
@@ -875,6 +878,8 @@ def get_claims_for_name(self, name):
                 decoded = smart_decode(claim['value'])
                 claim['value'] = decoded.claim_dict
                 claim['hex'] = decoded.serialized.encode('hex')
+                # lbry/issue#958
+                claim['permanent_url'] = claim["name"] + "#" + claim["claim_id"]
```
eukreign: Same as earlier, this is better:
```
claim['permanent_url'] = "{}#{}".format(claim["name"], claim["claim_id"])
```

lbrynet/core/Wallet.py
```diff
@@ -946,6 +951,9 @@ def claim_name(self, name, bid, metadata, certificate_id=None, claim_address=Non
             raise Exception(msg)
 
         claim = self._process_claim_out(claim)
+
+        # lbry/issue#958
+        claim['permanent_url'] = str(name) + "#" + claim["claim_id"]
```
eukreign: Hmm... is there a way to refactor all this so that the same code isn't duplicated three times?

In either case, should be using `.format()` like the others.

hackrush01:
Closed, logic ported to lbryum.

PR #1086, opened 2018-01-29
fixes for the --conf flag

ordex:
When starting a daemon with a different configuration,
the same should also be used by the cli tool.

Add the --conf flag to the cli command to allow using a
custom config file.

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -509,6 +509,10 @@ class Config(object):
                 converted[k] = v
         return converted
 
+    def initialize_post_conf_load(self):
+        settings.installation_id = settings.get_installation_id()
+        settings.node_id = settings.get_node_id()
+
     def load_conf_file_settings(self):
         if conf_file:
             path = conf_file
@@ -527,6 +531,9 @@ class Config(object):
             self._data[TYPE_PERSISTED].update(self._convert_conf_file_lists(decoded))
         except (IOError, OSError) as err:
             log.info('%s: Failed to update settings from %s', err, path)
+
+        #initialize members depending on config file
+        self.initialize_post_conf_load()
 
     def _fix_old_conf_file_settings(self, settings_dict):
         if 'API_INTERFACE' in settings_dict:
@@ -628,7 +635,5 @@ def initialize_settings(load_conf_file=True):
     if settings is None:
         settings = Config(FIXED_SETTINGS, ADJUSTABLE_SETTINGS,
                           environment=get_default_env())
-        settings.installation_id = settings.get_installation_id()
-        settings.node_id = settings.get_node_id()
         if load_conf_file:
             settings.load_conf_file_settings()
--- a/lbrynet/daemon/DaemonCLI.py
+++ b/lbrynet/daemon/DaemonCLI.py
@@ -38,8 +38,22 @@ def set_flag_vals(flag_names, parsed_args):
 
 
 def main():
-    if len(sys.argv[1:]):
-        method, args = sys.argv[1], sys.argv[2:]
+    argv = sys.argv[1:]
+
+    # check if a config file has been specified. If so, shift
+    # all the arguments so that the parsing can continue without
+    # noticing
+    if len(argv) and argv[0] == "--conf":
+        if len(argv) < 2:
+            print_error("No config file specified for --conf option")
+            print_help()
+            return
+
+        conf.conf_file = argv[1]
+        argv = argv[2:]
+
+    if len(argv):
+        method, args = argv[0], argv[1:]
     else:
         print_help()
         return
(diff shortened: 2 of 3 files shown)


ordex:
Without this flag it is not possible to interact with different daemons running on the same host while using different settings.

ordex:
changelog modified

ordex:
Renaming this PR as I am adding another fix that makes --conf behave saner.

lbrynet/daemon/DaemonCLI.py
```diff
@@ -38,8 +38,17 @@ def set_flag_vals(flag_names, parsed_args):
 
 
 def main():
-    if len(sys.argv[1:]):
-        method, args = sys.argv[1], sys.argv[2:]
+    argv = sys.argv[1:]
+
+    # check if a config file ha sbeen specified. If so, shift
```
kaykurokawa: spelling : check if a config file "has"

kaykurokawa:
Should adjust print_help() https://github.com/lbryio/lbry/blob/9460b57fcbe2a278f9f75ab91b86d1ed6bf46727/lbrynet/daemon/DaemonCLI.py#L184 to reflect changes

lbrynet/daemon/DaemonCLI.py
```diff
@@ -38,8 +38,17 @@ def set_flag_vals(flag_names, parsed_args):
...
+    # all the arguments so that the parsing can continue without
+    # noticing
+    if sys.argv[1] == "--conf":
```
lyoshenka: make sure you document that `--conf` must be the very first thing on the command line

ordex: I am documenting it as part of the print_help() usage example. Let me know if it's enough

ordex:
comments addressed. Thanks for the feedback!

lbrynet/daemon/DaemonCLI.py
```diff
@@ -38,8 +38,17 @@ def set_flag_vals(flag_names, parsed_args):
...
 def main():
-    if len(sys.argv[1:]):
-        method, args = sys.argv[1], sys.argv[2:]
+    argv = sys.argv[1:]
+
+    # check if a config file has been specified. If so, shift
+    # all the arguments so that the parsing can continue without
+    # noticing
+    if sys.argv[1] == "--conf":
+        conf.conf_file = sys.argv[2]
```
jackrobison: this should check len(sys.argv) > 2 first in order to return a more meaningful error message if the conf file wasn't provided than an `IndexError`

lbrynet/daemon/DaemonCLI.py
```diff
@@ -38,8 +38,17 @@ def set_flag_vals(flag_names, parsed_args):
...
 
 def main():
-    if len(sys.argv[1:]):
-        method, args = sys.argv[1], sys.argv[2:]
+    argv = sys.argv[1:]
+
+    # check if a config file has been specified. If so, shift
+    # all the arguments so that the parsing can continue without
+    # noticing
+    if sys.argv[1] == "--conf":
```
jackrobison: this `if` should be within `if len(argv):` below  since we only check for a `--conf` argument if there are arguments.

lbrynet/conf.py
```diff
@@ -624,7 +631,5 @@ def initialize_settings(load_conf_file=True):
     if settings is None:
         settings = Config(FIXED_SETTINGS, ADJUSTABLE_SETTINGS,
                           environment=get_default_env())
-        settings.installation_id = settings.get_installation_id()
-        settings.node_id = settings.get_node_id()
```
jackrobison: instead of adding `initialize_post_conf_load` you can just move these two lines to be below `settings.load_conf_file_settings()`

lbrynet/daemon/DaemonCLI.py
ordex: makes sense. Will fix!

lbrynet/daemon/DaemonCLI.py
ordex: ops, right. will take care of that

lbrynet/conf.py
ordex: right now there are at least two spots where initialize_post_conf_load() is invoked, therefore we should copy those two lines in both spots..I believed having a function could make the code slimmer.

On top of that, if in the future we should have more fields that require the config to be loaded first, we can simply add more line to that function while keeping the code compact.
Last thing: having them in a function with a proper name makes it clear "why" we are doing that (I know hat also a comment would work..but yeah).

Do you think I should still remove the function?

ordex:
@jackrobison first 2 comments addressed and pushed

ordex:
and rebased on top of latest master to get rid of the conflict

Issue #1093, opened 2018-01-30
connection to known DHT nodes attempted only once


ordex:
A DHT node attempts to connect to the configured `known DHT nodes` during DHT startup, however, if something fails the node won't try to reconnect again.

This leads to the node printing something like this:
`2018-01-30 17:20:23,350 WARNING  lbrynet.dht.node:596: This node doesnt know any other nodes`
and won't resolve by itself - it needs a restart.

This behaviour should be changed so that the node periodically tries to recontact the known DHT nodes if the first attempt failed.

tzarebczan:
Great find @ordex !

PR #1096, opened 2018-02-01
Re-join the DHT if the node has no active peers

ordex:
This PR addresses #1093 by periodically attempting to re-join the DHT if the node routing table is empty.
Check is performed every 60 seconds.

This fix required some refactoring, therefore I did split my work into several commits, hoping to make the review easier.

### DIFF
--- a/lbrynet/dht/node.py
+++ b/lbrynet/dht/node.py
@@ -132,17 +132,12 @@ class Node(object):
             self._listeningPort.stopListening()
         self.hash_watcher.stop()
 
-    @defer.inlineCallbacks
-    def joinNetwork(self, knownNodeAddresses=None):
-        """ Causes the Node to join the Kademlia network; normally, this
-        should be called before any other DHT operations.
-
-        @param knownNodeAddresses: A sequence of tuples containing IP address
-                                   information for existing nodes on the
-                                   Kademlia network, in the format:
-                                   C{(<ip address>, (udp port>)}
-        @type knownNodeAddresses: tuple
+    def startNetwork(self):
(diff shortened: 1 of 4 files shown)


ordex:
`lbrynet.tests.functional.test_misc.TestTransfer.test_lbry_transfer` test is failing in travis-ci with `sqlite3.OperationalError: disk I/O error`. Shouldn't be related to my PR.

lbrynet/dht/node.py
```diff
@@ -190,6 +201,12 @@ def printContacts(self, *args):
                 print "    %s:%i" % (contact.address, contact.port)
         print '=================================='
 
+    def hasContacts(self):
+        for i in range(len(self._routingTable._buckets)):
+            if len(self._routingTable._buckets[i]._contacts) > 0:
+                return True
+        return False
```
eukreign: Less complicated and maybe easier to read:
```python
    def hasContacts(self):
         for _ in self._routingTable._buckets:
             if self._routingTable._buckets[i]._contacts:
                 return True
         return False
```

ordex: Thanks! I overlooked that `list()` evaluates to False when empty.

About the external for: I copy/pasted from printContacts() :) maybe we should clean that too (and not only that)

ordex: change applied

ordex: I ended up using this simpler version:
```python
204     def hasContacts(self):
205         for bucket in self._routingTable._buckets:
206             if bucket._contacts:
207                 return True
208         return False
```

Issue #895, opened 2017-09-08
API call to blob_list with --uri paramater fails


tzarebczan:
## The Issue
When calling `blob_list` with `--uri=claimname` you get an ERROR: 'stream' response. See full stack below:

```
Here's the traceback for the error you encountered:
Traceback (most recent call last):
  File "site-packages\twisted\internet\defer.py", line 457, in callback

  File "site-packages\twisted\internet\defer.py", line 565, in _startRunCallbacks

  File "site-packages\twisted\internet\defer.py", line 651, in _runCallbacks

  File "site-packages\twisted\internet\defer.py", line 1355, in gotResult

--- <exception caught here> ---
  File "site-packages\twisted\internet\defer.py", line 1299, in _inlineCallbacks

  File "site-packages\lbrynet\daemon\Daemon.py", line 2536, in jsonrpc_blob_list

  File "site-packages\lbrynet\core\utils.py", line 125, in get_sd_hash

exceptions.KeyError: 'stream'

```
### Steps to reproduce
1. call `./lbrynet-cli blob_list --uri=one`

### Expected behavior
Should return a list of blobs

### Actual behavior
Error

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

KamilPiekutowski:
self.get_blobs_for_stream_hash does not get executed
although it is assigned to a callback here:
        d.addCallback(self.get_blobs_for_stream_hash)

lyoshenka:
fixed by https://github.com/lbryio/lbry/pull/1100

tzarebczan:
Reopening- https://github.com/lbryio/lbry/pull/1100#issuecomment-367078951

jackrobison:
fixed in https://github.com/lbryio/lbry/pull/1126

PR #1100, opened 2018-02-04
API call to blob_list with uri parameter now succeeds

atiaxi:
This is a fix for Issue #895 :

The shape of the claim data appears to have changed, from `['stream']['source']['source']`
to `['claim']['value']['stream']['source']['source']`.  I changed the path, and also wrote a utility function to safely delve along the path and return None if it doesn't work.  If that
happens, I log a warning so that if the shape changes again we know exactly where to look
and can easily see what the problem is from the logs.

I also found that attempting to get the blob_list via URI of an item that hadn't been downloaded
resulted in an error:

```
lbrynet.core.Error.NoSuchSDHash: 92b8aae7a901c56901fd5602c1f1acc0e63fb5492ef2a3cd5b9c631d92cab2e060e2a908baa922c24dee6c5229a98136

```

Whereas attempting to search via the `sd_hash` param performed normally; I've added the same
error handling functionality to the URI search that the hash search currently has.

### DIFF
--- a/lbrynet/core/utils.py
+++ b/lbrynet/core/utils.py
@@ -134,8 +134,31 @@ def get_sd_hash(stream_info):
         return None
     if isinstance(stream_info, ClaimDict):
         return stream_info.source_hash
-    return stream_info['stream']['source']['source']
+    path = ['claim', 'value', 'stream', 'source', 'source']
+    result = safe_dict_descend(stream_info, *path)
+    if not result:
+        log.warn("Unable to get sd_hash via path %s" % path)
+    return result
 
 
 def json_dumps_pretty(obj, **kwargs):
     return json.dumps(obj, sort_keys=True, indent=2, separators=(',', ': '), **kwargs)
+
+
+def safe_dict_descend(source, *path):
+    """
+    For when you want to do something like
+    stream_info['claim']['value']['stream']['source']['source']"
+    but don't trust that every last one of those keys exists
+    """
+    cur_source = source
+    for path_entry in path:
+        try:
+            if path_entry not in cur_source:
+                return None
+        except TypeError:
+            # This happens if we try to keep going along a path that isn't
+            # a dictionary (see e.g. test_safe_dict_descend_typeerror)
+            return None
+        cur_source = cur_source[path_entry]
+    return cur_source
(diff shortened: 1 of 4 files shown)


lbrynet/core/utils.py
```diff
@@ -134,8 +134,31 @@ def get_sd_hash(stream_info):
         return None
     if isinstance(stream_info, ClaimDict):
         return stream_info.source_hash
-    return stream_info['stream']['source']['source']
+    path = ['claim', 'value', 'stream', 'source', 'source']
+    result = safe_dict_descend(stream_info, *path)
```
eukreign: Would this be sufficient without introducing a lot of extra code?
```
stream_info.get('claim', {}).get('value', {}).get('stream', {}).get('source', {}).get('source')
```

atiaxi: After poking through the codebase it doesn't look like there are a lot of deep-dives like this, so `safe_dict_descend` is probably overkill.  While the `get` chain won't catch the case of e.g. a non-dict value ending up somewhere in the path, I don't think that's terribly likely and even if it was we'll still get a stacktrace.

I've written up a few tests to cover `get_sd_hash` to make up for the tests I'm taking out in this commit :)

tzarebczan:
@jackrobison / @eukreign - can you give this a try? I'm on RC35 and I see the below when running `lbrynet-cli blob_list one` / `lbrynet-cli blob_list --uri=one`

```
2018-02-20 13:44:37,608 WARNING  lbrynet.daemon.auth.server:228: error processing api request: 'CryptBlobInfo' object is not iterable
traceback:   File "site-packages\twisted\internet\defer.py", line 1297, in _inlineCallbacks

  File "site-packages\twisted\python\failure.py", line 389, in throwExceptionIntoGenerator

  File "site-packages\lbrynet\daemon\Daemon.py", line 2920, in jsonrpc_blob_list

  File "site-packages\twisted\internet\defer.py", line 651, in _runCallbacks

  File "site-packages\lbrynet\daemon\Daemon.py", line 961, in _get_blobs

  File "site-packages\lbrynet\daemon\Daemon.py", line 956, in _iter_blobs

exceptions.TypeError: 'CryptBlobInfo' object is not iterable
```

Issue #748, opened 2017-06-30
Cannot update claims larger than wallet balance


kauffj:
Originally lbry-app [#117](https://github.com/lbryio/lbry-app/issues/117)

tzarebczan:
An option for the user during claim update would make sense if it's technically possible.

kauffj:
An option would exist at the app/UI level, not the API level.

tzarebczan:
This is a pretty big UX issue at the moment, maybe we can bump the priority @kauffj 

We would need @kaykurokawa 's fixes from #747 which would allow the daemon to realize that it can use the original deposit amount during an update. So really only any difference in LBC (in case they are increasing the bid) needs to be in your balance (plus tx fees obviously).

tzarebczan:
@lyoshenka anyway we can bump this up the queue? Sounds like it would need some lbryum changes too.

hackrush01:
I am assigning it to myself currently, as I think I can solve it.

tzarebczan:
Thanks @hackrush01 . This might be related, or at least part of the same code, please keep it in mind: https://github.com/lbryio/lbry/issues/855

PR #1108, opened 2018-02-09
Include existing claim amount in the max biddable amount for a publish

hackrush01:
Fixes updating a claim when update's bid amount is greater than wallet balance but it can use the claim's original amount to complete the transaction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,18 +26,8 @@ at anytime.
   * `get` failing with a non-useful error message when given a uri for a channel claim
   * exception checking in several wallet unit tests
   * daemon not erring properly for non-numeric values being passed to the `bid` parameter for the `publish` method
-  *
-  * Fixed unnecessarily verbose exchange rate error (https://github.com/lbryio/lbry/issues/984)
-  * Merged two separate dht test folders into one
-  * Fixed value error due to a race condition when saving to the claim cache (https://github.com/lbryio/lbry/issues/1013)
-  * Fixed being unable to re-download updated content (#951)
-  * Fixed sending error messages for failed api requests
-  * Fixed the file manager startup being slow when handling thousands of files
-  * Fixed handling decryption error for blobs encrypted with an invalid key
-  * Fixed handling stream with no data blob (https://github.com/lbryio/lbry/issues/905)
-  * Fixed fetching the external ip
-  * Fixed API call to blob_list with --uri parameter (https://github.com/lbryio/lbry/issues/895)
-  * Fixed publish command to allow updating claims with bid amount higher than wallet balance(by spending the claimtrietx coin) (https://github.com/lbryio/lbry/issues/748)
+  * `publish` command to allow updating claims with a `bid` amount higher than the wallet balance, so long as the amount is less than the wallet balance plus the bid amount of the claim being updated (https://github.com/lbryio/lbry/issues/748)
+  * 
 
 ### Deprecated
   * `channel_list_mine`, replaced with `channel_list`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,17 @@ at anytime.
   * exception checking in several wallet unit tests
   * daemon not erring properly for non-numeric values being passed to the `bid` parameter for the `publish` method
   *
+  * Fixed unnecessarily verbose exchange rate error (https://github.com/lbryio/lbry/issues/984)
+  * Merged two separate dht test folders into one
+  * Fixed value error due to a race condition when saving to the claim cache (https://github.com/lbryio/lbry/issues/1013)
+  * Fixed being unable to re-download updated content (#951)
+  * Fixed sending error messages for failed api requests
+  * Fixed the file manager startup being slow when handling thousands of files
+  * Fixed handling decryption error for blobs encrypted with an invalid key
+  * Fixed handling stream with no data blob (https://github.com/lbryio/lbry/issues/905)
+  * Fixed fetching the external ip
+  * Fixed API call to blob_list with --uri parameter (https://github.com/lbryio/lbry/issues/895)
+  * Fixed publish command to allow updating claims with bid amount higher than wallet balance(by spending the claimtrietx coin) (https://github.com/lbryio/lbry/issues/748)
 
 ### Deprecated
   * `channel_list_mine`, replaced with `channel_list`
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -537,9 +537,6 @@ class Wallet(object):
         decoded = ClaimDict.load_dict(metadata)
         serialized = decoded.serialized
 
-        if self.get_balance() < Decimal(bid):
-            raise InsufficientFundsError()
-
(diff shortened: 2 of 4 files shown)


hackrush01:
Forgot changelog.

lbrynet/daemon/Daemon.py
```diff
@@ -715,8 +715,6 @@ def _publish_stream(self, name, bid, claim_dict, file_path=None, certificate_id=
         publisher = Publisher(self.session, self.lbry_file_manager, self.session.wallet,
                               certificate_id)
         parse_lbry_uri(name)
-        if bid <= 0.0:
```
jackrobison: The reason this is here is because otherwise we have to possibly make a stream before lbryum would tell us something is wrong

lbrynet/daemon/Daemon.py
```diff
@@ -1969,10 +1967,6 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
         if bid <= 0.0:
             raise Exception("Invalid bid")
 
-        if bid >= self.session.wallet.get_balance():
-            raise InsufficientFundsError('Insufficient funds. ' \
```
jackrobison: Same reason as the other bid check

eukreign: Also, this should be formatted as something like:
```python
raise InsufficientFundsError(
  'Insufficient funds. Make sure you have enough LBC to deposit.'
)
```
Or perhaps, if the message is constant it should just be built-in to the InsufficientFundsError. How many different messages do we need for this exception?

lbrynet/tests/unit/core/test_Wallet.py
```diff
@@ -217,7 +217,7 @@ def update_balance():
         d.addCallback(lambda _: self.assertEqual(5, wallet.get_balance()))
         d.addCallback(lambda _: wallet.reserve_points('testid', 2))
         d.addCallback(lambda _: wallet.claim_name('test', 4, test_claim_dict))
-        self.assertFailure(d, InsufficientFundsError)
+        self.assertFailure(d, Exception)
```
eukreign: Testing for `Exception` seems a bit dangerous. This would pass the test on any exception, even if it has nothing to do with the error being tested.

hackrush01: This is what is returned from lbryum I guess, maybe I need to look at the lbryum error messages too.
Also would it be bad practice if I check the error message string and then decide which exception to raise?? Something like
```python
if claim["msg"] == "Insufficient funds":
  Raise InsufficientFundsError()
```
here claim is the JSON signature returned from `lbryum`

hackrush01:
@jackrobison
`lbrynet.core.Wallet.update_balance` only gets the updated wallet balance, but the issue I'm trying to fix is this #748.

For e.g. I have:
10 LBC in wallet
Claim `qwerty` with bid 20 LBC
I should be able to update `qwerty`'s bid to 25 LBC, which would require spending the claimtrietx input coin. This already has support in lbryum, just removing the checks from the daemon for it.

lbrynet/tests/unit/core/test_Wallet.py
eukreign: Checking for an error condition in a response from lbryum and throwing an appropriate exception makes sense to me. Although lbryum is not consistent about the insufficient funds errors, maybe we need to add error codes first as @kauffj suggested in another issue.

hackrush01: Yes, the function for checking the above stuff in lbryum is a bit convoluted, with many abstract returns. Maybe I'll check for error message in daemon and raise specific Exceptions for possible cases, and a different but non-generic and descriptive errors for the rest of the cases.

jackrobison:
Perhaps you could add a function to the lbryum wallet to return the maximum spendable amount for an update. Something like:

```
def get_maximum_claim_amount(self, claim_id=None):
    return confirmed_balance + current_claim_amount or 0
```

Then have lbrynet `publish` check that the `bid` is less than this amount instead of the balance.

lbrynet/core/Wallet.py
```diff
@@ -975,6 +976,11 @@ def _update_balance(self):
             lambda result: Decimal(result['confirmed']) + Decimal(result.get('unconfirmed', 0.0)))
         return d
 
+    @defer.inlineCallbacks
+    def get_max_usable_balance_for_claim(self, claim_name):
+        amt = yield self._run_cmd_as_defer_to_thread('get_max_spendable_amt_for_claim', claim_name)
+        defer.returnValue(amt)
```
eukreign: You can simplify this to just:
```python
def get_max_usable_balance_for_claim(self, claim_name):
    return self._run_cmd_as_defer_to_thread('get_max_spendable_amt_for_claim', claim_name)
```

`@defer.inlineCallbacks` turns a method into something that returns a `Deferred` but `_run_cmd_as_defer_to_thread()` already returns a `Deferred`, so you can skip all the indirection and just return the result of `_run_cmd_as_defer_to_thread()`.

lbrynet/daemon/Daemon.py
```diff
@@ -2004,9 +2002,11 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
...
-                                         'Make sure you have enough LBC to deposit')
+        amt = yield self.session.wallet.get_max_usable_balance_for_claim(name)
+        if bid > amt:
+            raise InsufficientFundsError(
+                "Please lower the bid value, the max amount you can specify for this claim is {}"
```
eukreign: In user facing messages we should spell out 'max' as maximum. If this text doesn't make it to the user then never mind.

CHANGELOG.md
```diff
@@ -26,7 +26,8 @@ at anytime.
   * `get` failing with a non-useful error message when given a uri for a channel claim
   * exception checking in several wallet unit tests
   * daemon not erring properly for non-numeric values being passed to the `bid` parameter for the `publish` method
-  *
+  * publish command to allow updating claims with bid amount higher than wallet balance(by spending the claimtrietx coin) (https://github.com/lbryio/lbry/issues/748)
```
jackrobison: Commands, classes, functions, and argument names should be decorated with code ( ``` ` ```) tags. In this entry, `publish` and `bid` should be decorated.

jackrobison: >higher than wallet balance

How much higher? Could be rephrased: `with a bid amount higher than the wallet balance, so long as the amount is less than the wallet balance plus the bid amount of the claim being updated`

lbrynet/core/Wallet.py
```diff
@@ -975,6 +976,9 @@ def _update_balance(self):
             lambda result: Decimal(result['confirmed']) + Decimal(result.get('unconfirmed', 0.0)))
         return d
 
+    def get_max_usable_balance_for_claim(self, claim_name):
```
jackrobison: Please add a matching dummy method in `Wallet` too

hackrush01: Done!

hackrush01:
@jackrobison I made the required changes. Please check.

PR #1099, opened 2018-02-03
Add exception for non-float bids when pubishing

Jeremy1026:
When using the API, if you provide `"bid": .000001` you'll receive an exception for improperly formatted JSON, but if you use `"bid": ".000001"`you'll receive and exception stating "Insufficient funds. Make sure you have enough LBC to deposit". This is a (potentially) inaccurate message. The problem is that the API is comparing the bid,which was passed as a string, to the wallet balance, which is a float. This pull request changes that exception to give the developer more correct reasoning to why the API call failed.

PR #1110, opened 2018-02-10
Jeremy1026 add new exception

hackrush01:
For #1099

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -2001,11 +2001,11 @@ class Daemon(AuthJSONRPCServer):
         except (TypeError, URIParseError):
             raise Exception("Invalid name given to publish")
 
-        if not isinstance(bid, float):
-            raise Exception("Bid must be a float")
+        if not isinstance(bid, (float, int)):
+            raise TypeError("Bid must be a float or an integer.")
 
         if bid <= 0.0:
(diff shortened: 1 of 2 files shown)


lbrynet/daemon/Daemon.py
```diff
@@ -1966,6 +1966,9 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
         except (TypeError, URIParseError):
             raise Exception("Invalid name given to publish")
 
+        if not isinstance(bid, float):
+            raise Exception("Bid must be a float")
```
eukreign: Don't invent new exceptions unless it's somehow specific to the app. This particular exception looks like a TypeError, so just throw a TypeError. Or, pick something else more appropriate from here: https://docs.python.org/2/library/exceptions.html

eukreign: Sorry, this should be `ValueError`, not `TypeError`.

eukreign: Investigated this some more, `TypeError` is the correct exception:

`TypeError` - raised when a variable is of the wrong type. So, if you are expecting an int and it's a float, that's a **type** issue.

`ValueError` - is when you expect a string that you then want to convert to a float and it's an integer, then that's a value error. because string is correct **type**, but the **value** of the string is not valid.

PR #1115, opened 2018-02-14
Updated dependencies

hackrush01:
Added Cython and unqlite as dependencies.


jackrobison:
These are only needed to run the tests - I don't think they should be added to the module requirements since users of lbrynet do not need them, nor should they be bundled into daemon releases. `mock` is another testing-only requirement.

It would be better if the tests were updated to not use `unqlite` (they should use `sqlite3` instead). Since cython is only needed for unqlite it could be dropped as a testing requirement as well.

If there are still requirements only needed for testing, it would be better to put them in a separate `testing_requirements.txt` file.

hackrush01:
@jackrobison Aha, understood. Instead of making a new `testing_requiements.txt` file, it would be enough to just mention devs specific requirement somewhere. I guess...

PR #1132, opened 2018-03-06
test that docs are up to date in travis

lyoshenka:


hackrush01:
One solution maybe to install older version of tornado manually if it doesn't conflict with any other module, as older python support was discontinued in tornado 4.4(https://allmychanges.com/p/python/tornado/)

tzarebczan:
@lyoshenka / @hackrush01 what's the status on this?

tzarebczan:
@IGassmann what's next here?

IGassmann:
@tzarebczan I have some progress made, but it's on standby due to a more important issue.

IGassmann:
CI build fails if `mkdocs build` produces an output. @lyoshenka is this the expected behavior?

lyoshenka:
I don't think that's intentional, or I'm misunderstanding.

The idea is that if `mkdocs build` causes changes (as reported by git), then it should have been run by the dev, and the updated docs should have been committed.

If you can get travis to build and commit the docs itself, that would probably be better.

tzarebczan:
@IGassmann what's the next step here? Is it possible to make travis to commit the docs? If not, how do we get around it failing?

lyoshenka:
I'm passing this on to @eukreign since he's working on Travis now, and there's an impending switch to py3.

eukreign:
I would like to prioritize this immediately after 0.30 release.

lyoshenka:
@eukreign 30 is out. reassigning to you

PR #1137, opened 2018-03-07
Error cleanly when claiming a new channel with exact or higher amount than balance

### LINKED ISSUE #1107: Claiming with exact amount in wallet produces confusing result
## The Issue
This issue is related to https://github.com/lbryio/lbry/issues/855 and https://github.com/lbryio/lbry/issues/1045, but wanted to log it separately as the daemon is returning a confusing message - `Claimed a new channel` and then  `'success': False`.  On the app side, the channel is created but then a different error occurs when you try publishing because the channel does not exist. Will follow up with app side issue to address this while the daemon side is being worked on. 

When creating a channel with the exact amount in your wallet (not accounting for fees), you get this:
lbrynet.daemon.Daemon:1739: Claimed a new channel! Result: {'reason': 'Not enough funds', 'success': False} 

If you enter a really small amount like 0.00000001, you get: lbrynet.daemon.Daemon:1739: Claimed a new channel! Result: {'reason': 'Amount must be greater than 0', 'success': False}

### Steps to reproduce
1. Claim a channel for an amount that's exactly in your wallet
2. Claim a channel for 0.00000001 LBC
3. Check logs

### Expected behaviour
Daemon should not say the channel was claimed, but instead provide the correct warning/error to the app/user

### Actual behaviour
Channel looks like its claimed, but is not

## System Configuration

 

- LBRY Daemon version: 0.18.2
- LBRY App version: 0.20.0
- LBRY Installation ID:
- Operating system: Windows

hackrush01:
fixes #1107

### DIFF
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -503,8 +503,8 @@ class Wallet(object):
         log.info("Preparing to make certificate claim for %s", channel_name)
         channel_claim = yield self._claim_certificate(parsed_channel_name.name, amount)
         if not channel_claim['success']:
-            log.error(channel_claim)
-            msg = 'Claim to name {} failed: {}'.format(channel_name, channel_claim['reason'])
+            msg = 'Claiming of {} failed: {}'.format(channel_name, channel_claim['reason'])
+            log.error(msg)
             raise Exception(msg)
         yield self.save_claim(self._get_temp_claim_info(channel_claim, channel_name, amount))
         defer.returnValue(channel_claim)
@@ -563,8 +563,8 @@ class Wallet(object):
                                             bid, certificate_id, claim_address, change_address)
 
         if not claim['success']:
-            log.error(claim)
-            msg = 'Claim to name {} failed: {}'.format(name, claim['reason'])
+            msg = 'Claiming of name {} failed: {}'.format(name, claim['reason'])
+            log.error(msg)
             raise Exception(msg)
         claim = self._process_claim_out(claim)
         yield self.storage.save_claim(self._get_temp_claim_info(claim, name, bid), smart_decode(claim['value']))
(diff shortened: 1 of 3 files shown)


lbrynet/core/Wallet.py
```diff
@@ -502,6 +502,10 @@ def claim_new_channel(self, channel_name, amount):
             raise Exception("New channel claim should have no fields other than name")
         log.info("Preparing to make certificate claim for %s", channel_name)
         channel_claim = yield self._claim_certificate(parsed_channel_name.name, amount)
+        if not channel_claim['success']:
+            log.error(channel_claim)
```
eukreign: Is there an error message or something inside channel_claim that could be logged? Logging a dictionary seems strange.

lbrynet/daemon/Daemon.py
```diff
@@ -2028,11 +2032,11 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
...
             raise ValueError("Bid value must be greater than 0.0")
 
-        amt = yield self.session.wallet.get_max_usable_balance_for_claim(name)
-        if bid > amt:
+        balance = yield self.session.wallet.get_max_usable_balance_for_claim(name)
+        if bid > balance:
             raise InsufficientFundsError(
                 "Please lower the bid value, the maximum amount you can specify for this claim is {}"
-                .format(amt - MAX_UPDATE_FEE_ESTIMATE))
+                .format(balance - MAX_UPDATE_FEE_ESTIMATE))
```
eukreign: How come the condition check is `bid > balance` but in the error message it says that the max is `balance - MAX_UPDATE_FEE_ESTIMATE`. Should the condition be updated to be `bid > (balance - MAX_UPDATE_FEE_ESTIMATE)`?

hackrush01: Since this is an estimate of the max fee that can be used, I put it in the message because one could technically update for a bid amount higher than that.
Should I instead put it in the condition check itself? This will _probably_ be changed after [this](https://github.com/lbryio/lbryum/pull/234) get merged though.

PR #1155, opened 2018-03-26
Twisted Based Wallet Implementation

eukreign:
### For Reviewers

This PR is split across three repositories,
1) a brand new project called [torba](https://github.com/lbryio/torba) which is a generic implementation of bitcoin based wallets (integration tests run against bitcoinsegwit and eventually bitcoincash and others);
2) the custom LBRY extensions ontop of torba are part of this PR and found in [lbrynet/wallet](https://github.com/lbryio/lbry/tree/lbryum-refactor/lbrynet/wallet);
3) [orchstr8](https://github.com/lbryio/orchstr8) integration testing framework that orchestrates starting the blockchain binary (`-regtest`), electrumx (lbryumx) and the torba (lbrynet.wallet) and facilitating their interaction in a controlled environment.

This is a very major change to the daemon and should be evaluated carefully for security, performance and reliability before making it into a release. Furthermore, this PR also ports lbrynet to Python 3 and is no longer compatible with Python 2. **(see explanation at end)

Here are the main points of interest that should be reviewed:

### Coin (UTXO) Selection Algorithm

Torba implements the same coin selection algorithm as latest C++ Bitcoin. Please review the algorithm and tests for completeness and correctness:

Torba (impl): https://github.com/lbryio/torba/blob/master/torba/coinselection.py
Bitcoin (impl): https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp

Torba (tests): https://github.com/lbryio/torba/blob/master/tests/unit/test_coinselection.py
Bitcoin (tests): https://github.com/bitcoin/bitcoin/blob/master/src/wallet/test/coinselector_tests.cpp

Whitepaper: https://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf

### SQLite Storage

Old wallet stored things either in JSON in the wallet file or in memory in Python dictoinaries and lists when running. This has been completely replaced by an SQLite storage solution. Please review the base torba DDL and then the LBRY extension adding extra claim related attributes to the `txo` table:

Torba: https://github.com/lbryio/torba/blob/master/torba/basedatabase.py
LBRY: https://github.com/lbryio/lbry/blob/lbryum-refactor/lbrynet/wallet/database.py

Getting the DDL right from the beginning is a good idea, because updating code is cheap but updating DDL is time consuming for us (writing migrations) and the users (waiting for migrations to finish running).

### Initial Sync & Subsequent Starts

Faster wallet operation is the main reason for this PR; so it's important that the header and TX synchronization code paths fast and correct. Please review the following areas:

https://github.com/lbryio/torba/blob/d4107600a5882fd9d8ccd13a0d3694c96d2d7c05/torba/baseledger.py#L226-L249
`update_headers` is called on startup (or if you were offline for too long and have more than one header to catchup) while `process_header` is called on regular sequential updates from electrumx (but switched to `update_headers` if it determines we're behind more than one).

https://github.com/lbryio/torba/blob/d4107600a5882fd9d8ccd13a0d3694c96d2d7c05/torba/baseledger.py#L251-L331
`update_account` is called on startup to catchup with blockchain or recreate wallet from scratch, this is done by generating addresses, sync'ing them, then generating more addresses, sync'ing them, etc, this repeats until all addresses have been generated (stopping when address gap is reached).
after the initial startup most of the normal process is between `process_status` which receives events from electrumx and calls `update_history` which then proceeds to process the history for an address.
one complication to keep in mind with `update_history` is that electrumx will send duplicate updates for the same TX if it involves two of your addresses (common case when you pay and have change, or when you pay yourself, etc), it's important to avoid race conditions in those cases (done with `self._transaction_processing_locks`).

### Simple Payment Verification (SPV)

Torba is called an SPV wallet because it verifies that TXs are in the blockchain without having a copy of the blockchain. This verification is predicated on having correct headers (or at least the headers and the TX should come from different parties that were unlikely to have conspired with each other).

SPV works by organizing TXs in a block into merkle trees, then headers contain the root hash of this merkle tree and you can separately, for each TX, request all of the hashes between your TX and the root hash. Then you just traverse up the tree from your TX hashing every two complementary nodes until you reach the root hash, if the TX is valid the root hash in the header should mach the root hash you calculated. Done.

Here is the implementation, please verify for correctness:
https://github.com/lbryio/torba/blob/d4107600a5882fd9d8ccd13a0d3694c96d2d7c05/torba/baseledger.py#L189-L198

This feature is currently covered by integration tests and does not have dedicated unit tests.

### Bitcoin Script

Torba implements a general purpose extensible bitcoin script parser. LBRY extends this to add our 3 custom op codes.

Torba (impl): https://github.com/lbryio/torba/blob/master/torba/basescript.py
LBRY (impl): https://github.com/lbryio/lbry/blob/lbryum-refactor/lbrynet/wallet/script.py

Torba (tests): https://github.com/lbryio/torba/blob/master/tests/unit/test_script.py
LBRY (tests): https://github.com/lbryio/lbry/blob/lbryum-refactor/tests/unit/wallet/test_script.py

### Creating Transactions

Can't forget this one, the whole point we're all here :-D

There is a special `create` constructor on `Transaction` class that will accept any inputs/outputs and then try to balance the two sides of the equation until it is correct (adding necessary inputs/outputs and adding change for excessive inputs). On `lbrynet` side there are convenient wrappers around it to make it even easier to create LBRY specific claim transactions.

Torba: https://github.com/lbryio/torba/blob/d4107600a5882fd9d8ccd13a0d3694c96d2d7c05/torba/basetransaction.py#L380-L428
LBRY: https://github.com/lbryio/lbry/blob/lbryum-refactor/lbrynet/wallet/transaction.py

### All The Things

Above is a list of some of the critical aspects that should get some review before we go to releases. There is a lot more code than outlined above, if you feel motivated please review all of it and provide feedback. I would greatly appreciate it if people find issues that we can fix now before the code goes live! Thanks!

Essentially this means all of `torba`: https://github.com/lbryio/torba
Everything in `lbrynet.wallet` directory: https://github.com/lbryio/lbry/tree/lbryum-refactor/lbrynet/wallet

### Try Me

You can downoad automated builds of the new daemon here:
http://build.lbry.io/daemon/

** Python 3 port was done because 1) https://pythonclock.org  2) torba's testing framework [orchstr8](https://github.com/lbryio/orchstr8) loads electrumx ([lbryumx](https://github.com/lbryio/lbryumx)) and the wallet into a single Python process using the same asyncio reactor which makes debugging very nice (when you hit breakpoint, both server and client freeze), electrumx is Python 3 only, therefore torba had to straddle Python2/Python3 support to work in both `orchstr8` and in `lbrynet`, keeping this compatibility was a real time sink 3) Python 3 is significantly faster (thus new daemon should have perceptible speed increase) and stricter / failing faster (therefore we should hit bugs earlier in testing).

---

### Progress

This is a pure twisted wallet implementation for lbry providing increased stability and performance; completely replaces `lbryum`. Still a work in progress, **do not merge**.

Depends on this pull request in lbryum-server: https://github.com/lbryio/lbryum-server/pull/64

Each leaf task below is 5pt (about a day of work).

Core Functionality I1

- [X] groking existing thread/plain socket based architecture, considering how to refactor it
  - [X] lbry portion
  - [X] lbryum portion
- [x] header sync
  - [x] basic implementation / initial pass
  - [x] match current implementation + integration and/or unit tests (whichever ones don't get done as part of TDD)
- [x] subscribe to get address notifications
- [X] creating transactions
  - [x] refactor Transaction class
  - [x] round trip-able script parser / generator
    - [x] input scripts (pub key, script hash / multi-sig)
    - [x] output scripts (pub key, scripts hash) x (claim name, claim support, claim update)
  - [x] refactor Transaction class II
    - [x] use new script parser / generator
    - [x] coin chooser
    - [x] signing (single-sig and multi-sig)
- [x] move any code still being used from `lbryum` into `lbrynet/wallet`

Core Functionality I2
- [x] expose APIs to create/update/support name claims
- [x] use SQLiteStorage for storing transaction history
- [x] base wallet as separate module
- [x] studying https://cardanodocs.com/technical/formal-specification-for-a-cardano-wallet/

Core Functionality I3
- [x] fix orchstr8 to run with lbryumx
- [x] release generic parts of wallet as https://github.com/lbryio/torba
- [x] add regtest support to lbry-app (filed issue and provided branch with problematic code, https://github.com/lbryio/lbry-app/issues/1526)
- [x] works with lbry-app (ALPHA release, pending lbrycrd and lbry-app regtest support)

Separate Issues Created After I3
- [x] store almost everything in SQLite (https://github.com/lbryio/lbry/issues/1242)
- [x] SPV (https://github.com/lbryio/lbry/issues/1221)
- [x] save and use channel certificates (https://github.com/lbryio/lbry/issues/1223)
- [x] locking of addresses being used in transaction until the tx is written to block chain (https://github.com/lbryio/lbry/issues/1222)
- [x] migrating wallets (https://github.com/lbryio/lbry/issues/1224)
- [x] verify lbryumx responses (https://github.com/lbryio/lbry/issues/1272)
- [ ] plain API calls (such as resolve) (https://github.com/lbryio/lbry/issues/1220)
- [ ] performance test cases (https://github.com/lbryio/lbry/issues/1219)
- [ ] regression tests:
  - [ ] Allow update of claims even if they don't exist in file_list (https://github.com/lbryio/lbry/issues/1165)
  - [ ] UTXOs remain stuck unconfirmed (https://github.com/lbryio/lbry/issues/1192)


lyoshenka:
lex created separate issues for everything that's outstanding

eukreign:
errors reported by @jackrobison in slack regarding the `daemon-components` branch which is now been merged into this PR.

when hosting:
```
2018-07-05 16:02:25,313 WARNING  lbrynet.core.server.ServerRequestHandler:128: An error occurred handling a request. Error: 'Logger' object has no attribute 'trace'
Unhandled error in Deferred:
2018-07-05 16:02:25,313 CRITICAL twisted:154: Unhandled error in Deferred:

2018-07-05 16:02:25,314 CRITICAL twisted:154: 
  File "/home/jack/.virtualenvs/superfresh/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 651, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/jack/code/lbry/lbrynet/core/server/ServerRequestHandler.py", line 166, in send_response
    self.send_response(response)
  File "/home/jack/code/lbry/lbrynet/core/server/ServerRequestHandler.py", line 141, in send_response
    self._produce_more()
  File "/home/jack/code/lbry/lbrynet/core/server/ServerRequestHandler.py", line 61, in _produce_more
    log.trace("writing %s bytes to the client", len(chunk))
AttributeError: 'Logger' object has no attribute 'trace'
2018-07-05 16:02:32,256 WARNING  lbrynet.core.server.ServerRequestHandler:128: An error occurred handling a request. Error: 'Logger' object has no attribute 'trace'
Unhandled error in Deferred:
```

^ logger not being set up somewhere

on shutting down before a component has started:
```2018-07-05 16:33:37,376 CRITICAL twisted:154: 
  File "/home/jack/.virtualenvs/superfresh/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/home/jack/.virtualenvs/superfresh/local/lib/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/jack/code/lbry/lbrynet/daemon/Component.py", line 63, in _stop
    raise err
AttributeError: 'NoneType' object has no attribute 'stop'
```

lbrynet/wallet/manager.py
```diff
@@ -0,0 +1,350 @@
...
+            raise Exception("Invalid channel name")
+        if amount <= 0:
+            raise Exception("Invalid amount")
+        account = self.default_account
+        address = yield account.receiving.get_or_create_usable_address()
+        cert, key = generate_certificate()
+        tx = yield Transaction.claim(channel_name, cert, amount, address, [account], account)
+        yield account.ledger.broadcast(tx)
+        account.add_certificate_private_key(tx.outputs[0].ref, key.decode())
+        # TODO: release reserved tx outputs in case anything fails by this point
```
BrannonKing: Still functions without this TODO? (and others in this file)

eukreign:
@BrannonKing can you be specific, which `encode`/`decode` calls are you referring to?

Most of the `encode`/`decode` are related to python 3 because everything is unicode by default for example, when you load JSON data it is only ever unicode (even if you recieved it as bytes from file or over network), strings coming out of sqlite are also unicode... BUT... most hashing and blockchain related processing is in bytes. Therefore a majority of the encode/decode calls in the wallet is just shuttling data from bytes to unicode and back depending on where the data is going or where it's coming from.

lbrynet/wallet/__init__.py
```diff
@@ -0,0 +1,9 @@
+__node_daemon__ = 'lbrycrdd'
+__node_cli__ = 'lbrycrd-cli'
+__node_bin__ = ''
+__node_url__ = (
+    'https://github.com/lbryio/lbrycrd/releases/download/v0.12.2.1/lbrycrd-linux.zip'
```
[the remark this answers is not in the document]
eukreign: What is painful about it? Can you be more specific? Also, assuming it's painful, how would you make it less painful?

lbrynet/core/StreamDescriptor.py
```diff
@@ -365,7 +375,7 @@ def get_stream_hash(hex_stream_name, key, hex_suggested_file_name, blob_infos):
 
 def verify_hex(text, field_name):
     for c in text:
-        if c not in '0123456789abcdef':
+        if c not in b'0123456789abcdef':
```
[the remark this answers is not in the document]
eukreign: that just makes the string be `bytes` instead of `unicode`. In Python 3 strings are unicode by default, so to get the original behavior it had to be converted to `bytes` explicitely.

BrannonKing:
It looks like hashlib requires a "bytes" object; as long as there is some kind of error for a missing encode/decode I think that will be sufficient.

eukreign:
@BrannonKing In Python 3 everything is pretty strict when it comes to unicode vs bytes. Things crash early on if you pass a unicode string where a bytes string was expected.

lbrynet/dht/error.py
```diff
@@ -1,10 +1,10 @@
 import binascii
-import exceptions
+#import exceptions
 
 # this is a dict of {"exceptions.<exception class name>": exception class} items used to raise
 # remote built-in exceptions locally
 BUILTIN_EXCEPTIONS = {
-    "exceptions.%s" % e: getattr(exceptions, e) for e in dir(exceptions) if not e.startswith("_")
+#    "exceptions.%s" % e: getattr(exceptions, e) for e in dir(exceptions) if not e.startswith("_")
```
jackrobison: why disable this?

lbrynet/dht/node.py
```diff
@@ -155,10 +149,17 @@ def __init__(self, node_id=None, udpPort=4000, dataStore=None,
...
+    #    log.warning("unclean shutdown of the dht node")
+    #    if hasattr(self, "_listeningPort") and self._listeningPort is not None:
+    #        self._listeningPort.stopListening()
+
+    def __str__(self):
+        return '<%s.%s object; ID: %s, IP address: %s, UDP port: %d>' % (
+            self.__module__, self.__class__.__name__, binascii.hexlify(self.node_id), self.externalIP, self.port)
+
+    def __hash__(self):
+        return int(binascii.hexlify(self.node_id), 16)
```
jackrobison: why this, and why not the node id bytes?

lbrynet/dht/protocol.py
```diff
@@ -368,7 +371,7 @@ def _write(self, txData, address):
         if self.transport:
             try:
                 self.transport.write(txData, address)
-            except socket.error as err:
+            except OSError as err:
```
jackrobison: are socket.errors now OSErrors?

lbrynet/daemon/Publisher.py
```diff
@@ -56,28 +56,22 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
                 log.info("Removed old stream for claim update: %s", lbry_file.stream_hash)
 
         yield self.storage.save_content_claim(
-            self.lbry_file.stream_hash, "%s:%i" % (claim_out['txid'], claim_out['nout'])
+            self.lbry_file.stream_hash, tx.outputs[0].id
```
jackrobison: It's confusing that this is the outpoint string - I thought this was the claim id.

lbrynet/core/Wallet.py
```diff
@@ -1,3 +1,4 @@
+# pylint: skip-file
```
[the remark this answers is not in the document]
eukreign: `Wallet.py` isn't used in the new wallet implementation; I just left it for now as a reference but it fails pylint hence I turned off pylint checking. We can delete `Wallet.py` once all of the commands are implemented and working.

lbrynet/daemon/auth/client.py
```diff
@@ -1,10 +1,11 @@
+# pylint: skip-file
```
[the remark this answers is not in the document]
eukreign: The comment can be removed, I added it when the CLI wasn't working but @hackrush01 has ported it to Python 3; he should have removed the comment when the port was complete and working. I'll do it now.

lbrynet/dht/error.py
eukreign:
```
$ python3.7
Python 3.7.0 (default, Jun 28 2018, 02:32:19) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import exceptions
ModuleNotFoundError: No module named 'exceptions'
```
¯\_(ツ)_/¯

lbrynet/dht/node.py
shyba: It has to return an `int`. That just helps converting it. `struct.unpack` doesn't support 384 bits afaik.

lbrynet/dht/protocol.py
shyba: https://docs.python.org/3/library/socket.html#socket.error
```
exception socket.error
A deprecated alias of OSError.

Changed in version 3.3: Following PEP 3151, this class was made an alias of OSError.
```

lbrynet/core/Wallet.py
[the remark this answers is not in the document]
jackrobison: please delete Wallet.py if it's not used any more

lbrynet/core/client/ClientProtocol.py
```diff
@@ -71,17 +69,17 @@ def timeoutConnection(self):
         self.peer.report_down()
         self.transport.abortConnection()
 
-    def connectionLost(self, reason):
+    def connectionLost(self, reason=None):
         log.debug("Connection lost to %s: %s", self.peer, reason)
         self.setTimeout(None)
         self.connection_closed = True
-        if reason.check(error.ConnectionDone):
+        if reason.check(error.ConnectionDone) or reason is None:
```
jackrobison: If reason is `None` then `reason.check` will be an attribute error, should be `if reason is None or reason.check(error.ConnectionDone):`

lbrynet/daemon/DaemonControl.py
```diff
@@ -57,29 +49,29 @@ def start():
...
     log.debug('Final Settings: %s', conf.settings.get_current_settings_dict())
 
-    try:
-        log.debug('Checking for an existing lbrynet daemon instance')
-        JSONRPCProxy.from_url(conf.settings.get_api_connection_string()).status()
-        log.info("lbrynet-daemon is already running")
-        return
-    except Exception:
-        log.debug('No lbrynet instance found, continuing to start')
+    # fixme: fix that, JSONRPCProxy is gone on py3
```
jackrobison: please delete or update

lbrynet/daemon/Publisher.py
```diff
@@ -31,7 +29,7 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
             raise Exception("Cannot publish empty file {}".format(file_path))
 
         file_name = os.path.basename(file_path)
-        with file_utils.get_read_handle(file_path) as read_handle:
+        with open(file_path, 'rb') as read_handle:
```
jackrobison: Previously this was only rb on windows, don't know if it makes a difference

lbrynet/daemon/Publisher.py
```diff
@@ -43,11 +41,13 @@ def create_and_publish_stream(self, name, bid, claim_dict, file_path, claim_addr
...
         claim_dict['stream']['source']['version'] = "_0_0_1"  # need current version here
-        claim_out = yield self.make_claim(name, bid, claim_dict, claim_address, change_address)
+        tx = yield self.wallet.claim_name(
+            name, bid, claim_dict, self.certificate, holding_address
+        )
 
         # check if we have a file already for this claim (if this is a publish update with a new stream)
         old_stream_hashes = yield self.storage.get_old_stream_hashes_for_claim_id(
-            claim_out['claim_id'], self.lbry_file.stream_hash
+            tx.outputs[0].claim_id, self.lbry_file.stream_hash
```
jackrobison: always outpoint 0?

lbrynet/daemon/auth/client.py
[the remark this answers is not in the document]
jackrobison: please remove before merge

lbrynet/dht/contact.py
```diff
@@ -110,10 +113,13 @@ def __ne__(self, other):
         else:
             return True
 
+    def __hash__(self):
```
jackrobison: why this?

jackrobison: also, why sometimes the id and sometimes a sum made from the address? doesn't seem like a good key if it's being used in a dictionary.

lbrynet/dht/encoding.py
```diff
@@ -1,134 +1,75 @@
...
-            raise DecodeError('Cannot decode empty string')
+from .error import DecodeError
+
+
+def bencode(data):
+    """ Encoder implementation of the Bencode algorithm (Bittorrent). """
+    if isinstance(data, int):
+        return b'i%de' % data
+    elif isinstance(data, (bytes, bytearray)):
+        return b'%d:%s' % (len(data), data)
```
jackrobison: Is this needed - isn't everything converted to a string first?

lbrynet/dht/kbucket.py
```diff
@@ -135,8 +137,8 @@ def keyInRange(self, key):
                  if not.
         @rtype: bool
         """
-        if isinstance(key, str):
-            key = long(key.encode('hex'), 16)
+        if isinstance(key, bytes):
+            key = int(hexlify(key), 16)
```
jackrobison: could use `struct.unpack` instead

lbrynet/daemon/Daemon.py
```diff
@@ -1,18 +1,34 @@
-# coding=utf-8
-import binascii
 import logging.handlers
 import mimetypes
 import os
 import requests
 import urllib
 import json
 import textwrap
+
```
jackrobison: please delete

lbrynet/daemon/Daemon.py
```diff
@@ -348,49 +371,75 @@ def _download_failed(error, download_id, name, claim_dict):
...
                                log.exception)
         self.analytics_manager.send_claim_action('publish')
-        log.info("Success! Published to lbry://%s txid: %s nout: %d", name, claim_out['txid'],
-                 claim_out['nout'])
-        defer.returnValue(claim_out)
+        nout = 0
+        log.info("Success! Published to lbry://%s txid: %s nout: %d", name, tx.id, nout)
+        defer.returnValue(self._txo_to_response(tx, nout))
+
+    def _txo_to_response(self, tx, nout):
```
jackrobison: can this be moved to the ledger class?

lbrynet/daemon/Daemon.py
```diff
@@ -348,49 +371,75 @@ def _download_failed(error, download_id, name, claim_dict):
...
-            defer.returnValue(result)
+        page = kwargs.get('page', 0)
+        page_size = kwargs.get('page_size', 10)
+        check_cache = kwargs.get('check_cache', False) # TODO: put caching back (was force_refresh parameter)
+        results = yield self.wallet.resolve(*uris, page=page, page_size=page_size)
+        self.save_claims((value for value in results.values() if 'error' not in value))
+        yield defer.returnValue(results)
+
+    @defer.inlineCallbacks
+    def save_claims(self, claim_infos):
```
jackrobison: should go in storage.py

lbrynet/daemon/Daemon.py
```diff
@@ -962,11 +1011,11 @@ def jsonrpc_wallet_balance(self, address=None, include_unconfirmed=False):
         Returns:
             (float) amount of lbry credits in wallet
         """
-        if address is None:
-            return self._render_response(float(self.wallet.get_balance()))
-        else:
-            return self._render_response(float(
-                self.wallet.get_address_balance(address, include_unconfirmed)))
+        assert address is None, "Limiting by address needs to be re-implemented in new wallet."
```
jackrobison:
```
if not address:
    raise NotImplementedError("Limiting by address needs to be re-implemented in new wallet.")
```

lbrynet/daemon/Daemon.py
```diff
@@ -1037,12 +1086,12 @@ def jsonrpc_wallet_encrypt(self, new_password):
         defer.returnValue(response)
 
     @defer.inlineCallbacks
-    def jsonrpc_daemon_stop(self):
+    def jsonrpc_stop(self):
```
jackrobison: undocumented breaking change

lbrynet/daemon/Daemon.py
```diff
@@ -2352,36 +2383,6 @@ def _disp(address):
         d.addCallback(lambda address: self._render_response(address))
         return d
 
-    @requires(WALLET_COMPONENT, conditions=[WALLET_IS_UNLOCKED])
```
jackrobison: please be sure this being removed ends up in the changelog

lbrynet/daemon/Daemon.py
```diff
@@ -2958,27 +2942,6 @@ def jsonrpc_blob_availability(self, blob_hash, search_timeout=None, blob_timeout
 
         return self._blob_availability(blob_hash, search_timeout, blob_timeout)
 
-    @requires(UPNP_COMPONENT, WALLET_COMPONENT, DHT_COMPONENT, conditions=[WALLET_IS_UNLOCKED])
```
jackrobison: another silently removed api call

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3045,264 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+                {'amount': txo.amount, 'address': txo.get_address(ledger)}
+                for txo in tx.outputs
+            ],
+            'total_input': tx.input_sum,
+            'total_output': tx.input_sum,
+            'total_fee': tx.fee,
+            'xhex': hexlify(tx.raw).decode(),
+        }
+
+    def get_account_or_error(self, argument: str, account_name: str, lbc_only=False):
```
jackrobison: these functions shouldn't be in the daemon class

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3045,264 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+            wallet.accounts.insert(0, account)
+            wallet.save()
+
+        result = account.to_dict()
+        result.pop('certificates', None)
+        result['is_default'] = wallet.accounts[0] == account
+        return result
+
+    @requires("wallet")
+    def jsonrpc_balance(self, account_name=None, confirmations=6, include_reserved=False,
```
jackrobison: `wallet_balance`

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3045,264 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+        Options:
+            --account=<account_name>        : (str) account for which to get max gaps
+
+        Returns:
+            (map) maximum gap for change and receiving addresses
+        """
+        return self.get_account_or_error('account', account_name).get_max_gap()
+
+    @requires("wallet")
+    def jsonrpc_fund(self, to_account, from_account, amount=0,
```
jackrobison: `wallet_fund`

lbrynet/daemon/DaemonCLI.py
```diff
@@ -1,3 +1,4 @@
+# pylint: skip-file
```
jackrobison: please delete

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3045,264 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+                        args['include_claims'] = True
+                    return account.get_balance(**args)
+            raise Exception("Couldn't find an account named: '{}'.".format(account_name))
+        else:
+            if include_claims:
+                raise Exception("'--include-claims' requires specifying an LBC account.")
+            return self.wallet.get_balances(confirmations)
+
+    @requires("wallet")
+    def jsonrpc_max_address_gap(self, account_name):
```
jackrobison: `wallet_max_address_gap`

lbrynet/daemon/Publisher.py
eukreign: we create the TX a little bit before this line ourselves, so we know upfront that claim txo is 0.

lbrynet/core/Wallet.py
```diff
@@ -1,1325 +0,0 @@
...
-            wallet.synchronize()
-        self.wallet = wallet
-        self.is_wallet_unlocked = not self.wallet.use_encryption
-        self._check_large_wallet()
-        return defer.succeed(True)
-
-    def _check_large_wallet(self):
-        addr_count = len(self.wallet.addresses(include_change=False))
-        if addr_count > 1000:
-            log.warning("Your wallet is excessively large (%i addresses), "
```
lbrynaut: Other than the wallet being slow at this point (via implementation), what's the harm?  If the wallet is used, it's inevitably going to reach this point.

lbrynet/core/Wallet.py
```diff
@@ -1,1325 +0,0 @@
...
-        return defer.succeed(self.wallet.create_new_address(account=account,
-                                                            for_change=for_change))
-
-    # Get the balance of a given address.
-    def get_address_balance(self, address, include_balance=False):
-        c, u, x = self.wallet.get_addr_balance(address)
-        if include_balance is False:
-            return Decimal(float(c) / COIN)
-        else:
-            return Decimal((float(c) + float(u) + float(x)) / COIN)
```
lbrynaut: ^ These kinds of conversions will be inaccurate at some point.  These are a show-stopper to me until corrected.

lbrynet/cli.py
```diff
@@ -0,0 +1,162 @@
...
+
+def print_help_for_command(command):
+    fn = Daemon.callable_methods.get(command)
+    if fn:
+        print(dedent(fn.__doc__))
+    else:
+        print("Invalid command name")
+
+
+def normalize_value(x, key=None):
```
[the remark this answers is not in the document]
eukreign: This is only used for parsing command line arguments, would that be affected by lbrycrd changes?

lbrynet/core/StreamDescriptor.py
```diff
@@ -344,18 +354,18 @@ def get_blob_hashsum(b):
...
-    blob_hashsum.update(iv)
-    blob_hashsum.update(str(length))
+        blob_hashsum.update(blob_hash.encode())
+    blob_hashsum.update(str(blob_num).encode())
+    blob_hashsum.update(iv.encode())
+    blob_hashsum.update(str(length).encode())
     return blob_hashsum.digest()
 
 
 def get_stream_hash(hex_stream_name, key, hex_suggested_file_name, blob_infos):
```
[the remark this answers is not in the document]
eukreign: This is just for making a hash, it only needs to be deterministic. Do you expect that normalization in lbrycrd would break the hash?

lbrynet/core/Wallet.py
```diff
@@ -1,1325 +0,0 @@
...
-        payment, before the service is rendered
-
-        @param identifier: The peer to which the payment will ultimately be sent
-
-        @param amount: The amount of points to reserve
-
-        @return: A ReservedPoints object which is given to send_points
-            once the service has been rendered
-        """
-        rounded_amount = Decimal(str(round(amount, 8)))
```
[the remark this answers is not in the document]
eukreign: You are commenting on a file that was deleted.

lbrynet/core/Wallet.py
```diff
@@ -1,1325 +0,0 @@
...
-        d = self.get_least_used_address()
-        d.addCallback(set_address_for_peer)
-        return d
-
-    def _send_payments(self):
-        payments_to_send = {}
-        for address, points in self.queued_payments.items():
-            if points > 0:
-                log.debug("Should be sending %s points to %s", str(points), str(address))
-                payments_to_send[address] = points
```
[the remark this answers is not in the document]
eukreign: You are commenting on a file that was deleted.

lbrynet/core/Wallet.py
eukreign: You are commenting on a file that was deleted.

lbrynet/core/Wallet.py
eukreign: You are commenting on a file that was deleted.

lbrynet/daemon/Daemon.py
eukreign: Are there other kinds of balances that are common? I think `balance` is pretty obvious that it's balance of your wallet.

lbrynet/daemon/Daemon.py
eukreign: In what context would `fund` mean something other than wallet related funding operations?

lbrynet/daemon/Daemon.py
eukreign: Is there a significant benefit to adding `wallet` to this commands name? Seems like it just forces users to type more on the command line...

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3073,245 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
-            --pos_arg=<pos_arg>                : (int) pos arg
-            --pos_args=<pos_args>              : (int) pos args
-            --pos_arg2=<pos_arg2>              : (int) pos arg 2
-            --pos_arg3=<pos_arg3>              : (int) pos arg 3
+            --account_name=<account_name>   : (str) name of the account to create or update
+            --create                        : (bool) create the account
+            --delete                        : (bool) delete the account
+            --single_key                    : (bool) create single key account, default is multi-key
+            --seed=<seed>                   : (str) seed to generate new account from
+            --private_key=<private_key>     : (str) private key for new account
```
lbrynaut: Command line data leak.

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3073,245 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+        if create:
+            self.error_if_account_exists(account_name)
+            if single_key:
+                address_generator = {'name': SingleKey.name}
+            else:
+                address_generator = {
+                    'name': HierarchicalDeterministic.name,
+                    'receiving': {
+                        'gap': receiving_gap or 20,
+                        'maximum_uses_per_address': receiving_max_uses or 2},
```
lbrynaut: Why 2?

lbrynet/daemon/Daemon.py
lbrynaut: An `Index fund` :-P

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3073,245 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+                for chain_name in address_changes:
+                    chain = getattr(account, chain_name)
+                    for attr, value in address_changes[chain_name].items():
+                        if value is not None:
+                            setattr(chain, attr, value)
+                            change_made = True
+            if change_made:
+                wallet.save()
+
+        if default:
```
lbrynaut: Might not be worth noting, but wallets recovered from seeds lose this kind of info.

lbrynet/daemon/Daemon.py
```diff
@@ -3082,29 +3073,245 @@ def jsonrpc_stream_availability(self, uri, search_timeout=None, blob_timeout=Non
...
+            --everything                  : (bool) transfer everything (excluding claims), default: false.
+            --outputs=<outputs>           : (int) split payment across many outputs, default: 1.
+            --broadcast                   : (bool) actually broadcast the transaction, default: false.
+
+        Returns:
+            (map) maximum gap for change and receiving addresses
+
+        """
+        to_account = self.get_account_or_error('to_account', to_account)
+        from_account = self.get_account_or_error('from_account', from_account)
```
lbrynaut: Hrm, where are you storing the mapping from named accounts to their index?  Just noting that those will also be lost in the case of seed recovery.

lbrynet/daemon/auth/util.py
```diff
@@ -9,21 +9,22 @@
...
 
 
-def sha(x):
+def sha(x: bytes) -> bytes:
     h = hashlib.sha256(x).digest()
     return base58.b58encode(h)
 
 
-def generate_key(x=None):
+def generate_key(x: bytes = None) -> bytes:
```
lbrynaut: Might be worth indicating this returns a base58 encoded value.  And same for sha above, as it's not generally assumed that sha hashes and keys are base58 encoded.

lbrynet/dht/distance.py
```diff
@@ -1,21 +1,23 @@
...
-class Distance(object):
+class Distance:
     """Calculate the XOR result between two string variables.
 
     Frequently we re-use one of the points so as an optimization
-    we pre-calculate the long value of that point.
+    we pre-calculate the value of that point.
     """
 
     def __init__(self, key):
```
lbrynaut: What format are these incoming keys?  It looks odd to encode it as base16, then decode it from base16 to get an int.

lbrynet/core/utils.py
```diff
@@ -88,15 +89,19 @@ def version_is_greater_than(a, b):
     return pkg_resources.parse_version(a) > pkg_resources.parse_version(b)
 
 
+def rot13(some_str):
+    return codecs.encode(some_str, 'rot_13')
+
+
 def deobfuscate(obfustacated):
```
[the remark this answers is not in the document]
eukreign: I think here we're just porting it to py3. Considering if it even makes sense to keep this function would be for a future PR.

lbrynet/core/utils.py
```diff
@@ -62,9 +63,9 @@ def safe_stop_looping_call(looping_call):
 def generate_id(num=None):
     h = get_lbry_hash_obj()
     if num is not None:
-        h.update(str(num))
+        h.update(str(num).encode())
     else:
-        h.update(str(random.getrandbits(512)))
+        h.update(str(random.getrandbits(512)).encode())
```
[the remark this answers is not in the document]
eukreign: It's used to track installations. Important to LBRY but probably not important to users. I think here we're just porting it to py3; without making any assertions about it's correctness or applicability outside of just making it work the same in py3 as it did in py2.

lbrynet/daemon/Daemon.py
```diff
@@ -1540,29 +1554,24 @@ def jsonrpc_channel_new(self, channel_name, amount):
...
-            elif amount > max_bid_amount:
-                raise InsufficientFundsError(
-                    "Please wait for any pending bids to resolve or lower the bid value. "
-                    "Currently the maximum amount you can specify for this channel is {}"
-                    .format(max_bid_amount)
-                )
+        amount = self.get_dewies_or_error("amount", amount)
 
-        result = yield self.wallet.claim_new_channel(channel_name, amount)
+        if amount <= 0.0:
```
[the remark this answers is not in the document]
eukreign: Good catch! Amount was converted to ints by `get_dewies_or_error()` so the comparison here should be using `0`.

lbrynet/daemon/Daemon.py
eukreign: Changed it to 1.

lbrynet/daemon/Daemon.py
eukreign: You're saying the seed isn't stored after restoring the account from a seed? Or did you mean something else, re: "this kind of info"?

lbrynet/daemon/Daemon.py
eukreign: You may be thinking of accounts under a single HD chain. Here each "Account" is a root of an HD chain. Basically, you can have more than one HD root in the same wallet.

lbrynet/dht/distance.py
eukreign: @jackrobison or @shyba can you guys answer this one?

lbrynet/daemon/auth/util.py
eukreign: I think we're going to do a types audit later and improved the typing. For now this is better than it was before :-D (before you didn't even know if it returned bytes or unicode string :scream:)

lbrynet/daemon/Daemon.py
eukreign: After reading the next comment I know I know what you meant by "this kind of info". You're talking about the sub-account in an HD chain. It's not an issue here because each Account is actually an independent root. When restoring from seed you're restoring a single Account and adding it to the list of other accounts also restored from a seed.

The reason for having a `default` account is that the current desktop app does not know that  wallet can have more than one account and also because it's convenient to have a default account which is used for payments while having other accounts as read-only or for other reasons.

lbrynet/dht/distance.py
shyba: nice catch! they come in big endian unsigned bytes. This function makes them into hex, then converts hex to int. It was like that in py2, but there is a better readable way to do in py3, see:
https://github.com/lbryio/lbry/pull/1370/files#diff-3a31e4218bf33300ff7b84e484ef1c2bR140
I was solving those cases, but totally forgot that one.

lbrynet/daemon/Daemon.py
lbrynaut: Maybe we're talking about the same thing, but given that HD accounts are integer indexed and your accounts are strings, I assume that you wanted _named_ accounts, such as 'checking', or 'saving' which maps to index 0 and 1 respectively (as an example).  Further, setting the 'saving' at index 1 as the default is well and good, but when recovering the wallet from seed (i.e. on a new machine) the names and default account mappings would be lost, no?

lbrynet/daemon/Daemon.py
lbrynaut: Right.  BTW, which BIP32 derivation are you following?

lbrynet/daemon/Daemon.py
eukreign: Just to reiterate the `Account` in the code above is *not* the same as HD accounts. An `Account` represents a distinct HD root (completely unrelated to other `Account` roots). When you restore from a seed you are restoring a single `Account` only, unrelated to other accounts.

lbrynet/daemon/Daemon.py
eukreign: The new wallet is backwards compatible with the old one in terms of how the `receiving` and `change` address chains are defined. Which is what we inherited from electrum. I don't believe it follows the BIP32 spec exactly.

The code for this is here:
https://github.com/lbryio/torba/blob/master/torba/bip32.py
https://github.com/lbryio/torba/blob/master/torba/baseaccount.py

kaykurokawa:
Any reason why the LBRY implementation was separated into the lbry repo ? (i.e. why is there no coin/lbrycrd.py)

lbrynet/daemon/Daemon.py
lbrynaut: Got it. thanks.

lbrynet/daemon/Daemon.py
lbrynaut: Electrum is definitely BIP32 compatible.

eukreign:
The reason lbry stuff is not in `torba` is because it makes issue tracking and managing releases easier and faster. Eventually the goal is that `torba` will have much fewer changes and will stay constant despite more frequent `lbrynet` releases.

Previously with lbryum there was additional administrative and tracking overhead in having lbry daemon issues spread across multiple repos. You have to always think which repo a new issue should go and in cases where an issue touched both lbrynet and the wallet there is no good answer. Now everything lbrynet daemon specific is in the lbry repo.

lbrynet/dht/distance.py
eukreign: @lbrynaut, @shyba made the changes and they have been merged into this PR.

PR #1161, opened 2018-03-31
fix lbrynet-cli when using authentication

jackrobison:
-add explanation of daemon authentication to AuthJSONRPCServer docstring
-remove auth_required decorator, use auth for all api methods if use_authentication is true

### DIFF
--- a/lbrynet/daemon/auth/client.py
+++ b/lbrynet/daemon/auth/client.py
@@ -1,19 +1,24 @@
-import urlparse
-import logging
-import requests
 import os
-import base64
 import json
-
-from lbrynet.daemon.auth.util import load_api_keys, APIKey, API_KEY_NAME, get_auth_message
-from lbrynet import conf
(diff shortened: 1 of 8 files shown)


lbrynet/daemon/auth/client.py
```diff
@@ -130,7 +121,7 @@ def config(cls, key_name=None, key=None, pw_path=None,
             except AttributeError:
                 pass
             authpair = user + b':' + passwd
-            auth_header = b'Basic ' + base64.b64encode(authpair)
+            auth_header = b'Basic ' + base58.b58encode(authpair)
```
hackrush01: In the issue comment you said that you'll be using b64 instead of b58, but is seems like opposite is done. Please crosscheck.

jackrobison: I rebased this out (since it makes urls containing bad characters), will update the comment

PR #1162, opened 2018-04-01
Register API calls during daemon startup

hackrush01:


lbrynet/daemon/Daemon.py
```diff
@@ -168,7 +168,8 @@ class Daemon(AuthJSONRPCServer):
     """
 
     allowed_during_startup = [
-        'daemon_stop', 'status', 'version', 'wallet_unlock'
+        'daemon_stop', 'status', 'version', 'wallet_unlock', 'commands', 'help', 'settings_get',
```
jackrobison: Let's move this out to a decorator

lbrynet/daemon/auth/server.py
```diff
@@ -187,6 +194,13 @@ def __init__(self, use_authentication=None):
     def setup(self):
         return NotImplementedError()
 
+    def register_commands(self, command_type):
+        for methodname in dir(self):
+            if methodname.startswith("jsonrpc_"):
+                method = getattr(self, methodname)
+                if hasattr(method, '_requires_' + command_type):
+                    self.allowed_during_startup.append(methodname.split("jsonrpc_")[1])
```
jackrobison: why is this appending to allowed_during_startup?

lbrynet/daemon/Daemon.py
jackrobison: Actually, let's have commands with no required components be allowed during startup. Then we can get rid of this attribute.

PR #1166, opened 2018-04-18
fix cli kwargs again??

hackrush01:


lbrynet/daemon/DaemonCLI.py
```diff
@@ -125,7 +125,7 @@ def main():
     # TODO: check if port is bound. Error if its not
 
     try:
-        result = api.call(method, kwargs)
+        result = api.call(method, **kwargs)
```
jackrobison: `call` doesn't take variable keyword args: https://github.com/lbryio/lbry/blob/4e7bdafdcfb80419234a5db0dcce12134f015a7b/lbrynet/daemon/auth/client.py#L48

jackrobison:
Can you post the error? These commands run fine for me with and without kwargs currently, and with this change would raise `TypeError: call() got an unexpected keyword argument` with the keyword arg

hackrush01:
@jackrobison 
```
lbrynet-cli status
ERROR: sequence item 0: expected string, dict found
```

```
2018-04-18 19:54:09,408 ERROR    lbrynet.daemon.auth.server:227: sequence item 0: expected string, dict found
2018-04-18 19:54:09,408 WARNING  lbrynet.daemon.auth.server:211: error processing api request: sequence item 0: expected string, dict found
traceback:
```

jackrobison:
fixed by https://github.com/lbryio/lbry/commit/450bed6350f3d4aab3a49432a59f45f62bad69ec

PR #1169, opened 2018-04-19
return error correctly when resolving invalid/expired url. fixes #1024

### LINKED ISSUE #1024: Batch resolve should return null or an appropriate error for unclaimed, abandoned, or expired URLs.
## The Issue
In a batched resolution, the entire resolve command fails if one of the URLs errors out. This is happening on the homepage at the moment because of an expired claim in the resolve call.  

### Steps to reproduce
1. batch a URI resolution with a known bad URL
2. observe failure

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

## System Configuration

 

- LBRY Daemon version: 0.18.2
- LBRY App version: 0.19rc1
- LBRY Installation ID:
- Operating system:

lyoshenka:
I don't know if this is the right solution. The whole thing is a mess of raising and reraising exceptions, and sometimes not raising but just setting an `error` value on a dict. This seems to work. If there's a better way to do this, please teach me.

The error was actually happening when `Wallet.save_claim()` was being called on a claim with a value of `None`. I see several other places where that's in danger of happening (does it actually happen there? idk). 

I think the REAL solution is for lbryschema to raise something more descriptive than `DecodeError` (is it really a decoding error when you successfully decode an expired url?) and for the daemon to catch that error and handle it correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ at anytime.
   * `use_auth_http` in a config file being overridden by the default command line argument to `lbrynet-daemon`, now the command line value will only override the config file value if it is provided
   * `lbrynet-cli` not automatically switching to the authenticated client if the server is detected to be using authentication. This resulted in `lbrynet-cli` failing to run when `lbrynet-daemon` was run with the `--http-auth` flag
   * fixed error when using `claim_show` with `txid` and `nout` arguments
+  * fixed resolving expired URLs
 
 ### Deprecated
   *
--- a/lbrynet/core/Wallet.py
+++ b/lbrynet/core/Wallet.py
@@ -366,9 +366,10 @@ class Wallet(object):
             claim['value'] = claim_dict
             claim['hex'] = decoded.serialized.encode('hex')
         except DecodeError:
-            claim['hex'] = claim['value']
-            claim['value'] = None
-            claim['error'] = "Failed to decode value"
+            raise UnknownURI(claim['permanent_url'] if 'permanent_url' in claim else "")
+            # claim['hex'] = claim['value']
+            # claim['value'] = None
+            # claim['error'] = "Failed to decode value"
         return claim
 
     def _handle_claim_result(self, results):


jackrobison:
What's the value being decoded? It's probably not a claim if `DecodeError` is being raised, more likely an empty string or `None`.

Additionally, a claim that can't be decoded (for example a claim with an empty value) is totally unrelated to it being expired.

jackrobison:
The error is fixed in https://github.com/lbryio/lbry/commit/1f1b34863805f4954fbef3f163ef65268a66771a

`xzxzx#34` correctly resolves to 

```{
  "xzxzx#34": {
    "error": "URI lbry://xzxzx#34 cannot be resolved"
  }
}
```
With the given commit `mindblown#02da15270141483a76a7b8372d69653b735fb0b9` resolves to

```
{
  "mindblown#02da15270141483a76a7b8372d69653b735fb0b9": {
    "claim": {
      "address": "bSEdnYftD9suojvYkGZseTVMEutjCCg2aD",
      "amount": 1.0,
      "claim_id": "02da15270141483a76a7b8372d69653b735fb0b9",
      "claim_sequence": 1,
      "decoded_claim": false,
      "depth": 353907,
      "error": "Failed to decode value",
      "height": 2806,
      "hex": "7b226465736372697074696f6e223a2022696d706f737369626c65222c20226c6963656e7365223a2022222c2022617574686f72223a2022222c20226c616e6775616765223a20224e6f6e65222c20227469746c65223a2022776f6168222c2022736f7572636573223a207b226c6272795f73645f68617368223a2022663261313064636365646134303830333962633264623135636434343332323130323665656431313339333537383231316636326366663264386264626336303232636335326266353732613837383030343833383037666132336266653135227d2c2022636f6e74656e742d74797065223a2022696d6167652f676966227d",
      "name": "mindblown",
      "nout": 1,
      "permanent_url": "mindblown#02da15270141483a76a7b8372d69653b735fb0b9",
      "txid": "3d4e97fce6b0cd4fe48b1741d04924b3b1edf6ccaa5e3cbaaf4f2e090f1eab23",
      "value": null
    }
  }
}
```

This is still not correct, as mindblown#02da15270141483a76a7b8372d69653b735fb0b9 is expired. This must be fixed in lbryum server to not return them (lbrycrdd returns `{}` when looking up the claim by claim id), and should also be checked in lbryum client (by claim depth)

PR #1174, opened 2018-04-28
Implement sorting in file_list

### LINKED ISSUE #1143: Add --sort option to file_list
## The Issue
I've noticed previously that my Downloads items don't show in the correct order - not sure exactly what causes them to get out of sync in file_list. The question is - should file_list have a field which has the download date (or a better name?) for a particular file? There doesn't seem to be any information currently available which would help with sorting a user's downloads. 

## Jeremy's Revised Issue
- Add a `--sort` flag to `file_list`
- Sort should take the following options: name, date, price
- Sort can be ASC or DESC

kafene:
This resolves #1143 by implementing sort in file_list based on name/date/price, in ascending or descending order.

I do have a standing [question in that issue](https://github.com/lbryio/lbry/issues/1143#issuecomment-384476235) regarding the addition of something like a `created_at` field in the `files` table. For now, the code in this PR depends on the insertion order of records into the database, which is related to the date that the record was created, but does not keep a record of it.


lbrynet/daemon/Daemon.py
```diff
@@ -97,6 +97,16 @@
...
+FILE_SORT_FIELD_NAME = 'name'
+FILE_SORT_FIELD_DATE = 'date'
+FILE_SORT_FIELD_PRICE = 'price'
+
+FILE_SORT_ORDER_ASCENDING = 'asc'
+FILE_SORT_ORDER_DESCENDING = 'desc'
+FILE_SORT_ORDERS = [
+    FILE_SORT_ORDER_ASCENDING,
+    FILE_SORT_ORDER_DESCENDING,
+]
```
eukreign: Could be a `tuple` instead of a `list`, conveys that this won't be modified elsewhere and takes up less memory.

lbrynet/daemon/Daemon.py
```diff
@@ -925,11 +935,13 @@ def _get_lbry_file(self, search_by, val, return_json=False, full_status=False):
         defer.returnValue(lbry_file)
 
     @defer.inlineCallbacks
-    def _get_lbry_files(self, return_json=False, full_status=True, **kwargs):
+    def _get_lbry_files(self, return_json=False, full_status=True, sorts=None, **kwargs):
```
eukreign: Would you consider `sort_by` as variable name instead of `sorts`?

lbrynet/daemon/Daemon.py
```diff
@@ -939,6 +951,33 @@ def _get_lbry_files(self, return_json=False, full_status=True, **kwargs):
         log.debug("Collected %i lbry files", len(lbry_files))
         defer.returnValue(lbry_files)
 
+    def _sort_lbry_files(self, lbry_files, sorts):
+        for field, order in sorts:
+            reverse = order == FILE_SORT_ORDER_DESCENDING
```
eukreign: Would you consider `direction` instead of `order` and `is_reversed` instead of `reverse`?

lbrynet/daemon/Daemon.py
```diff
@@ -939,6 +951,33 @@ def _get_lbry_files(self, return_json=False, full_status=True, **kwargs):
...
+            if field == FILE_SORT_FIELD_NAME:
+                lbry_files = sorted(lbry_files, key=lambda f: f.file_name, reverse=reverse)
+            elif field == FILE_SORT_FIELD_DATE:
+                lbry_files = sorted(lbry_files, reverse=reverse)
+            elif field == FILE_SORT_FIELD_PRICE:
+                lbry_files = sorted(lbry_files, key=lambda f: f.points_paid, reverse=reverse)
+            else:
+                raise Exception('Unrecognized sort field "{}"'.format(field))
+        return lbry_files
```
eukreign: Instead of having the variable `lbry_files` I think it's fine to just have `return` in front of each `sorted()` call. Less code, equally easy to read, less variables, etc.

kafene: The main reason I did this was to support multiple passes, so for example

    lbrynet-cli file_list --sort=name,desc --sort=price,asc

would produce a list sorted by price, but items with the same price would be sorted by name.

kafene:
Thanks for the review, I will make the suggested changes and work on unit tests for the parsing of sort options and sorting functionality with some mock data.

kafene:
I added a pretty thorough unit test for file_list sorting, using the [Faker](https://github.com/joke2k/faker) library for generating some somewhat-realistic looking fake file objects, and added both Faker and mock (which I was missing) to `requirements_testing.txt`.

The following should quickly reproduce the test results:

    source lbry-venv/bin/activate
    pip install -U -r requirements_testing.txt
    trial lbrynet/tests/unit/lbrynet_daemon/test_Daemon.py

eukreign:
@kafene, I apologize for the long period without feedback. I've actually reviewed this more closely since your last commits and we've even discussed it internally.

The main issue blocking this pull request is that there is too much hard coded logic in this implementation. For example, imagine needing to add a new field in the future... this would require changes in several different places and adding new if statements to handle the new field.

Could you refactor this implementation so that the field names are inferred from the data model (as best as you can, with possible dictionary mapping between user friendly field names and the actual object attributes) with the goal being that adding a new field should be a simple and obvious addition to the mapping instead of making changes in various places (adding new constants, modifying if statements, etc).

kafene:
I agree. It should simply be any field that is visible when doing a normal `file_list` command.

One issue with the generic approach is that I had the sort applied outside of the transformation to JSON that is ultimately displayed to the end user. It looks like the only place that sorting this way is needed is when JSON is being returned, so it would make sense to only support sorting when `return_json=True` and to sort the list of objects returned by `_get_lbry_file_dict`, so the sort would actually be applied to the data seen in the output.

This is definitely doable. I would also like to support some notation like `--sort=metadata.title`.

kafene:
Okay, next round.

Sorting can now be done on any property of an object found in the results of `file_list`. This includes nested fields like `metadata.author` or `metadata.title`.

I added a test case for sorting by nested fields, currently only applicable to `metadata`.

I noticed that sorting by a field that does not exist produced a very vague error message with just the name of the key that was not found, so I added a more meaningful error message.

I also fixed up the unit tests to not use `@inlineCallbacks` and instead use other means of testing deferred objects as [suggested in the Twisted docs](https://twistedmatrix.com/documents/current/core/howto/trial.html#testing-deferreds-without-the-reactor).

Things to consider:

Sorting by date is still not explicitly supported. The real fix for this is to add a date field to the output, which also means storing it somewhere, most likely in the `file` table of the database. That's out of scope for this PR, I think.

If in the future any lists/arrays are added to the results, sorting on numerically-indexed keys of these would not be supported. A simple solution would be to add something like `key = int(key) if isinstance(value, (list, tuple))` in the `key_getter` callback function, I just refrained from doing so for now because it's a case that currently doesn't occur (premature optimization?). Let me know if you think it should be accounted for regardless and I can add it.

docs/cli/index.html
```diff
@@ -496,6 +496,13 @@
     publish
   </a>
   
+</li>
```
jackrobison: The docs files get built automatically and shouldn't be manually changed (they are generated from the api function docstrings)

jackrobison:
@eukreign @tzarebczan @kafene 

https://github.com/lbryio/lbry/issues/1227

https://github.com/lbryio/lbry/issues/1228

PR #1175, opened 2018-04-29
Initialize settings in auth unit tests

orblivion:
I was hitting a snag when running unit tests:

https://github.com/lbryio/lbry/blob/master/lbrynet/daemon/auth/server.py#L175

`conf.settings` was still set to `None`


orblivion:
Robot tells me to update changelog. Changelog says to include "notable" changes. This not being so notable, I will defer to human instructions.

jackrobison:
@orblivion @eukreign 

I ran into this too with `trial lbrynet.tests.unit.lbrynet_daemon`

I'm fixing the broken tests on appveyor now, I'll include a fix for this too.

PR #1177, opened 2018-04-30
download headers from s3 when more than 10 chunks behind

jackrobison:


kauffj:
Yes!! If this is safe, can we get daemon patch release for this?

@seanyesmunt please monitor this and get into app release ASAP.

Also @robvsmith one for you to know about (fixes first run timing).

tzarebczan:
I thought we didn't want to do this as it won't be necessary when lbryumx is live, but I support it in the mean time.

Regarding the patch release, Igor is still working through a daemon upgrade issue that's blocking it.

tzarebczan:
Jack said he'd add some debugging to this to see why it would fail to download from S3.

PR #1194, opened 2018-05-08
track successful reflector uploads in sqlite

jackrobison:
-add `reflected_stream` table to the database to track reflector uploads so we don't check reflector server unnecessarily 
-increase the default `auto_re_reflect_interval` to a day

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -280,10 +280,10 @@ ADJUSTABLE_SETTINGS = {
     'peer_port': (int, 3333),
     'pointtrader_server': (str, 'http://127.0.0.1:2424'),
     'reflector_port': (int, 5566),
-    # if reflect_uploads is True, send files to reflector (after publishing as well as a
-    # periodic check in the event the initial upload failed or was disconnected part way through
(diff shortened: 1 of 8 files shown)


lbrynet/conf.py
```diff
@@ -283,7 +283,7 @@ def _convert_value(value):
     # if reflect_uploads is True, send files to reflector (after publishing as well as a
     # periodic check in the event the initial upload failed or was disconnected part way through
```
shyba: maybe update this comment? `reflect_uploads` doesn't control the re-reflect anymore

shyba: Sorry, it also does, but this became a bit confusing. Maybe the `# set to 0 to disable` added below is enough. Or explain that they both need to be active for enabling re-reflect.

shyba:
Except for that detail on the older comment, the code looks very good to me. The old comment is not even wrong, just confusing.

lbrynet/conf.py
jackrobison: If `reflect_uploads` is true, streams made with `publish` will be uploaded to reflector. If `reflect_uploads` is true and `auto_re_reflect_interval` > 0 files in the file manager will be reflected.

Issue #1098, opened 2018-02-02
claim_list_mine should return results ordered consistently (by height?)


tzarebczan:
## The Issue
Recently noticed an issue where newly published items were showing up out of order on my published list. Checked with the app team and they take the result of claim_list_mine and reverse it to display on the page. This means that claim_list_mine isn't returning the results in a consistent manner.  

My claim_list_mine output is below. It's returning heights in this order:  307153, 314458, 314495, 311418

```
{
  "id": 566,
  "jsonrpc": "2.0",
  "result": [
    {
      "address": "bEuHsiU6j56qf2MaYgKABvdFsncdsHZP98",
      "amount": 0.009625,
      "blocks_to_expiration": 255625,
      "category": "update",
      "claim_id": "d079c45f310727eedf104223846e7419e6b8c0f7",
      "confirmations": 7349,
      "decoded_claim": true,
      "expiration_height": 570127,
      "expired": false,
      "has_signature": false,
      "height": 307153,
      "is_pending": false,
      "is_spent": false,
      "name": "sideshowtest03atomic",
      "nout": 0,
      "txid": "5fd2db4a33176f61687589ee80ea1776b297cb32a8297524191595e916729822",
      "value": {
        "claimType": "streamType",
        "stream": {
          "metadata": {
            "author": "",
            "description": "",
            "language": "en",
            "license": "None",
            "licenseUrl": "",
            "nsfw": false,
            "preview": "",
            "thumbnail": "",
            "title": "test sideshow 03 moov atom command",
            "version": "_0_1_0"
          },
          "source": {
            "contentType": "video/mp4",
            "source": "ff8d5a23bb6e938be0d9498ed7b6e1e7cd628227f756182f6d10035648b5238c653050a02c677f63b311f761dd729013",
            "sourceType": "lbry_sd_hash",
            "version": "_0_0_1"
          },
          "version": "_0_0_1"
        },
        "version": "_0_0_1"
      }
    },
    {
      "address": "bED4V6pRTDM3SBBM9vngo3Gr5S9mN5crrS",
      "amount": 0.01,
      "blocks_to_expiration": 262930,
      "category": "claim",
      "claim_id": "2895e7138b9cf05f1469bbe5a4cb0073676429aa",
      "confirmations": 44,
      "decoded_claim": true,
      "expiration_height": 577432,
      "expired": false,
      "has_signature": false,
      "height": 314458,
      "is_pending": false,
      "is_spent": false,
      "name": "lbry0200",
[...]

PR #1208, opened 2018-05-16
WIP: feature/1098/sorted-claim-results

tau3:
Hi! 

I've implemented a fix and unit test for #1098, but i'm not sure what 'outpoint' property is.

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -36,6 +36,7 @@ from lbrynet.daemon.Downloader import GetStream
 from lbrynet.daemon.Publisher import Publisher
 from lbrynet.daemon.ExchangeRateManager import ExchangeRateManager
 from lbrynet.daemon.auth.server import AuthJSONRPCServer
+from lbrynet.daemon.claims_comparator import arrange_results
 from lbrynet.core.PaymentRateManager import OnlyFreePaymentsManager
 from lbrynet.core import utils, system_info
 from lbrynet.core.StreamDescriptor import StreamDescriptorIdentifier, download_sd_blob
@@ -2293,6 +2294,7 @@ class Daemon(AuthJSONRPCServer):
         """
 
         d = self.session.wallet.get_name_claims()
+        d.addCallback(arrange_results)
         d.addCallback(lambda claims: self._render_response(claims))
         return d
 
@@ -2331,7 +2333,8 @@ class Daemon(AuthJSONRPCServer):
         """
 
         claims = yield self.session.wallet.get_claims_for_name(name)
-        defer.returnValue(claims)
+        result = arrange_results(claims)
+        defer.returnValue(result)
 
     @defer.inlineCallbacks
     def jsonrpc_claim_list_by_channel(self, page=0, page_size=10, uri=None, uris=[]):
--- a/lbrynet/daemon/claims_comparator.py
+++ b/lbrynet/daemon/claims_comparator.py
@@ -1,0 +1,36 @@
+_comparison_order = ['height', 'name', 'claim_id']  # TODO outpoint
+
+
+def arrange_results(claims):
+    for claim in claims:
+        results = claim['result']
+        sorted_results = sorted(results, cmp=_compare_results)
+        claim['result'] = sorted_results
+    return claims
+
+
+def _compare_results(left, right):
+    """
+    :type left: dict
+    :type right: dict
+    """
+    result = 0
+
+    for attribute in _comparison_order:
+        left_value = left[attribute]
+        right_value = right[attribute]
+        sub_result = _cmp(left_value, right_value)
+        if sub_result is not 0:
+            result = sub_result
+            break
+
+    return result
+
+
+def _cmp(left, right):
+    if left == right:
+        return 0
+    elif left < right:
+        return -1
+    else:
+        return 1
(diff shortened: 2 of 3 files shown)


tzarebczan:
@tau3 , thanks for this contribution! Please check out https://lbry.io/faq/tips 

We'll get a proper review on this PR from one of the devs. 

To answer your question about outpoint, this is the txid+ nout, separated by a colon, i.e. 88f880fb40c296f1a1b64c0bba240c046892c5d2000dd7c5090e8e4d19911f7b:1 - So you'd be comparing those 2 fields. We use that connotation more on the [file_list](https://lbryio.github.io/lbry/cli/#file_list), sorry for the confusion.

tau3:
@eukreign ok, it's done 😄

lbrynet/daemon/claims_comparator.py
```diff
@@ -0,0 +1,10 @@
+def arrange_results(claims):
+    for claim in claims:
+        results = claim['result']
+        sorted_results = sorted(results, key=lambda d: (d['height'], d['name'], d['claim_id'], _outpoint(d)))
+        claim['result'] = sorted_results
+    return claims
+
+
+def _outpoint(claim):
+    return '{}:{}'.format(claim['txid'], claim['nout'])
```
eukreign: Can you simplify it more and move it into the `Daemon.py`, for example:
```python
def sort_claim_results(claims):
    for claim in claims:
        claim['result'].sort(
            key=lambda d: (d['height'], d['name'], d['claim_id'], d['txid'], d['nout'])
        )
    return claims
```

lbrynet/tests/unit/daemon/test_claims_comparator.py
```diff
@@ -0,0 +1,27 @@
...
+import unittest
+
+from lbrynet.daemon.claims_comparator import arrange_results
+
+
+class ClaimsComparatorTest(unittest.TestCase):
+    def setUp(self):
+        with open('claims_comparator_cases.json') as f:
+            document = json.load(f)
+        self.cases = document['cases']
```
eukreign: What is the rationale for storing the dictionaries in a separate json file instead of just having this directly in the code?

lbrynet/tests/unit/daemon/test_claims_comparator.py
```diff
@@ -0,0 +1,27 @@
...
+        for case in self.cases:
+            results = case['results']
+            data = {'result': results}
+            expected = case['expected']
+
+            claims = arrange_results([data])
+            claim = claims[0]
+            actual = claim['result']
+
+            self.assertEqual(expected, actual, case['description'])
```
eukreign: A loop like this makes it impossible to set breakpoints on a specific test case. Can you refactor this so that there is one line per test case which can accept a breakpoint in debugger if necessary. Even better if each test case is its own method. If there are 4 use cases then there should be 4 methods (descriptively named explaining what is being tested).

tau3:
@eukreign thank you for useful advices! They're implemented.

lbrynet/daemon/Daemon.py
```diff
@@ -163,6 +163,12 @@ def __call__(self):
         return d
 
 
+def arrange_results(claims):
```
eukreign: Can you rename this to something more specific such as `sort_claim_results`? Just `arrange_results` doesn't give you enough information on what it's for.

lbrynet/tests/unit/daemon/test_claims_comparator.py
```diff
@@ -0,0 +1,214 @@
...
+                    "txid": "aecfaewcfa"
+                },
+                {
+                    "height": 1,
+                    "name": "res",
+                    "claim_id": "ccc",
+                    "nout": 0,
+                    "txid": "fdsafa"
+                }
+            ])
```
eukreign: This is looking much better; you can reduce the code even more if you make the `expected` part above just be a list of `txid`s. So, in above case, the second argument to `run_test` would just be `["w5tv8uorgt", "aecfaewcfa", "fdsafa"]`.

Then in `run_test`:
```python
self.assertEqual(expected, [r['txid'] for r in actual])
```

tau3:
delivered :)

eukreign:
This has been reverted and needs to be fixed still. @nikooo777 can you please provide the exception that you were getting. @jackrobison can you provide some feedback on best way to resolve this.

nikooo777:
so, I called claim_list_mine and got this:
```
{
    "error": {
        "code": -32500,
        "data": [
            "  File \"/home/niko/.virtualenvs/lbrynet/local/lib/python2.7/site-packages/twisted/internet/defer.py\", line 651, in _runCallbacks",
            "    current.result = callback(current.result, *args, **kw)",
            "  File \"/home/niko/work2/repositories/lbry/lbrynet/daemon/Daemon.py\", line 168, in sort_claim_results",
            "    claim['result'].sort(key=lambda d: (d['height'], d['name'], d['claim_id'], d['txid'], d['nout']))",
            "exceptions.KeyError: 'result'"
        ],
        "message": "'result'"
    },
    "id": null,
    "jsonrpc": "2.0"
}
```
I was working with regtest but I guess this applies to live net too.
The steps I took to trigger this message are:
1) I created a channel
2) I used channel_export to export it
3) deleted the wallet
4) recreated one
5) imported the channel
6) published something with that channel
7) called claim_list_mine

Issue #1109, opened 2018-02-10
Calling `settings_set` writes invalid config in the daemon_settings.yml


hackrush01:
`daemon_settings.yml` before and after passing a valid setting(happens for any setting, either valid or invalid)
Specifically `lbryum_servers` field is overwritten in the python's list format instead of colon separated
```shell
$ cat ~/.lbrynet/daemon_settings.yml
 blockchain_name: lbrycrd_regtest
 download_directory: /home/hackrush/Downloads
 lbryum_servers:
 - localhost:50001
 reflect_uploads: false
 share_usage_data: false
 use_upnp: false

$ lbrynet-cli settings_set --download_directory="/home/hackrush/Downloads" > /dev/null

$ cat ~/.lbrynet/daemon_settings.yml
 blockchain_name: lbrycrd_regtest
 download_directory: /home/hackrush/Downloads
 lbryum_servers:
 - [localhost, 50001]
 reflect_uploads: false
 share_usage_data: false
 use_upnp: false
```

KamilPiekutowski:
I got this if you don't mind.

lyoshenka:
that file is created when you edit settings. you can create it yourself.

tzarebczan:
@hackrush01 did https://github.com/lbryio/lbry/pull/1104 possibly fi x this?

KamilPiekutowski:
@hackrush01 should I still look into this?

tzarebczan:
@KamilPiekutowski can you see if this still happens on 0.19 RC37 using his example. You should be able to create that file from scratch with his settings and see if it still messes up the sever list.

hackrush01:
@tzarebczan #1104 did not fix this.
@KamilPiekutowski The problem still persists, you're most welcome to look into it. It is most likely that the file `lbrynet/conf.py` is just dumping the data in python's dict format, whereas the data in `daemon_settings.yml` needs to be in yml(maybe json?)  format.

PR #1209, opened 2018-05-17
fixes bug when saving server list to conf file (issues #1109)

jleute:
When server lists (e.g. lbryum_servers) are loaded from the configuration file, the format address:port is converted to a tuple (address, port). When settings are changed and the configuration file gets overwritten, this conversion must be reversed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ at anytime.
   * `use_auth_http` in a config file being overridden by the default command line argument to `lbrynet-daemon`, now the command line value will only override the config file value if it is provided
   * `lbrynet-cli` not automatically switching to the authenticated client if the server is detected to be using authentication. This resulted in `lbrynet-cli` failing to run when `lbrynet-daemon` was run with the `--http-auth` flag
   * fixed error when using `claim_show` with `txid` and `nout` arguments
+  * fixed error when saving server list to conf file (issue #1109)
 
 ### Deprecated
   *
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -168,6 +168,8 @@ def server_port(server_and_port):
 def server_list(servers):
     return [server_port(server) for server in servers]
 
+def server_list_reverse(servers):
+    return ["%s:%s" % (server, port) for server, port in servers]
 
 class Env(envparse.Env):
     """An Env parser that automatically namespaces the variables with LBRY"""
@@ -267,7 +269,7 @@ ADJUSTABLE_SETTINGS = {
     'is_generous_host': (bool, True),
     'announce_head_blobs_only': (bool, True),
     'concurrent_announcers': (int, DEFAULT_CONCURRENT_ANNOUNCERS),
-    'known_dht_nodes': (list, DEFAULT_DHT_NODES, server_list),
+    'known_dht_nodes': (list, DEFAULT_DHT_NODES, server_list, server_list_reverse),
     'lbryum_wallet_dir': (str, default_lbryum_dir),
     'max_connections_per_stream': (int, 5),
     'seek_head_blob_first': (bool, True),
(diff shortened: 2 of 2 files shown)


eukreign:
Thanks for the pull request. Could you add some unit tests illustrating the problem and proving that this pull request solves it.

jleute:
The problem is best illustrated with a real config file that gets loaded, saved and loaded again, which results in an error. 
But I added a unit test that shows that the conversion reversal works properly.

tzarebczan:
@jleute thanks for this contribution! Please review our [contributing guide](https://lbry.io/faq/contributing) if you have not done so already. Also, [click here](https://lbry.io/faq/tips) if you are interested in an LBC tip for your work :)

eukreign:
@jleute thank you for adding the tests. Could you refactor the tests so that it's clear and straightforward to run the tests without your fix and see the test fail, then apply your fix and see it pass. Currently the test is calling out to the new method you added directly and not using `save_conf_file_settings()`.

lbrynet/tests/unit/test_conf.py
```diff
@@ -77,3 +79,19 @@ def test_data_dir(self):
...
+        conf.conf_file = conf_temp
+        adjustable_settings={'data_dir': (str, conf.default_data_dir),
+                'lbryum_servers': (list, [('localhost', 5001)],
+                    conf.server_list, conf.server_list_reverse)}
+        env = conf.Env(**adjustable_settings)
+        settings = conf.Config({}, adjustable_settings, environment=env)
+        conf.settings = settings
+        settings.load_conf_file_settings()
+        settings.save_conf_file_settings()
+        settings.load_conf_file_settings()
```
eukreign: The test needs to assert something. This just proves that no exceptions are thrown, but doesn't guarantee that the desired behavior is occurring.

lbrynet/tests/util.py
```diff
@@ -19,6 +19,10 @@
 
 log = logging.getLogger("lbrynet.tests.util")
 
+def create_conf_file(entry):
+    with tempfile.NamedTemporaryFile(delete=False, suffix='.yml') as conf:
+        conf.write(entry)
+        return conf.name
```
eukreign: When does the temporary file get cleaned up?

tzarebczan:
@eukreign can you check if the last commit satisfied your concerns? Do we still need a review from @jackrobison here?

tzarebczan:
@jackrobison / @hackrush01 please weigh in.

hackrush01:
Okay, I know this is probably a lot of reviews for the contribution(also we gave late feedback too, sorry), but there's one small thing that needs to be knocked out in the test.(The rest of the code works fine)

Here goes

This is the error when I comment out *all* the code in you fix in `conf.py`
```
===============================================================================
[ERROR]
  File "/home/hackrush/git-repos/lbry-repos/lbry/lbrynet/tests/unit/test_conf.py", line 86, in test_load_save_config_file
    conf.server_list, conf.server_list_reverse)}
exceptions.AttributeError: 'module' object has no attribute 'server_list_reverse'

test_conf.SettingsTest.test_load_save_config_file
-------------------------------------------------------------------------------
Ran 1 tests in 0.077s
```

As you see it says `exceptions.AttributeError: 'module' object has no attribute 'server_list_reverse'` that means the test is expecting that there is a function in `conf.py` by that name. But there isn't because *all* of the fix was commented out(to replicate conditions before the fix)

This is the exact line causing the test to fail(after fix is commented out)https://github.com/lbryio/lbry/pull/1209/files#diff-f0f237c9e78a1b113640af03f121e1faR86

Now, basically two things are expected from the PR(specifically test)

1. fix commented out should error with: Expected "localhost:50001" but got "[localhost, 50001]".
2. fix commented back in should pass the test.
If test that is written is high level that makes no assumptions about internal implementation details, then it should work like above

It will create some config, update it, check what the value is, fail that it's a list instead of colon separated string then after the fix the same test should pass because it's a colon separated string

I hope that makes thing clear.

PR #1211, opened 2018-05-21
Refactor conditions to add a contact to the routing table

jackrobison:
* changed the bucket splitting condition in the dht routing table to be more aggressive (https://stackoverflow.com/a/32187456)
  * ping dht nodes who have stored to us periodically to determine whether we should include them as an active peer for the hash when we are queried. Nodes that are known to be not reachable by the node storing the record are no longer returned as peers by the storing node.
  * temporarily disabled data price negotiation, treat all data as free
  * changed dht bootstrap join process to better populate the routing table initially
  * cache dht node tokens used during announcement to minimize the number of requests that are needed
  * implement BEP0005 dht rules to classify nodes as good, bad, or unknown and for when to add them to the routing table (http://www.bittorrent.org/beps/bep_0005.html)
  * refactored internal dht contact class to track failure counts/times, the time the contact last replied to us, and the time the node last requested something fom us
  * refactored dht iterativeFind
  * sort dht contacts returned by `findCloseNodes` in the routing table
  * disabled Cryptonator price feed

  * added functional tests for bootstrapping the dht, announcing and expiring hashes, finding and pinging nodes, protocol version 0/1 backwards/forwards compatibility, and rejoining the network
  * added protocol version to dht requests and to the response from `findValue`
  * added `port` field to contacts returned by `routing_table_get`


tzarebczan:
@jackrobison you are reviewing this with @lyoshenka correct?

PR #1213, opened 2018-05-24
added certifi package for Twisted SSL verification on Windows

akinwale:


tzarebczan:
No more errors on Windows!

lbrynet/daemon/DaemonControl.py
```diff
@@ -1,3 +1,12 @@
+import certifi
+import os
+import sys
+
+# Set SSL_CERT_FILE env variable for Twisted SSL verification on Windows
+# This needs to happen before anything else
+if 'win' in sys.platform:
+    os.environ['SSL_CERT_FILE'] = certifi.where()
+
```
eukreign: Seems a bit dirty to put this up in the imports. Is that the only way to do it? (if that's the only way then the code looks good and I can merge it)

eukreign: What if instead of putting this in code it gets set in the environment as an actual environment variable?

akinwale: Pretty much, yes, according to the usage doc at https://pypi.org/project/certifi/

> What if instead of putting this in code it gets set in the environment as an actual environment variable?
Then we may have to distribute the `ca-cert.pem` file with the executable.

eukreign:
@akinwale  Can you move the import statement into the if statement, this way it's not imported on OSes other than Windows.

PR #1216, opened 2018-05-25
fix error in claim_list_mine

tau3:


tzarebczan:
Thanks for following up on this @tau3 !

eukreign:
@nikooo777 does this solve the problem?

nikooo777:
I can test the case that broke the daemon but in regards of the code It's up to either your or @jackrobison to approve.

I'll let you know ASAP

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ at anytime.
   * download blockchain headers from s3 before starting the wallet when the local height is more than `s3_headers_depth` (a config setting) blocks behind
   * track successful reflector uploads in sqlite to minimize how many streams are attempted by auto re-reflect
   * increase the default `auto_re_reflect_interval` to a day
+  * `claim_list` and `claim_list_mine` in Daemon `return` sorted results
```
jackrobison: "return" doesn't need code tags

lbrynet/daemon/Daemon.py
```diff
@@ -163,6 +163,11 @@ def __call__(self):
         return d
 
 
+def sort_claim_results(claims):
+    claims.sort(key=lambda d: (d['height'], d['name'], d['claim_id'], d['txid'], d['nout']))
+    return claims
```
jackrobison: This doesn't need to return, the provided `claims` list is sorted in place.

eukreign: `sort_claim_results` is provided as a Deferred callback so if it doesn't return then the subsequent callbacks will not get the claims.

eukreign:
@tau3 looks like you just need to make the fix to the CHANGELOG per @jackrobison comment

@jackrobison other than the CHANGELOG does the implementation look correct to you?

lbrynet/daemon/Daemon.py
jackrobison: Ok, makes sense

tzarebczan:
Yep, let's merge and I'll open a separate issue for allowing sorting arguments, similar to what we did recently for file_list.

tzarebczan:
Thanks for patching this us @tau3! Same as with the last PR, please see https://lbry.io/faq/tips if you are interested in a tip for this contribution :)

PR #1217, opened 2018-05-28
docker: Lbrynet as a container

chamunks:
## Documentation 

Added some basic documentation to docker/README.md it can be used elsewhere but at some point if we add other things which end up justifying that the end user run a `git clone` and then `docker-compose up -d` from inside of the git repo of lbrycrd we can let them store data in docker/data/ with some fancy git ignoring.

## Docker Compose
For the sake of getting people launched asap I've included a basic docker-compose.yml file which can launch the app with `docker-compose up -d` then commands can be run using `docker exec CONTAINERNAME lbrynet-cli commands`

## .gitlab-ci.yml
This is for CI/CD prebuild and deploy if we want to use GitLab for building these containers rather than the hub.docker.com route.  Currently, this is just configured to show what it would look like with my repositories.

## Dockerfile

Currently, the dockerfile is very basic and just installs the daemon into the containers $PATH for easy execution then it stores the daemons content in /app/ as that is defined as the unprivileged daemon processes home directory. 

## Commit size
https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention the version I was working from in my topic branch was hugely out of date so rather than carrying over all of the commits I just added everything fresh I know that it's lazy but this code doesn't meddle with anyone's stuff so it seemed like it should be okay to do it this way.

## Maintainance

This container should be reasonably future proof for now as I've defined all non-ephemeral blobs to reside in the /app/ directory inside of the container which can be updated at a later time just fine.

The configuration will be added using a shell script in the future which can be optionally included.  The goal is to allow configuration using environment variables in the docker-compose.yml file which will be the officially supported launch method as it's the least maintenance method.


chamunks:
Someone will likely need to setup a lbryio account on https://hub.docker.com and get this repository on there.  I can do it if someone wants me to do this but I don't want to take over the lbryio namespace personally.

Right now we won't have any fancy tags or anything yet as I have no idea what kind of setup we would really want for that.  Also the README.md in the docker directory should probably be used to some degree for the Docker hub page readme.

kauffj:
Great stuff @chamunks! Not qualified to review this really, but did catch a spee.ch reference in the docs, presumably from a copy/paste.

eukreign:
Thanks for the contribution @chamunks . From the README it appears that your goal is to create a docker container for spee.ch? If that is the case then it would probably make sense to do this pull request on the speech repository (https://github.com/lbryio/spee.ch).

Regardless of the end goal I would suggest that you first make a repository under your own github account and then setup a personal hub.docker.com account to get the automated builds working (both are free and easy to setup). Once you have something working then we can review it and decide the best way to go forward (which repo to add your Docker files to) and whether it makes sense to add your docker image to lbry on docker hub.

Furthermore, if the goal is specifically to build docker images for spee.ch that others can customize then maybe the solution here is to just do a tutorial or write some documentation on how to do it (and paste the code in this pull request as examples in the tutorial/docs)? Also, I expect there would be different images for production release vs. development of spee.ch. I'd encourage you to get some feedback from @billbitt on all this and collaborate with him.

If I misunderstood and your goal isn't to build container for spee.ch but for lbry in general then perhaps you should take a look at several existing efforts at LBRY to build docker containers:

https://github.com/lbryio/lbry-in-a-box
https://github.com/lbryio/lbryumx (https://hub.docker.com/r/lbry/lbryumx/)

chamunks:
Crap the only reference to spee.ch in the Readme was supposed to be mentioning that I need this accepted eventually to build a container cluster for Spee.ch starting with lbrynet.  I can re-submit with fixes in a few hours.  This container itself is intended for lbrynet if you look over the Dockerfile and give me some feedback that would be wonderful.

eukreign:
Some questions:
- How is this different from the other two repos I linked to earlier?
- Would you consider contributing to one of the other lbryio repos instead (lbrumx or lbryum-in-a-box)?
- What are some general purpose use cases for having lbrynet in a container by itself?

We want to avoid having code in lbry repo that isn't being used by lbry developers (because it adds to technical-debt and if we're not using it, it will not be maintained). I'd prefer that it go in a different repo where it will serve a specific purpose and get regularly maintained (by those using it).

tzarebczan:
@kauffj any comments on the above suggestions from Lex as far as where this code should live? 

@eukreign this first started out as an effort to dockerize spee.ch, which is when @chamunks began to dockerize lbrynet as it's the prime component. This docker image will be critical if we plan to dockerize more applications/services on LBRY and as a standalone package for those who prefer to install docker images when running the daemon alone.

eukreign:
@tzarebczan there are too many ways to dockerize something depending on use cases and needs. for example, the Dockerfile in this pull request pulls the daemon from a release page, therefore there isn't even any benefit to having this Dockerfile in the lbry repo, it can live anywhere and still work.

There are many ways to do a Dockerfile:
- fetch and build from a released zip file (like the on in pull request)
- git pull latest master during Dockerfile build (like the lbry-in-a-box does it)
- build directly from source in which the Dockerfile lives (this is a common use case for development)

Since the Dockerfile in the pull request doesn't benefit from being in the lbry repo and isn't used as part of lbry daemon development I don't think it makes sense for it to go in this repo.

My recommendation would be to create a new repo (based on or even directly forked from lbry-in-a-box) and call it something like `lbry-cluster` and put everything you need in there to build a lbry cluster.

kauffj:
I'm ambivalent. The two sides:

- On one hand, Docker is a generally useful technology that lots of people like to use. If we want people to build things on top of LBRY, and Docker makes it easier to do it, providing Docker set up increases the amount of building that happens. 
- On the other, I agree with @eukreign that adding Docker both: a) increases costs, since this code needs to be maintained and b) risks breaking if we're not using it ourselves.

From here I would suggest three things:
1) We look at designs that isolate Docker code outside of core components. I have no idea if this is smart or stupid (I know very little about Docker). But it keeps this code cleanly separate.
2) We look at the ability for external community members to specifically own Docker maintenance, hopefully eliminating the cost on core devs.
3) @lyoshenka @nikooo777 @billbitt consider whether it makes sense for us to make larger usage of Docker internally, which resolves some of the tensions above.

kauffj:
@chamunks I'm not pushing lbry-in-a-box as a solution. Can you comment on how orthodox or unusual it is for Docker code to be maintained inside vs. outside the repository it is for?

eukreign:
I'll just add two more thoughts:

1. If the lbry dev team decides to use docker during development this docker container would very likely be specifically built for development, it would build from local source (instead of fetching pre-build releases) and it would have security and performance settings disabled/optimized for development instead of production deployment. It would be more complicated to have one Docker file that does everything than to just have two purpose built Dockerfiles with one optimized for development and one optimized for production / specific use cases. The production / specific use case Dockerfiles should live in designated repos for that specific use case (functional testing, spee.ch deployment, etc).

2. None of the popular Docker containers on hub.docker.com keep the Dockerfile in the project repository. Usually the opposite is true, the Dockerfile is maintained by Docker experts in the community in their own private repos or by Docker organization (for "official" packages, see: https://github.com/docker-library/official-images ).

3. `lbry-in-a-box` is an excellent template for creating various custom deployment solutions for LBRY using Docker. It already does everything, so to customize it is just a matter of removing things and stripping it down to just what you need.

4. A repo with purpose built collection of Dockerfile's and `docker-compose.yml` is much more user friendly. If someone says they want to deploy spee.ch you can just point them to a repo that they check out and `docker-compose up` and they are live! Rather than having to mess around with a potentially generic / developer friendly Dockerfile from the main lbry repo.

chamunks:
## Clarification

Certainly, also I definitely don't dislike the concept of the LIAB, in general, I love instant-deploy stacks.  However, I do think that it should be easier if we have a stack of microservices built for each of the respective building blocks within LBRY.

## More direct answer.

To answer your question of how normal it is to keep a Dockerfile inside of the repository of the package itself.  I've been noticing a trend from my perspective where a lot of software in the OpenSource ecosystem is starting to include base Dockerfiles.

Ideally at some point we should consider contributing to the official Docker library and get our containers published in the [docker official library](https://docs.docker.com/docker-hub/official_repos/) however I think that we should work on getting a bit closer to adopting it internally before we consider taking that step.

## Further expansion

Further expansion, there are two schools of thought here.

#### Replacement of current dev cycle
As you mentioned the Docker containers could eventually replace things like the CI/CD process as is which I think they are very good as they are currently.  
##### Pros
- We have a dev cycle that is completely independent of LBRY's internal dev cycle architecture.
- People would be completely equipped to build their own internal code of our repositories.

##### Cons
- Added complexity/Increase of Technical Debt
- More dev hours sunk into something that's currently already doing a good job.

#### Organic Adoption
Or we could go off of simple release builds exclusively.  There are benefits and drawbacks to both options.  The way that I've designed this container however, is generally fairly future proof as I've followed a good number of the Docker best practices as best as I can and will plan to continue to add more eventually.

##### Pros
- We can work out the details as we go and tend to prefer something that doesn't exactly have to be perfect directly out of the box.
- We could see who internally would be highly allergic to the concept of moving to this methodology and figure out why organically.
##### Cons
- Similar issues to the above.
- We would probably end up having to spend more time developing it this way in the long run.

chamunks:
@eukreign you have some very good points.  Granted, I think there is something extra to be added to some of them.

## Point number 4
Mentioning the docker containers in a purpose built repository sounds like it would be more user friendly if you assume that the end users or community want to maintain the containers themselves.  My own learning curve getting my head around the complexity and nuances of how just the lbrynet-daemon itself should be a good indicator of how complex it is currently to get ones head around the component itself.

## Point number 2
I wouldn't go so far as to use generalized statements such as `None of the popular Docker containers on hub.docker.com keep the Dockerfile in the project repository. ` We could absolutely add our Dockerfile to that stage but as you mentioned in `point number 1` it would make a lot of sense to have two container files.  You could go so far as to have one for development and one for deployment.  In a world where trusting others should generally be something you're averse to as a generally good policy when it comes to infosec there is a lot to be said for building the binary from source.  That said it would be a little absurd to expect everyone to want to build the binaries from source all the time.

All in all I think that we're generally on the same page with the direction the Docker stuff should go in some regards.  I really appreciate the application of critical thinking and sticking to the keep it simple ethos.

Looking forward to the next responses.

chamunks:
## Official Docker Repository Proposal Criteria
#### Generally Useful
If we want any Docker repositories in the Official Docker Library we will need to prove the general usefulness of our repositories as well.  Since this project is still fairly new it might be difficult to prove this just yet.

#### Security Responsibility
We would adopt a new level of concern having to monitor the supply chain of code we use and monitor any CVS's that are published using anything that we utilize.

## Conclusion
I definitely agree that our end goal should definitely be to publish to the official repositories as it gives us a level of extra cred in the Docker ecosystem, however, I think that it should definitely be further down the line.

chamunks:
Sounds good.  However, the only issue that remains is in order to get a spee.ch repo up I would require one with this repo first.

eukreign:
The Dockerfile you have in this pull request does not depend on this repo, it pulls the source from a release build. As I've said a few times, take a look at `lbry-in-a-box`: it achieves all of the things you're trying to do within a single dedicated repository.

Furthermore, I'd suggest that the first container you make for spee.ch is just a single docker image with several services running in one container. I realize this goes against best practices but it has the advantage that a user of this does not need to know much about docker and no longer even needs `docker-compose` to try it. Also, newer versions of docker appear to even support this out of the box (see `--init` flag): https://docs.docker.com/config/containers/multi-service_container/  I've only used the second option (`supervisord`) before but the new way seems rather convenient as well.

Once you have the single container version working and documented (targeted at new users of docker and lbry) we can look at a more comprehensive production deployable version with `docker-compose` and multiple containers (one for each service).

Does this seem reasonable to you as a starting point? Can you put together such a multi-process single-image container for spee.ch?

chamunks:
@eukreign I'm looking to still try and aim for stateless docker practice adherent containers but this is the one I've got started here.

[https://github.com/chamunks/lbry-docker/](https://github.com/chamunks/lbry-docker/)

So far the speech container is the biggest focus as I think its an important focus.  Eventually I would like to expand on the LBRY-Daemon container a bit more but for the most part, the way it is will basically be fairly easy to add configuration and installation magic.  

So far the status of the spee.ch container itself is not currently working as there are some issues launching the main process.  The method I'm using to start the cluster is `docker-compose up -d` from within the `$REPOPATH/www.spee.ch/`

The situation is that basically, I want to be able to deploy an instantly functional container cluster.  Then if the end user wants to add things like configurations and custom `js/css` in volumes at `/app/` then it shouldn't overwrite those changes which currently that stuff functions.  I have the goal of also re-asserting directory ownerships and permissions but my current system needs some tweaking as I've not pulled it from a git repo before but frequently use these little scripts from [https://github.com/chamunks/docker-support/](https://github.com/chamunks/docker-support/)

You'll have to excuse my brevity I'm writing this reply while on the go.

eukreign:
@chamunks nice work, keep going and keep us posted on your progress!

@billbitt does anything in @chamunks update above ring a bell as easy fixes that you might suggest?

eukreign:
@chamunks was checking on your progress and wanted to share the issue you had created here to help anyone else interested in this follow along: https://github.com/chamunks/lbry-docker/issues/1

chamunks:
Thanks @eukreign I'll have more time this coming week to get back to it I've had other obligations this past week.  

Also, keep in mind that I'm mostly only interested in the Docker stuff for the sake of improving Lbry's offerings and not necessarily because I want to host anything personally any time soon so I'm asking you guys a lot of questions to see what would fit with you all.

eukreign:
I think it's really important to [eat your own dog food](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), maybe you can find someone on [LBRY discord](https://chat.lbry.io/) that wants to do a spee.ch type site and work with them to get one running with docker? @billbitt @jsigwart @kauffj

chamunks:
Mentioning dogfooding is definitely part of the reason I'm here mentioning this stuff.  I'm getting to a bit of a point here where it's starting to seem pointless to chase this project down.

PR #1230, opened 2018-06-04
adds publish_time to source on publish

shyba:
Depends on https://github.com/lbryio/lbryschema/pull/46


eukreign:
@shyba says to close

PR #1238, opened 2018-06-09
Publish command accepts decimal string for bid.

eukreign:


kauffj:
@eukreign is it possible to still get this into 0.20? Otherwise need to patch this somewhere else so we can move forward with lbry.tech.

eukreign:
I don't see any reason it can't make it into .20. It should be backwards compatible; we can make it required at a later time.

@jackrobison can you do a sanity check on this PR. also, do we have a way to respond with a deprecation warning when someone sends float amounts?

jackrobison:
lgtm

Issue #1165, opened 2018-04-05
Allow update of claims even if they don't exist in file_list


tzarebczan:
## The Issue
If you try to update a claim that hasn't been downloaded (doesn't exist in file_list), you'll see the below error and won't be able to update. As a workaround, the file needs to be downloaded first.

Edit: What it actually does is create a new claim with the same name, which causes a bigger problem for future updates because now there are 2 of the same claim names and it doesn't know which to update. 

```
2018-04-05 10:02:37,832 WARNING  lbrynet.daemon.auth.server:228: error processing api request: stream not found
traceback:   File "site-packages\twisted\internet\defer.py", line 1297, in _inlineCallbacks
  File "site-packages\twisted\python\failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages\lbrynet\daemon\Daemon.py", line 2158, in jsonrpc_publish
  File "site-packages\twisted\internet\defer.py", line 1297, in _inlineCallbacks
  File "site-packages\twisted\python\failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages\lbrynet\daemon\Daemon.py", line 717, in _publish_stream
  File "site-packages\twisted\internet\defer.py", line 1297, in _inlineCallbacks
  File "site-packages\twisted\python\failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages\lbrynet\daemon\Publisher.py", line 62, in publish_stream
  File "site-packages\twisted\internet\defer.py", line 1297, in _inlineCallbacks
  File "site-packages\twisted\python\failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages\lbrynet\database\storage.py", line 672, in save_content_claim
  File "site-packages\twisted\python\threadpool.py", line 246, in inContext
  File "site-packages\twisted\python\threadpool.py", line 262, in <lambda>
  File "site-packages\twisted\python\context.py", line 118, in callWithContext
  File "site-packages\twisted\python\context.py", line 81, in callWithContext
  File "site-packages\twisted\enterprise\adbapi.py", line 477, in _runInteraction
  File "site-packages\twisted\enterprise\adbapi.py", line 467, in _runInteraction
  File "site-packages\lbrynet\database\storage.py", line 654, in _save_content_claim
exceptions.Exception: stream not found

```

### Steps to reproduce
1. Create claim
2. run file_delete on claim name
3. edit claim

### Expected behaviour
Should be able to update the claim even if you don't have file_list info

[...]

PR #1250, opened 2018-06-14
fix database error when updating a claim that we don't have a file for

jackrobison:
https://github.com/lbryio/lbry/issues/1165


tzarebczan:
Tested and working without error now!

Issue #1243, opened 2018-06-13
Change documentation build process to support lbry.tech


kauffj:
- @Krisseck has a modified version of `gen_docs.py` that will work for usage with Slate.
- His script will create `api.md`, a markdown file that lbry.tech will use to generate documentation via Slate
- Script should be run automatically or by hand on all production releases.

kauffj:
@Krisseck will update this ticket when he has opened a PR with his modification

eukreign:
We have a script now to generate a JSON formatted API that feeds into lbry.tech.

PR #1253, opened 2018-06-15
Make API docs suitable for Slate

Krisseck:
Per issue #1243, this generates api.md and cli.md in the root of the project folder, which will used as a base for Slate documentation.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,4 @@ txJSON-RPC==0.5
 wsgiref==0.1.2
 zope.interface==4.3.3
 treq==17.8.0
+tabulate==0.8.2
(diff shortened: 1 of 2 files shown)


kauffj:
@eukreign is this mergable? What is status of this?

@NetOperatorWibby is waiting on green light from daemon team to finish API page

eukreign:
I generated a new `api.md` using the latest script and it's still pretty far from what's here: https://github.com/lbryio/lbry.tech/blob/master/content/.vuepress/public/api.md

I will continue working on it from where @Krisseck left off.

eukreign:
After discussing this with @NetOperatorWibby we decided it would be better to generate HTML directly instead of having an intermediate step of Markdown. Markdown is great if humans will be editing the file but in the case of the API it's completely generated and not intended to be interacted with directly. Further discussion here: https://github.com/lbryio/lbry.tech/issues/118

Issue #1245, opened 2018-06-14
Header download can lock startup if one server is out


shyba:
## The Issue

In order to start the daemon, 
as a user, 
I want to be able to start even if one of the wallet servers is out.

### Steps to reproduce
1. add `does.godzilla.exists.lbry.io:50001` to the list of lbryum servers (with the working ones)
2. try to start the daemon

### Expected behaviour
It tells this server is bad, but connects to the others and go ahead.

### Actual behaviour
It connects to all servers, but the one that doesnt reply doesnt release the startup code.

## Anything Else
It's possibly caused by `txlbryum/client.py` and `txlbryum/factory.py`. They fire a callback on successful connection, but there is no code for when the connection is lost or the server doesnt even resolve.

tzarebczan:
@jackrobison / @shyba We should try to fix this before a 0.20 patch release if possible.

tzarebczan:
Related: https://github.com/lbryio/lbry/issues/1084

PR #1260, opened 2018-06-18
fix wallet startup locking up if a lbryumx server is unavailable

jackrobison:
fixes https://github.com/lbryio/lbry/issues/1245


tzarebczan:
Tested and working as expected.

Issue #1172, opened 2018-04-27
Download stops after first blob, payment sent, restarts in stopped status


tzarebczan:
## The Issue
This has happened to me twice (update, now 3 times!) now, not sure how to reproduce it consistently.  

```
2018-04-27 15:08:20,618 INFO     lbrynet.daemon.Downloader:93: Download lbry://two status changed to initializing
2018-04-27 15:08:20,624 INFO     lbrynet.daemon.ExchangeRateManager:221: Converting 50.000000 LBC to LBC, rates: [Currency pair:('USD', 'BTC'), spot:0.000108066589509, ts:1524856044, Currency pair:('BTC', 'LBC'), spot:29350.7670639, ts:1524856044, Currency pair:('BTC', 'LBC'), spot:29351.9505936, ts:1524856044, Currency pair:('USD', 'BTC'), spot:0.00010799, ts:1524855145, Currency pair:('BTC', 'LBC'), spot:29239.7660819, ts:1524856045]
2018-04-27 15:08:20,624 INFO     lbrynet.daemon.ExchangeRateManager:221: Converting 10.000000 LBC to LBC, rates: [Currency pair:('USD', 'BTC'), spot:0.000108066589509, ts:1524856044, Currency pair:('BTC', 'LBC'), spot:29350.7670639, ts:1524856044, Currency pair:('BTC', 'LBC'), spot:29351.9505936, ts:1524856044, Currency pair:('USD', 'BTC'), spot:0.00010799, ts:1524855145, Currency pair:('BTC', 'LBC'), spot:29239.7660819, ts:1524856045]
2018-04-27 15:08:20,625 INFO     lbrynet.daemon.Downloader:93: Download lbry://two status changed to downloading_metadata
2018-04-27 15:08:20,973 INFO     lbrynet.core.client.BlobRequester:414: Offered rate 0.000000/mb accepted by 18.221.194.193
2018-04-27 15:08:20,973 INFO     lbrynet.core.client.BlobRequester:487: Requesting blob f204bb18354b803c17ecbfd86e796a3ec4fd0097506812a97189a6ffe6f2a9f41dcfd02463f33a587015ffef870bf0fe from 18.221.194.193:3333
2018-04-27 15:08:21,086 INFO     lbrynet.core.client.BlobRequester:563: Blob f204bb18354b803c has been successfully downloaded from 18.221.194.193:3333
[...]

PR #1262, opened 2018-06-19
fix download being stopped if the sd blob downloaded and data did not start within the timeout

jackrobison:
fixes https://github.com/lbryio/lbry/issues/1172


tzarebczan:
Thanks Jack, squashed another nasty one with a one-liner haha.

PR #1266, opened 2018-06-21
add download progress and is downloading flag to daemon status

akinwale:


lbrynet/daemon/Daemon.py
```diff
@@ -1053,6 +1053,8 @@ def jsonrpc_status(self, session_status=False):
                     'blocks': local blockchain height,
                     'blocks_behind': remote_height - local_height,
                     'best_blockhash': block hash of most recent block,
+                    'headers_download_progress': the download progress of the blockchain_headers file,
```
shyba: Maybe show only that? While testing I noticed it goes back to 0 when download is skipped (unnecessary).
I would suggest something like (programming in comments reads bad, feel free to implement it readable):
`headers_download_progress = headers_download_progress if is_downloading_headers else ((local_height/remote_height) if remote_height else 100)`

shyba:
Worked fine on testing a new download, but for a synced daemon startup I got this:
```
{                                                                                                                                                                                                          [7/1878]
  "blockchain_status": {                                                                                                                                                                                           
    "best_blockhash": "6a5ffe2de394b19def1f9620212586b8af91b3126d5d201ea3bd949705c1572a",                                                                                                                          
    "blocks": 390714,                                                                                                                                                                                              
    "blocks_behind": 0,                                                                                                                                                                                            
    "headers_download_progress": 0,                                                                                                                                                                                
    "is_downloading_headers": false                                                                                                                                                                                
  },     
```
Left a suggestion on how to avoid that, but basically I just wanted to reduce the amount of fields here. Aside from this, looks good.

lbrynet/core/Wallet.py
```diff
@@ -105,14 +110,25 @@ def fetch_headers_from_s3(self):
         elif final_size_after_download and not final_size_after_download % HEADER_SIZE:
             s3_height = (final_size_after_download / HEADER_SIZE) - 1
             local_height = self.local_header_file_height()
+
             if s3_height > local_height:
+                self.headers_download = True
```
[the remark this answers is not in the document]
akinwale: The reason I set it here is to indicate that the download is going to happen before the file actually starts downloading. Setting it at L123 alone would set it to true after the download has started.

lbrynet/daemon/Daemon.py
akinwale: I thought about this initially, and having the `is_downloading_headers` flag would help with differentiating between a first time run (no `blockchain_headers` file present), and a subsequent run (when it's trying to catch up with blocks).

lbrynet/daemon/Daemon.py
```diff
@@ -1067,12 +1070,18 @@ def jsonrpc_status(self, session_status=False):
...
         # on startup, the wallet or network won't be available but we still need this call to work
-        has_wallet = self.session and self.session.wallet and self.session.wallet.network
-        local_height = self.session.wallet.network.get_local_height() if has_wallet else 0
-        remote_height = self.session.wallet.network.get_server_height() if has_wallet else 0
-        best_hash = (yield self.session.wallet.get_best_blockhash()) if has_wallet else None
-        wallet_is_encrypted = has_wallet and self.session.wallet.wallet and \
+        has_wallet = self.session and self.session.wallet
+        downloading_headers = self.session.wallet.is_downloading_headers() if has_wallet else False
+        has_wallet_with_network = has_wallet and self.session.wallet.network
+        local_height = self.session.wallet.network.get_local_height() if has_wallet_with_network else 0
```
jackrobison: If the wallet has not been initialized yet all of these will throw attribute errors. To fix this, reverse the `if`, ie: `local_height = 0 if not has_wallet_with_network else self.session.wallet.network.get_local_height()`

PR #1301, opened 2018-07-13
Fix high cpu usage on incomplete streams

shyba:
If the connected peer doesnt have a blob we need we would trigger the same request in a loop, causing a high CPU usage for the time the download is stuck.

### DIFF
--- a/lbrynet/core/client/BlobRequester.py
+++ b/lbrynet/core/client/BlobRequester.py
@@ -354,6 +354,10 @@ class AvailabilityRequest(RequestHelper):
         log.debug("Received a response to the availability request")
         # save available blobs
         blob_hashes = response_dict['available_blobs']
+        if not blob_hashes:
+            # should not send any more requests as it doesnt have any blob we need
+            self.peer.update_score(-10.0)
+            return True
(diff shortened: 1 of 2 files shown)


lbrynet/core/client/BlobRequester.py
```diff
@@ -354,6 +354,10 @@ def _handle_availability(self, response_dict, request):
         log.debug("Received a response to the availability request")
         # save available blobs
         blob_hashes = response_dict['available_blobs']
+        if not blob_hashes:
+            # should not send any more requests as it doesnt have any blob we need
```
jackrobison: What if it has other blobs for other streams? Won't this punish it globally for not having the blobs for this stream?

tzarebczan: @shyba ? Is this something we want to get into the next patch release?

shyba: Yes, we just talked about it on the standup. The score seems local to the blob requester, since each file is a downloader and each downloader has its own requester. Jack will take a look on it and see if its ok to do this way or if there is a better approach.

PR #1309, opened 2018-07-19
A More Informative Error Message

Invariant-Change:
## A More Informative Error Message

Currently, when a bid is **pending**, you can not increase another bid unless you have the **total amount** in your **balance**. You have to **wait** for any **pending** claims to **resolve** before you can do this.

The current error message, in this case, can give the **impression** that the app is **broken**.

This new error message **amendment** will prevent **angry** users from launching vigilante attacks on the LBRY HQ and mass riots!

### When Pending:
image

### When Resolved

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ can and probably will change functionality and break backwards compatability
 at anytime.
 
 ## [Unreleased]
+
+## [0.20.3] - 2018-07-20
+### Changed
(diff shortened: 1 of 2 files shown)


Invariant-Change:
Done: https://github.com/lbryio/lbry/pull/1310

tzarebczan:
@eukreign can you fix this on your end? @Invariant-Change is not running pycharm/python IDE, this was a simple text change via the IDE.

PR #1310, opened 2018-07-19
A More Informative Error Message

Invariant-Change:
Additional information added to the balance error message when editing a claim.


tzarebczan:
@Invariant-Change please keep the changelog in the same PR as the original. Also, there should be an unreleased section of the changelog that the entry needs to go into.

PR #1315, opened 2018-07-27
refactor mirroring

shyba:
While trying to solve a bug reported by @tzarebczan and @lyoshenka I saw it wasn't behaving properly with two layers of error handling and complexity was above desired, so I simplified error handling to one layer and used a semaphore instead of the looping call. This version downloads the next 2 blobs concurrently with peers, stops immediately when request and handle better cases where p2p wins (or some other part of our code asks writers to stop).
I allowed 2 concurrent downloads because that's what worked faster on my bad connection, but I'm happy to change back to 1 if necessary.

### DIFF
--- a/lbrynet/core/HTTPBlobDownloader.py
+++ b/lbrynet/core/HTTPBlobDownloader.py
@@ -3,7 +3,8 @@ import logging
 
 from twisted.internet import defer
 import treq
-from twisted.internet.task import LoopingCall
+
+from lbrynet.core.Error import DownloadCanceledError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


lbrynet/core/HTTPBlobDownloader.py
```diff
@@ -21,65 +22,76 @@ def __init__(self, blob_manager, blob_hashes=None, servers=None, client=None):
...
+    @defer.inlineCallbacks
+    def _write_blob(self, writer, blob):
+        response = yield self.client.get(url_for(choice(self.servers), blob.blob_hash))
+        if response.code != 200:
+            log.debug('[Mirror] Missing a blob: %s', blob.blob_hash)
+            if blob.blob_hash in self.blob_hashes:
+                self.blob_hashes.remove(blob.blob_hash)
+            defer.returnValue(False)
+
+        log.debug('[Mirror] Download started: %s', blob.blob_hash)
```
jackrobison: the logging formatter will include the file and lineno, so `[Mirror]` isn't necessary

shyba: :man_facepalming: yes, of course! fixing it!

lbrynet/core/HTTPBlobDownloader.py
```diff
@@ -21,65 +22,76 @@ def __init__(self, blob_manager, blob_hashes=None, servers=None, client=None):
...
+        dl = []
         for blob_hash in self.blob_hashes:
             blob = yield self.blob_manager.get_blob(blob_hash)
             if not blob.verified:
-                self.download_blob(blob)
-                return
-        self.stop()
+                dl.append(self.semaphore.run(self.download_blob, blob))
+        self.deferreds = dl
+        yield defer.DeferredList(dl, consumeErrors=True)
```
jackrobison: Do you mean to consume all errors here? What errors can be raised?

```
d = self.semaphore.run(self.download_blob, blob)
d.addErrback(lambda err: err.check(defer.TimeoutError, defer.CancelledError))
dl.append(d)
```

PR #1325, opened 2018-07-28
Use binascii hexlify/unhexlify in storage instead of encode/decode

hackrush01:


eukreign:
This is already done in `lbryum-refactor` branch and it's not possible to hit this issue unless you run the new wallet in python 3, create some claims and then switch to the old wallet in python 2 and try to view those claims.

Upgrading from old wallet+py2 `lbrynet` to the new wallet+py3 `lbrynet` is not a reversible operation so we don't need to support new wallet stuff in the old wallet.

PR #1326, opened 2018-07-28
fixed improper error handling when data is not valid JSON

hackrush01:


jackrobison:
What does this fix?

lbrynet/daemon/auth/server.py
```diff
@@ -298,12 +298,13 @@ def _render_error(self, failure, request, id_):
...
+            if not failure.check(ComponentsNotStarted, ComponentStartConditionNotMet):
+                log.warning("error processing api request: %s\ntraceback: %s", error.message,
+                            "\n".join(error.traceback))
         else:
             # last resort, just cast it as a string
             error = JSONRPCError(str(failure))
-        if not failure.check(ComponentsNotStarted, ComponentStartConditionNotMet):
-            log.warning("error processing api request: %s\ntraceback: %s", error.message,
-                        "\n".join(error.traceback))
+
```
eukreign: Above code fails in cases where `failure` is not a `Failure` and thus there is no `.check()` method. Instead it is moved to be inside `if isinstance(failure, Failure)` where `failure.check()` works.

eukreign: @jackrobison ^

jackrobison: :+1:

PR #1339, opened 2018-08-05
fix tracebacks for startup and shutdown errors

jackrobison:


eukreign:
Can the start (setup) / stop methods go back to being inline callbacks but using the new decorator?

lbrynet/daemon/Component.py
```diff
@@ -50,26 +51,26 @@ def stop(self):
...
-            log.exception("Error stopping %s", self.__class__.__name__)
-            raise err
+            return result
+
+        def handle_error(err):
+            log.error("Error stopping %s\n%s", self.component_name or self.__class__.__name__, err.getTraceback())
+
+        return maybe_deferred_trap_and_trace(
+            (defer.CancelledError, AlreadyQuit), set_running, handle_error
+        )(self.stop)()
```
eukreign: `)(self.stop)()` looks a bit gnarly :-D

Can `_setup`/`_stop` go back to being `@inlineCallbacks` with the new decorator you've written?

jackrobison:
will redo this

PR #1367, opened 2018-08-14
use one event type for downloads

jackrobison:


lbrynet/analytics.py
```diff
@@ -11,9 +11,10 @@
 SERVER_STARTUP = 'Server Startup'
 SERVER_STARTUP_SUCCESS = 'Server Startup Success'
 SERVER_STARTUP_ERROR = 'Server Startup Error'
-DOWNLOAD_STARTED = 'Download Started'
-DOWNLOAD_ERRORED = 'Download Errored'
-DOWNLOAD_FINISHED = 'Download Finished'
+DOWNLOAD = 'Download'
+DOWNLOAD_STARTED = 'Started'
```
lyoshenka: DOWNLOAD_STARTED, DOWNLOAD_FINISHED, etc should not be in the same group as the rest of these constants. all the constants in this group are event types, and each has its own table in the db. the started/finished/errored constants are statuses for the DOWNLOAD event.

lyoshenka:
@jackrobison how do you feel about https://github.com/lbryio/lbry/pull/1385 instead of this?

jackrobison:
closing in favor of https://github.com/lbryio/lbry/pull/1385

PR #1368, opened 2018-08-14
author is not a field (or at least not required)

kauffj:


eukreign:
@kauffj can you provide some background on this? the command function itself still accepts an `author` argument, should this be dropped as well?

kauffj:
I actually thought `author` had been dropped when we added channels. IMO it should be removed from lbryschema and dropped entirely in favor of using channels. 

The bare minimum here is to:
1) Stop indicating it is required in docs
2) If mentioned in the docs at all, mention it is deprecated and one should use channels instead.

I'll leave it to you if you want to prioritize other changes like removing from lbryschema

jackrobison:
If I publish the pdf (to my channel or not) of _The Odyssey_, the `author` is `Homer`. I think we should keep the field - I'll update the docstring to explain the distinction between usages.

jackrobison:
@kauffj https://github.com/lbryio/lbry/pull/1424

PR #1381, opened 2018-08-22
wallet_send command working

hackrush01:


lbrynet/daemon/Daemon.py
```diff
@@ -2388,12 +2388,13 @@ def jsonrpc_send_amount_to_address(self, amount, address):
         elif not amount:
             raise NullFundsError()
 
+        amount = self.get_dewies_or_error("amount", amount)
```
eukreign: Shouldn't getting the amount happen *before* the amount checks above?

lbrynet/daemon/Daemon.py
```diff
@@ -2388,12 +2388,13 @@ def jsonrpc_send_amount_to_address(self, amount, address):
         elif not amount:
             raise NullFundsError()
```
eukreign: This is wrong, if amount is `None` then the `amount < 0` will throw an exception. First there need to be a check that amount is not `None` and then a comparison can be done.

```python
>>> None < 0
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

lbrynet/daemon/Daemon.py
```diff
@@ -2429,7 +2430,7 @@ def jsonrpc_wallet_send(self, amount, address=None, claim_id=None):
             raise Exception("Given both an address and a claim id")
         elif not address and not claim_id:
             raise Exception("Not given an address or a claim id")
-        if amount < 0:
+        if amount < 0.0:
             raise NegativeFundsError()
         elif not amount:
             raise NullFundsError()
```
eukreign: Same as earlier comment, this is doing less than comparison before a null comparison. Also, why the change from `0` to `0.0`? The amount should be in dewies which is an `int`.

lbrynet/wallet/manager.py
```diff
@@ -15,6 +15,12 @@
 log = logging.getLogger(__name__)
 
 
+class ReservedPoints:
+    def __init__(self, identifier, amount):
+        self.identifier = identifier
+        self.amount = amount
+
+
```
eukreign: Why did this get moved? Actually, it would be even better if it was deleted altogether, could you do the necessary refactoring to make that happen? I don't think it's used in many places.

lbrynet/wallet/transaction.py
```diff
@@ -72,6 +72,15 @@ class Transaction(BaseTransaction):
...
     output_class = Output
 
+    @classmethod
+    def pay(cls, amount: int, address: bytes, funding_accounts: List[Account], change_account: Account):
+        ledger = cls.ensure_all_have_same_ledger(funding_accounts, change_account)
+        hash_of_address = ledger.address_to_hash160(address)
+
+        output = Output.pay_pubkey_hash(amount, hash_of_address)
+
+        return cls.create([], [output], funding_accounts, change_account)
```
eukreign: Can you reduce the number of blank lines and inline `hash_of_address` to be consistent with how all of the other `Transaction` methods are implemented.

PR #1382, opened 2018-08-22
wallet_send_command working

eukreign:


lbrynet/daemon/Daemon.py
```diff
@@ -2389,12 +2389,13 @@ def jsonrpc_send_amount_to_address(self, amount, address):
         elif not amount:
             raise NullFundsError()
 
+        amount = self.get_dewies_or_error("amount", amount)
```
eukreign: Shouldn't getting the amount happen before the amount checks above?

lbrynet/daemon/Daemon.py
```diff
@@ -2389,12 +2389,13 @@ def jsonrpc_send_amount_to_address(self, amount, address):
         elif not amount:
             raise NullFundsError()
```
eukreign: This is wrong, if amount is `None` then the `amount < 0` will throw an exception. First there need to be a check that amount is not `None` and then a comparison can be done.

```python
>>> None < 0
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

lbrynet/daemon/Daemon.py
```diff
@@ -2430,13 +2431,7 @@ def jsonrpc_wallet_send(self, amount, address=None, claim_id=None):
...
-        try:
-            amount = Decimal(str(amount))
-        except InvalidOperation:
-            raise TypeError("Amount does not represent a valid decimal.")
-
-        if amount < 0:
+        if amount < 0.0:
             raise NegativeFundsError()
         elif not amount:
             raise NullFundsError()
```
eukreign: Same as earlier comment, this is doing less than comparison before a null comparison. Also, why the change from `0` to `0.0`? The amount should be in dewies which is an `int`.

lbrynet/wallet/manager.py
```diff
@@ -15,6 +15,12 @@
 log = logging.getLogger(__name__)
 
 
+class ReservedPoints:
+    def __init__(self, identifier, amount):
+        self.identifier = identifier
+        self.amount = amount
+
+
```
eukreign: Why did this get moved? Actually, it would be even better if it was deleted altogether, could you do the necessary refactoring to make that happen? I don't think it's used in many places.

lbrynet/wallet/transaction.py
```diff
@@ -72,6 +72,15 @@ class Transaction(BaseTransaction):
...
 
+    @classmethod
+    def pay(cls, amount: int, address: bytes, funding_accounts: List[Account], change_account: Account):
+        ledger = cls.ensure_all_have_same_ledger(funding_accounts, change_account)
+        hash_of_address = ledger.address_to_hash160(address)
+
+        output = Output.pay_pubkey_hash(amount, hash_of_address)
+
+        return cls.create([], [output], funding_accounts, change_account)
+
```
eukreign: Can you reduce the number of blank lines and inline `hash_of_address` to be consistent with how all of the other `Transaction` methods are implemented.

CHANGELOG.md
```diff
@@ -29,7 +29,7 @@ at anytime.
   *
 
 ### Removed
-  *
+  * deprecated `send_amount_to_address` command from daemon
```
eukreign: removed commad `send_amount_to_address` which was previously marked as deprecated

lbrynet/daemon/Daemon.py
```diff
@@ -2427,19 +2405,26 @@ def jsonrpc_wallet_send(self, amount, address=None, claim_id=None):
...
+        amount = self.get_dewies_or_error("amount", amount)
         if address and claim_id:
             raise Exception("Given both an address and a claim id")
         elif not address and not claim_id:
             raise Exception("Not given an address or a claim id")
-        if amount < 0.0:
+
+        if not amount:
+            raise NullFundsError
+        elif amount < 0:
```
eukreign: Move the amount checks to come right after the amoutn variable is set.

CHANGELOG.md
hackrush01: :+1:

lbrynet/wallet/transaction.py
eukreign: An address is already a hash of the public key. It's not useful to have this variable since nothing happens with it and it's just doing a common conversion.

lbrynet/daemon/Daemon.py
hackrush01: :+1:

lbrynet/wallet/transaction.py
hackrush01: image

lbrynet/wallet/manager.py
eukreign: Yes, please remove it and make the necessary changes.

lbrynet/wallet/transaction.py
eukreign: We use `hexlify()` and `unhexlify()` all over the place and at least half of the usage is inline (doing the conversion just as the variable is passed to some function). This hex address to hash160 conversions is similar to `hexlify/unhexlify` and is pretty boring and standard operation. It doesn't need to be pulled out and highlighted to bring attention to it, if anything it distracts from the main flow of the function.

hackrush01: `hexlify`/`unhexlify` is a built-in function which everyone knows about. But not the conversion for our hash. So this can be separate variable to make things clearer.

lbrynet/wallet/manager.py
hackrush01: Ok, will/might take some time. If I don't push a commit for this today, then don't treat this as a blocker. But rest assured this is on top of my list.

PR #1390, opened 2018-08-24
Fix some commands

shyba:
- block_show using block hash
- block_show using height
- claim_show from claim_id
- claim_show from outpoint

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -487,7 +487,7 @@ class Daemon(AuthJSONRPCServer):
         Resolve a name and return the estimated stream cost
         """
 
-        resolved = (yield self.wallet.resolve(uri))
+        resolved = yield self.wallet.resolve(uri)
         if resolved:
             claim_response = resolved[uri]
         else:
@@ -2514,15 +2514,12 @@ class Daemon(AuthJSONRPCServer):
         """
 
         if blockhash is not None:
-            d = self.wallet.get_block(blockhash)
+            return self.wallet.get_block(blockhash)
         elif height is not None:
-            d = self.wallet.get_block_info(height)
+            return self.wallet.get_block_info(height)
         else:
             # TODO: return a useful error message
             return server.failure
-
-        d.addCallback(lambda r: self._render_response(r))
-        return d
 
     @requires(WALLET_COMPONENT, DHT_COMPONENT, BLOB_COMPONENT, RATE_LIMITER_COMPONENT, PAYMENT_RATE_COMPONENT,
               conditions=[WALLET_IS_UNLOCKED])
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -2517,7 +2517,6 @@ class Daemon(AuthJSONRPCServer):
             d = self.wallet.get_block(blockhash)
         elif height is not None:
             d = self.wallet.get_block_info(height)
-            d.addCallback(lambda b: self.wallet.get_block(b))
         else:
             # TODO: return a useful error message
             return server.failure
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -487,7 +487,7 @@ class Daemon(AuthJSONRPCServer):
         Resolve a name and return the estimated stream cost
         """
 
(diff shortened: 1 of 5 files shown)


lbrynet/daemon/Daemon.py
```diff
@@ -487,7 +487,7 @@ def get_est_cost_from_uri(self, uri):
         Resolve a name and return the estimated stream cost
         """
 
-        resolved = (yield self.wallet.resolve(uri))[uri]
+        resolved = (yield self.wallet.resolve(uri))
```
eukreign: Don't need the extra parentheses anymore.

lbrynet/daemon/Daemon.py
```diff
@@ -2517,7 +2517,6 @@ def jsonrpc_block_show(self, blockhash=None, height=None):
             d = self.wallet.get_block(blockhash)
         elif height is not None:
             d = self.wallet.get_block_info(height)
-            d.addCallback(lambda b: self.wallet.get_block(b))
         else:
             # TODO: return a useful error message
             return server.failure
```
eukreign: You could refactor this whole function to just `return` `get_block()` and `get_block_info()`, `self._render_response()` doesn't do anything.

lbrynet/wallet/ledger.py
```diff
@@ -49,6 +49,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fee_per_name_char = self.config.get('fee_per_name_char', self.default_fee_per_name_char)
 
+    @property
+    def resolver(self):
+        return Resolver(self.headers.claim_trie_root, self.headers.height, self.transaction_class,
+                        hash160_to_address=lambda x: self.hash160_to_address(x), network=self.network)
```
eukreign: Why not just `ash160_to_address=self.hash160_to_address`?

lbrynet/wallet/ledger.py
```diff
@@ -57,9 +62,21 @@ def resolve(self, page, page_size, *uris):
...
-        resolver = Resolver(self.headers.claim_trie_root, self.headers.height, self.transaction_class,
-                            hash160_to_address=lambda x: self.hash160_to_address(x), network=self.network)
-        defer.returnValue((yield resolver._handle_resolutions(resolutions, uris, page, page_size)))
+        defer.returnValue((yield self.resolver._handle_resolutions(resolutions, uris, page, page_size)))
+
+    def get_claim_by_claim_id(self, claim_id):
+        d = self.network.get_claims_by_ids(claim_id)
+        d.addCallback(lambda result: result.pop(claim_id) if result else {})
+        d.addCallback(self.resolver.get_cert_and_validate_result)
+        return d
```
eukreign: Can you refactor this to use `@defer.inlineCallbacks`?

lbrynet/wallet/resolve.py
```diff
@@ -145,6 +145,17 @@ def _handle_resolve_uri_response(self, uri, resolution, page=0, page_size=10, ra
 
         defer.returnValue(result)
 
+    @defer.inlineCallbacks
+    def get_cert_and_validate_result(self, claim_result):
```
eukreign: Should be `get_certificate_and_validate_result()`.

lbrynet/wallet/ledger.py
```diff
@@ -57,9 +62,20 @@ def resolve(self, page, page_size, *uris):
             except URIParseError as err:
                 defer.returnValue({'error': err.message})
         resolutions = yield self.network.get_values_for_uris(self.headers.hash().decode(), *uris)
-        resolver = Resolver(self.headers.claim_trie_root, self.headers.height, self.transaction_class,
-                            hash160_to_address=lambda x: self.hash160_to_address(x), network=self.network)
-        defer.returnValue((yield resolver._handle_resolutions(resolutions, uris, page, page_size)))
+        defer.returnValue((yield self.resolver._handle_resolutions(resolutions, uris, page, page_size)))
```
eukreign: Does `return yield self.resolver.....` work? In twisted on py3 `defer.returnValue` is no longer necessary. I'm wondering if it could be dropped here.

shyba: Good catch. `return yield` doesn't from my last experiments, but `return (yield d)` did. Gonna try that.

PR #1395, opened 2018-08-28
Integration test for wallet_send

hackrush01:


tests/integration/wallet/test_commands.py
```diff
@@ -10,9 +10,13 @@
 import lbryschema
 lbryschema.BLOCKCHAIN_NAME = 'lbrycrd_regtest'
 
+from torba.constants import COIN
+
+
```
eukreign: I don't think 2 blank lines is necessary between import statements. One is enough if you are trying to distinguish blocks of imports.

tests/integration/wallet/test_commands.py
```diff
@@ -298,3 +306,29 @@ def test_no_this_is_not_a_test_its_an_adventure(self):
...
+        # After abandoning he just waits for his LBCs to be returned to his account
+        yield  self.d_generate(5)
+        result = yield self.daemon.jsonrpc_wallet_balance()
+        self.assertEqual(round(result, 2), 8.97)
+
+        # Amidst all this Chris45 receives a call from his friend Ramsey54
+        # who says that it is of utmost urgency that Chris45 transfer him
+        # 1 LBC to which Chris45 ready obliges
+        ramsey_account = Account.generate(self.ledger, self.wallet, "Ramsey54")
+        yield self.ledger.update_account(ramsey_account)
```
eukreign: Should be using `jsonrpc_account_create`.

tests/integration/wallet/test_commands.py
```diff
@@ -79,9 +83,13 @@ def shutdown(self):
     def send_claim_action(self, action):
         pass
 
+    def send_credits_sent(self):
+        pass
+
 
 class CommandTestCase(IntegrationTestCase):
 
+    timeout = 300
```
eukreign: Why does this need to change?

hackrush01: Earlier I thought the test was failing due to timeout(the tests take 90 or so secs now btw) so I increased it and kept it that way. I can remove this if needed. Also this helps in debugging. Though I have no strong opinions.

tests/integration/wallet/test_commands.py
eukreign: Will be soon.

tests/integration/wallet/test_commands.py
eukreign: How long does it currently take to run the test? If we are close to 120 already then updating is fine, if we're still far from 120 then let's just wait (and maybe even make the test run faster).

tests/integration/wallet/test_commands.py
eukreign: Great!

Issue #1278, opened 2018-06-28
Revisit wallet_prefill_addresses command


tzarebczan:
## The Issue
Do we really want this function creating multiple wallet addresses, or should it be sending multiple inputs to the same address? This is effectively the same result, but without the address bloat, right? 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

eukreign:
Depends on the use case for this command. What is the use case for this command?

From looking at the code the command does what I'd expect it to do.

I can see this type of command being useful in a high throughput services where you want to be able to handle a lot of new TXs without running out of TXOs.

If you have 1 TXO valued at 10 credits. That means you can only do 1 TX at a given point in time spending up-to the 10 credits. Using this command you can ask to have the 10 credits spread across 5 new addresses each with 2 credits. Now you can create 5 TX each spending up-to 2 credits simultaneously. This is useful for spee.ch to increase throughput.

My preference would be to drop this command altogether and just make it possible to do it with two new / updated commands: 1) a command to fetch `X` number of unused addresses and 2) a command that takes an amount and a list of addresses to pay that amount to. This would allow a user to replicate what `wallet_prefill_addresses` does but in a more configurable way (for example, instead of just unused addresses, a user may want to use other heuristics to figure out which addresses they want to fund... or maybe they just want to pay the same amount to a bunch of people, etc).

tzarebczan:
@nikooo777 we should confirm that the new `fund` command works for your YT sync scenario when we test the wallet.

tzarebczan:
Tested on new wallet (on spee.ch), doesn't look right:
lbrynet-cli wallet_prefill_addresses 50 .04

and this was the result (I was expecting 50 UTXOs with 0.04):
https://explorer.lbry.io/tx/cb08b32de15b003e7930ced91e7226a02f6e8fab53ddabd3709947bee6b950fe

PR #1421, opened 2018-08-29
Removed unnecessary yield and inlineCallbacks, rectified command behaviour

hackrush01:


hackrush01:
Also fixes #1278 
According to this: https://github.com/lbryio/lbry/issues/1278#issuecomment-416981496

lbrynet/daemon/Daemon.py
```diff
@@ -3263,10 +3263,11 @@ def jsonrpc_fund(self, to_account, from_account, amount=0,
             raise ValueError("--outputs must be an integer.")
         if everything and outputs > 1:
             raise ValueError("Using --everything along with --outputs is not supported.")
-        return from_account.fund(
+        tx = yield from_account.fund(
             to_account=to_account, amount=amount, everything=everything,
             outputs=outputs, broadcast=broadcast
         )
+        return tx
```
eukreign: This doesn't look correct to me. Was something fixed by this change?

PR #1422, opened 2018-08-29
[API] use ssl for all api requests (WIP)

jackrobison:


lbrynet/daemon/auth/auth.py
```diff
@@ -24,29 +24,11 @@ def requestAvatar(self, avatarId, mind, *interfaces):
...
-
-    @classmethod
-    def load(cls, password_dict):
-        passwords = {key: password_dict[key].secret for key in password_dict}
-        log.info("Loaded %i api key(s)", len(passwords))
-        return cls(passwords)
+    def __init__(self, keyring: Keyring):
+        self.api_key = keyring.api_key
 
     def requestAvatarId(self, creds):
```
eukreign: Can this method name also be renamed to follow PEP standard? I realize it's not part of the PR but if it's easy.

lbrynet/daemon/auth/auth.py
```diff
@@ -24,29 +24,11 @@ def requestAvatar(self, avatarId, mind, *interfaces):
...
-
-        if creds.username in password_dict_bytes:
-            pw = password_dict_bytes.get(creds.username)
-            pw_match = creds.checkPassword(pw)
-            if pw_match:
-                return defer.succeed(creds.username)
+        if creds.checkPassword(self.api_key.secret.encode()) and creds.username == self.api_key.name.encode():
+            return defer.succeed(creds.username)
         log.warning('Incorrect username or password')
         return defer.fail(cred_error.UnauthorizedLogin('Incorrect username or password'))
```
eukreign: Why does this return Deferreds if everything is sync?

lbrynet/daemon/auth/client.py
```diff
@@ -100,28 +102,26 @@ def f(*args, **kwargs):
...
 
-        pw_path = os.path.join(conf.settings['data_dir'], ".api_keys")
-        keys = load_api_keys(pw_path)
-        api_key = keys.get(api_key_name, False)
+        api_key = keyring.api_key
 
-        login_url = "http://{}:{}@{}:{}".format(api_key_name, api_key.secret, conf.settings['api_host'],
-                                                conf.settings['api_port'])
+        login_url = "http{}://{}:{}@{}:{}".format("" if not conf.settings['use_https'] else "s", api_key_name,
+                                                  api_key.secret, conf.settings['api_host'], conf.settings['api_port'])
```
eukreign: This URL building is duplicated 3 times in the PR, can it be extracted into some kind of utility function or method?

lbrynet/daemon/auth/keyring.py
```diff
@@ -0,0 +1,134 @@
...
+        return sha(x)
+
+
+class APIKey:
+    def __init__(self, secret: str, name: str):
+        self.secret = secret
+        self.name = name
+
+    @classmethod
+    def new(cls, seed=None, name=None):
```
eukreign: Can you use `create` instead of `new`.

lbrynet/daemon/auth/keyring.py
```diff
@@ -0,0 +1,134 @@
...
+        with open(api_ssl_cert_path, 'rb') as f:
+            public_cert = f.read().decode()
+        return cls(api_key, public_cert)
+
+    @classmethod
+    def generate_and_save(cls):
+        country: str = "US"
+        organization: str = "LBRY"
+        common_name: str = "LBRY API"
+        expiration: int = 365
```
eukreign: Local variables don't need to be explicitly typed, their types can be inferred from the value being set. PyCharm is pretty good about inferring types of local values like this.

lbrynet/daemon/auth/auth.py
jackrobison: no, it's required to be a `ICredentialsChecker`

lbrynet/daemon/auth/keyring.py
jackrobison: done

lbrynet/daemon/auth/keyring.py
jackrobison: done

PR #1425, opened 2018-08-31
[bug] remove get_block_info as it doesnt make sense, unify to get_block

shyba:


lbrynet/wallet/manager.py
```diff
@@ -291,13 +291,13 @@ def save(self):
         for wallet in self.wallets:
             wallet.save()
 
-    def get_block(self, block_hash):
+    def get_block(self, block_hash=None, height=None):
+        if height:
+            block_hash = self.ledger.headers.hash(height).decode()
```
eukreign: What if height=0? Actually, there is a bug on torba related to this as well.

Issue #1342, opened 2018-08-06
Ensure first blob is written to output file first (MP4 streaming issues)


tzarebczan:
## The Issue
Per discussion with @jackrobison, it's possible that the 2nd (or others) blob is written to the output file before the first one, relevant code: https://github.com/lbryio/lbry/blob/e50ac69358bda507a7acd298ebe767cf7cba438e/lbrynet/daemon/Downloader.py#L64

This causes intermediate issues when streaming MP4s which have metadata required to start the stream in the first few bytes. If the 2nd blob is written first, it will get stuck on "waiting for metadata" until you leave the claim and come back (https://github.com/lbryio/lbry-desktop/issues/754). 

Is there any downside to waiting for the first blob to download before writing the file? 

Notes from chat:
```
jack
what it does now: 
1. `get` request comes in
2. resolve the uri, throw error if it doesnt resolve to a stream
3. download sd blob, return sd timeout if fails
4. start downloading stream from many peers
5. as soon as a peer starts to send the first blob file to us, return the result from `get`
6. blob finishes (or maybe even fails before finishing) after we've returned from `get`. blob file on disk is removed on fail or saved and decrypted and added to the output file in the downloads folder

jack 
just checked, there are edge cases the way it is that could break this
jack
it's _supposed_ to block on the first data blob being completed
jack
it actually blocks on _any_ data blob being completed
jack
if it blocks on the first one that should be very close to the time it's decrypted and written to the output file
```

### Actual behaviour
Tell us what happens instead

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

skhameneh:
Repeat of partial notes from lbrio/lbry-desktop :
>
>...
[...]

PR #1432, opened 2018-09-10
[API] block get on bytes being written to the output file

jackrobison:


tzarebczan:
Can this be extended to support https://github.com/lbryio/lbry/issues/1342 ? Is there any reason we wouldn't want to wait until the first blob is downloaded before the output file is written?

tzarebczan:
Also, see https://github.com/lbryio/lbry/issues/854

jackrobison:
@eukreign all that's left on this is updating `tests.unit.test_daemon.test_Downloader`

eukreign:
what's the status of this? can it be merged or closed or what action can be taken on this?

tzarebczan:
Tested this for Jack briefly and it works as expected.

PR #1436, opened 2018-09-12
Implemented removal of float usage. Floats converted to int internally.

tiger5226:


tiger5226:
@eukreign https://travis-ci.org/lbryio/lbry/jobs/428255582#L760 looking here it appears the integration tests provide an int instead of a float. Should we adjust the regex to allow for 1 or 1.0? Or should we update the integration test? I presume updating the integration test could break functionality if we ever rely on that scenario. Figured I would ask you before changing anything.

tiger5226:
@eukreign Just to confirm. All of the integration tests use numbers or floats, and our method only accepts strings now. `1` and `1.0` do not work, but `"1.0"` does work.

eukreign:
Yes, do it per @lyoshenka requirements. Since we're breaking backwards compatibility anyways it makes sense to do it right and to have one and only one way to do this, to avoid any ambiguities in the future.

PR #1438, opened 2018-09-14
adapt content payment calls to new wallet

shyba:


lbrynet/daemon/Downloader.py
```diff
@@ -110,7 +112,7 @@ def check_fee_and_convert(self, fee):
             'amount': converted_fee_amount,
             'address': fee.address
         }
-        return Fee(converted_fee)
+        defer.returnValue((yield Fee(converted_fee)))
```
eukreign: I think the original code is correct here: `return Fee(converted_fee)`

PR #1448, opened 2018-09-18
Fix account_encrypt, account_decrypt, and account_unlock

jackrobison:
-add `account_lock`

requires https://github.com/lbryio/torba/pull/10

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -175,7 +175,7 @@ class DHTHasContacts(RequiredCondition):
         return len(component.contacts) > 0
 
 
-class WalletIsLocked(RequiredCondition):
+class WalletIsUnlocked(RequiredCondition):
     name = WALLET_IS_UNLOCKED
     component = WALLET_COMPONENT
     message = "your wallet is locked"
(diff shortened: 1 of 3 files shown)


lbrynet/daemon/auth/factory.py
```diff
@@ -22,7 +36,11 @@ def getChild(self, name, request):
         request.setHeader('expires', '0')
         return self if name == '' else resource.Resource.getChild(self, name, request)
 
-    def getServerFactory(self):
+    def getServerFactory(self, factory: AuthHTTPJSONRPCFactory=None) -> server.Site:
+        if factory and type(factory) in (AuthHTTPJSONRPCFactory, AuthHTTPSJSONRPCFactory):
+            pass
+        else:
+            factory = AuthHTTPSJSONRPCFactory
```
eukreign: Is `factory` a class or an instance of a class?

lbrynet/daemon/Daemon.py
```diff
@@ -1361,64 +1361,75 @@ def jsonrpc_account_set(
...
+            account_lock [<account_id> | --account_id=<account_id>]
+
+        Options:
+            --account_id=<account_id>        : (str) id for the account to lock
+
+        Returns:
+            (bool) true if account is locked, otherwise false
+        """
+
+        return self.wallet_manager.lock_account(account_id)
```
eukreign: Convert `account_id` to an actual account using `self.get_account_or_default(account_id)`.

lbrynet/daemon/Daemon.py
```diff
@@ -1361,64 +1361,75 @@ def jsonrpc_account_set(
...
         Options:
-            None
+            --account_id=<account_id>  : (str) id for the account to decrypt
 
         Returns:
             (bool) true if wallet is decrypted, otherwise false
         """
-        return self.wallet_manager.decrypt_wallet()
+
+        return self.wallet_manager.decrypt_account(account_id)
```
eukreign: Convert `account_id` to an actual account using `self.get_account_or_default(account_id)`.

lbrynet/daemon/Daemon.py
```diff
@@ -1361,64 +1361,75 @@ def jsonrpc_account_set(
...
-        # the wallet_unlocked_d deferred ¯\_(ツ)_/¯
-        if not self.wallet_manager.check_locked():
-            d = self.wallet_manager.wallet_unlocked_d
-            d.callback(password)
-            result = yield d
-        else:
-            result = True
-        response = yield self._render_response(result)
-        defer.returnValue(response)
+        return self.wallet_manager.unlock_wallet(password.encode(), account_id)
```
eukreign: Convert `account_id` to an actual account using `self.get_account_or_default(account_id)`.

lbrynet/daemon/Daemon.py
```diff
@@ -1361,64 +1361,75 @@ def jsonrpc_account_set(
...
 
         Returns:
             (bool) true if wallet is decrypted, otherwise false
         """
-        return self.wallet_manager.encrypt_wallet(new_password)
+
+        return self.wallet_manager.encrypt_account(
+            new_password.encode(),
+            account_id
+        )
```
eukreign: Convert `account_id` to an actual account using `self.get_account_or_default(account_id)`.

lbrynet/wallet/manager.py
```diff
@@ -54,19 +55,65 @@ def network(self):
 
     @property
     def use_encryption(self):
-        # TODO: implement this
-        return False
+        return self.default_account.serialize_encrypted
 
     @property
     def is_first_run(self):
         return True
```
eukreign: Do we still need `is_first_run`? I know it's not part of this PR, but just wondering if we can drop it.

lbrynet/wallet/manager.py
```diff
@@ -54,19 +55,65 @@ def network(self):
...
 
     @property
     def is_wallet_unlocked(self):
-        return True
+        return not self.default_account.encrypted
 
     def check_locked(self):
-        return defer.succeed(False)
+        return defer.succeed(self.default_account.encrypted)
```
eukreign: Can this be changed to a sync method instead of returning Deferred?

lbrynet/wallet/manager.py
```diff
@@ -54,19 +55,65 @@ def network(self):
...
+            return self.default_account
+        for account in self.default_wallet.accounts:
+            if account.id == account_id:
+                if lbc_only and not isinstance(account, LBCAccount):
+                    raise ValueError(
+                        "Found '{}', but it's an {} ledger account. A LBC Account is needed."
+                        .format(account_id, account.ledger.symbol)
+                    )
+                return account
+        raise ValueError("Couldn't find account: {}.".format(account_id))
```
eukreign: This shouldn't be here, it's a copy of what's already in `Daemon.py`.

lbrynet/wallet/manager.py
```diff
@@ -54,19 +55,65 @@ def network(self):
...
+        assert account.encrypted, "account is not locked"
+        account.decrypt(password)
+        return defer.succeed(not account.encrypted)
+
+    def lock_account(self, account_id):
+        account = self.get_account_or_error(account_id, lbc_only=False)
+        assert account.password is not None, "account is already locked"
+        assert not account.encrypted and account.serialize_encrypted, "account is not encrypted"
+        account.encrypt(account.password)
+        return defer.succeed(account.encrypted)
```
eukreign: The 4 methods above should all accept an account instance instead of an `account_id`. Also, can you make them return plain values instead of Deferreds, none of them are actually async.

tests/unit/wallet/test_encryption.py
```diff
@@ -0,0 +1,108 @@
...
+
+        self.assertEqual(account.seed, self.unencrypted_account['seed'])
+        self.assertEqual(account.private_key.extended_key_string(), self.unencrypted_account['private_key'])
+
+        self.assertEqual(account.to_dict()['seed'], self.encrypted_account['seed'])
+        self.assertEqual(account.to_dict()['private_key'], self.encrypted_account['private_key'])
+
+        account.serialize_encrypted = False
+        self.assertEqual(account.to_dict()['seed'], self.unencrypted_account['seed'])
+        self.assertEqual(account.to_dict()['private_key'], self.unencrypted_account['private_key'])
```
eukreign: Is this duplicate of what you added to torba? I would drop it from lbrynet unless there are lbrynet specific things it needs to ensure. I'd rather rely on integration tests that use the Daemon encrypt/decrypt API for sanity testing and leave the unit tests in torba.

Issue #1302, opened 2018-07-13
--conf doesn't error when file doesn't exist


eukreign:
When passing a custom settings file to `lbrynet-daemon` it should fail if the settings file doesn't exist, instead lbrynet-daemon starts anyways without any messages telling you that your config file isn't being used. On the other hand if the file (whether it exists or doesn't) does not end in `.yaml` then `lbrynet-daemon` does crash fast:

In the below examples the files `monkey` or `monkey.yml` don't exist.

No `.yml` fails:
```
[lbry]$ lbrynet-daemon --conf monkey
Traceback (most recent call last):
  File "/home/lex/.virtualenvs/lbry/bin/lbrynet-daemon", line 11, in <module>
    load_entry_point('lbrynet', 'console_scripts', 'lbrynet-daemon')()
  File "/home/lex/projects/lbry/lbrynet/daemon/DaemonControl.py", line 64, in start
    conf.settings.load_conf_file_settings()
  File "/home/lex/projects/lbry/lbrynet/conf.py", line 547, in load_conf_file_settings
    assert decoder is not False, 'Unknown settings format %s' % ext
AssertionError: Unknown settings format 
```

With `.yml` but still non-existent file works:
```
[lbry]$ lbrynet-daemon --conf monkey.yml
2018-07-13 11:10:33,184 INFO     lbrynet.daemon.DaemonControl:84: Starting lbrynet-daemon from command line
2018-07-13 11:10:39,249 INFO     lbrynet.daemon.DaemonServer:65: Using non-authenticated API
2018-07-13 11:10:39,251 INFO     lbrynet.daemon.DaemonServer:42: lbrynet API listening on TCP localhost:5279
2018-07-13 11:10:39,252 INFO     lbrynet.daemon.Daemon:208: Starting lbrynet-daemon
```

Expected behavior:
1. If it's easy, allow settings files with any extension (including no extension).
1. If you pass `--conf` to `lbrynet-daemon` and the settings file referenced doesn't exist, `lbrynet-daemon` should fail fast with useful error about file not being found.
1. Bonus: If user did something like: `lbrynet-daemon --conf ../../blahblah.yml` and this file doesn't exist (perhaps because they were wrong about their relative location to file) then in the error message expand the relative path to absolute path which will make it easier for the user to see their mistake (all paths can be expanded, no need to check if it's relative, there are Python built-in libs to expand paths to absolute).

whatbatman:
[...]

PR #1455, opened 2018-09-19
--conf option improvements

ttkopec:
Fix for [#1302](https://github.com/lbryio/lbry/issues/1302)
Changes:
- if provided config file does not exist, throw descriptive error with absolute path of a file
- throw ValueError, not AssertionError if provided config file has unsupported extension
- simple test for loading file

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -492,7 +492,10 @@ class Config:
         rev = self._convert_conf_file_lists_reverse(self._data[TYPE_PERSISTED])
         ext = os.path.splitext(path)[1]
         encoder = settings_encoders.get(ext, False)
-        assert encoder is not False, 'Unknown settings format %s' % ext
+
+        if not encoder:
+            raise ValueError('Unknown settings format %s' % ext)
+
         with open(path, 'w') as settings_file:
             settings_file.write(encoder(rev))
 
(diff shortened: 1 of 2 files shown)


lbrynet/conf.py
```diff
@@ -523,21 +526,30 @@ def load_conf_file_settings(self):
...
             with open(path, 'r') as settings_file:
                 data = settings_file.read()
             decoded = self._fix_old_conf_file_settings(decoder(data))
             log.info('Loaded settings file: %s', path)
             self._validate_settings(decoded)
             self._data[TYPE_PERSISTED].update(self._convert_conf_file_lists(decoded))
-        except (IOError, OSError) as err:
-            log.info('%s: Failed to update settings from %s', err, path)
+        except FileNotFoundError as err:
+            raise FileNotFoundError('Error while reading settings file \'{}\': {}'.format(os.path.abspath(path), err.strerror))
```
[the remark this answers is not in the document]
ttkopec: I wanted to augment error message with absolute path of a file. If it is not needed, then I can do as you mentioned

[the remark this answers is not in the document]
eukreign: For now just propagate the original exception.

lbrynet/conf.py
```diff
@@ -583,6 +584,9 @@ def get_db_revision_filename(self):
         return os.path.join(self.ensure_data_dir(), self['DB_REVISION_FILE_NAME'])
 
     def get_conf_filename(self):
+        """
+        Return local configuration filename. If file does not exist, return None.
+        """
```
eukreign: Instead of the comment you added I'd like to rename the method name to something like: `get_valid_settings_filename`. This will explicitly document everywhere that the settings filename returned will be validated (at least that it exists). This is especially important since typically a method named similar to `get_conf_filename` would appear to be without performance penalty but in this case you are actually accessing the file system and checking that the file exists... every time it's called; the method name should reflect this or standout somehow.

lbrynet/conf.py
```diff
@@ -591,7 +595,7 @@ def get_conf_filename(self):
         elif os.path.isfile(json_path):
             return json_path
         else:
-            return yml_path
+            return None
```
eukreign: This is redundant here, you can remove it:
```
else:
  return None
```

ttkopec:
@tzarebczan Yes, sorry, I've been recently quite busy but will try to finish this PR as fast as possible

lbrynet/conf.py
ttkopec: Ok, I got it :) Then I'll remove it as you requested

PR #1456, opened 2018-09-19
WIP: fixes for various query commands

eukreign:


lbrynet/daemon/Daemon.py
```diff
@@ -1281,16 +1281,16 @@ def jsonrpc_account_remove(self, account_id):
         Remove an existing account.
 
         Usage:
-            account (<account_id> | --account_id=<account_id>)
+            account (<account_id> | --account=<account_id>)
```
hackrush01: @eukreign Are we not keeping this as `--account_id`

PR #1461, opened 2018-09-19
New wallet claim abandon command

hackrush01:
Returns only tx in the output

### DIFF
--- a/lbrynet/daemon/Daemon.py
+++ b/lbrynet/daemon/Daemon.py
@@ -2326,18 +2326,20 @@ class Daemon(AuthJSONRPCServer):
 
     @requires(WALLET_COMPONENT, conditions=[WALLET_IS_UNLOCKED])
     @defer.inlineCallbacks
-    def jsonrpc_claim_abandon(self, claim_id=None, txid=None, nout=None):
+    def jsonrpc_claim_abandon(self, claim_id=None, txid=None, nout=None, account_id=None):
         """
         Abandon a name and reclaim credits from the claim
 
         Usage:
             claim_abandon [<claim_id> | --claim_id=<claim_id>]
                           [<txid> | --txid=<txid>] [<nout> | --nout=<nout>]
+                          [--account_id=<account_id>]
 
         Options:
             --claim_id=<claim_id> : (str) claim_id of the claim to abandon
             --txid=<txid> : (str) txid of the claim to abandon
             --nout=<nout> : (int) nout of the claim to abandon
+            --account_id=<account_id> : (str) id of the account to use
 
         Returns:
             (dict) Dictionary containing result of the claim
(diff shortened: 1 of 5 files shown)


lbrynet/wallet/manager.py
```diff
@@ -264,7 +264,12 @@ def tip_claim(self, amount, claim_id, account):
     @defer.inlineCallbacks
     def abandon_claim(self, claim_id, txid, nout):
         account = self.default_account
```
eukreign: Pass in the `account` as an argument. We need to start moving towards having account argument on all daemon commands.

lbrynet/wallet/manager.py
```diff
@@ -264,7 +264,12 @@ def tip_claim(self, amount, claim_id, account):
...
-        claim = yield account.get_claim(claim_id)
+        if claim_id is not None:
+            claim = yield account.get_claim(claim_id=claim_id)
+        elif txid is not None and nout is not None:
+            claim = yield account.get_claim(txid=txid, nout=nout)
+        else:
+            raise Exception('Must specify claim_id, or txid and nout')
```
eukreign: No check for claim being null, should raise an error that claim was not found.

Also, you can reduce the if statements to just: `account.get_claim(claim_id=claim_id, txid=txid, nout=nout)`. Would have the same result but much less code.

lbrynet/daemon/Daemon.py
```diff
@@ -2326,39 +2326,44 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
...
     @defer.inlineCallbacks
-    def jsonrpc_claim_abandon(self, claim_id=None, txid=None, nout=None):
+    def jsonrpc_claim_abandon(self, claim_id=None, txid=None, nout=None, account_id=None):
         """
         Abandon a name and reclaim credits from the claim
 
         Usage:
             claim_abandon [<claim_id> | --claim_id=<claim_id>]
                           [<txid> | --txid=<txid>] [<nout> | --nout=<nout>]
+                          [--account_id=<account_id>]
```
eukreign: I would like to standardize on using `--account=` across all commands. I'll update the other commands today. Function argument itself should remain `<account_id>`.

lbrynet/daemon/Daemon.py
```diff
@@ -2326,39 +2326,44 @@ def jsonrpc_publish(self, name, bid, metadata=None, file_path=None, fee=None, ti
...
 
         Options:
             --claim_id=<claim_id> : (str) claim_id of the claim to abandon
             --txid=<txid> : (str) txid of the claim to abandon
             --nout=<nout> : (int) nout of the claim to abandon
+            --account_id=<account_id> : (str) id of the account to use
```
eukreign: I would like to standardize on using `--account=` across all commands. I'll update the other commands today. Function argument itself should remain `<account_id>`.

lbrynet/daemon/Daemon.py
hackrush01: Ok, will push that along with the integration test

lbrynet/daemon/Daemon.py
hackrush01: Same as above

PR #1466, opened 2018-09-21
send startup event right after setup

### LINKED ISSUE #1420: Startup analytics missing since component refactor (0.21.x) release
## The Issue
Looks like we lost startup analytics based on http://traction.lbry.io/question/19 when we refactored components. @shyba took a quick look and we are storing successes but not startups/failures. 

See: 
select * from segment.server_startup order by received_at desc limit 20;
select * from segment.server_startup_success order by received_at desc limit 20;
select * from segment.server_startup_error order by received_at desc limit 20;

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

shyba:
fixes #1420


eukreign:
The integration tests fail, need to add a mock analytics method for this PR.

eukreign:
@shyba can you fix the build on this by adding the missing method on mock analytics in integration tests

PR #1469, opened 2018-09-23
Fixes Issues within naming conventions

osilkin98:
There are a couple of names that use angle brackets which cause an issue when generating these functions from scratch. There was also a variable type "Decimal" which should be changed to "float" as all the others are.

### DIFF
--- a/docs/api.json
+++ b/docs/api.json
@@ -367,7 +367,7 @@
             {
                 "is_required": false, 
                 "type": "int", 
-                "name": "amount<amount>", 
+                "name": "amount", 
                 "description": "Amount of credits to claim name for, defaults to the current amount on the claim"
             }
         ], 
@@ -478,7 +478,7 @@
             {
                 "is_required": false, 
                 "type": "str", 
-                "name": "file_name<file_name>", 
+                "name": "file_name", 
                 "description": "delete by file name in downloads folder"
             }, 
             {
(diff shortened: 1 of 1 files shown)


docs/api.json
```diff
@@ -367,7 +367,7 @@
             {
                 "is_required": false, 
                 "type": "int", 
-                "name": "amount<amount>", 
+                "name": "amount",
```
osilkin98: The angle brackets need to be removed so my API Code generation can work properly

docs/api.json
```diff
@@ -478,7 +478,7 @@
             {
                 "is_required": false, 
                 "type": "str", 
-                "name": "file_name<file_name>", 
+                "name": "file_name",
```
osilkin98: The angle brackets need to be removed so my API Code generation can work properly

docs/api.json
```diff
@@ -1230,7 +1230,7 @@
         "arguments": [
             {
                 "is_required": true, 
-                "type": "Decimal", 
+                "type": "float",
```
osilkin98: This needs to change to float as it is just inconsistent

eukreign:
The `api.json` file shouldn't be hand edited since it's generated from `Daemon.py`. I've fixed the issues you found in `Daemon.py` and regenerated `api.json` in this PR: https://github.com/lbryio/lbry/pull/1473

Therefore I'm closing this PR. If you find new typos in the generated `api.json` please feel free to file new issue. Thank you for your contribution.

PR #1480, opened 2018-09-25
Deprecated wallet_public_key and claim_renew commands

hackrush01:


lbrynet/cli.py
```diff
@@ -145,7 +145,12 @@ def main(argv=None):
         if method not in Daemon.deprecated_methods:
             print('{} is not a valid command.'.format(method))
             return 1
+
         new_method = Daemon.deprecated_methods[method].new_command
+        if new_method is None:
+            print("{} is permanently deprecated and will be removed in a future release.".format(method))
```
eukreign: This returns 0 but the message says "will be removed in a future release". If the command isn't run then isn't it basically removed now?

hackrush01:
Removing would mean that the message would say "invalid command", as the
command with cease to exist anymore.

hackrush01:
```python
>>> class Existentiality:
...     def exist(self):
...         print("This method does not do anything, just prints this message. It is here just for existing.")

>>> being = Existentiality()
>>> being.exist()
This method does not do anything, just prints this message. It is here just for existing.

>>> being.does_not_exist()
  File "<input>", line 1, in <module>
AttributeError: 'Existentiality' object has no attribute 'does_not_exist'
```

lbrynet/cli.py
hackrush01: Basically, maybe..!! But if you ponder into the deep ponds of philosophy and the definition of what existence means can one really say that something does or does not exist? 😂

On a less serious note: The command is still callable(though deprecated and doesn't return anything) so it is not completely removed(we can remove it completely if needed). When calling the command is an error then it will be said to be completely removed.

(Please do not solve your philosophical questions related to existence using this argument, some things are different for programs as compared to philosophy)

eukreign: The message should then say: `"{} is permanently deprecated and does not have a replacement command."`

PR #1493, opened 2018-09-28
fix wallet directory not being created

jackrobison:


jackrobison:
@eukreign I'm going to merge this since right now this issue breaks first run is broken on master. If we decide to migrate to a single directory that can be added before bumping out of rc

PR #1495, opened 2018-10-01
[DHT] retry on protocol level

shyba:
this makes retry happens at the protocol level, so we don't disrupt higher level operations due packet loss, which is expected on UDP protocols anyway.

### DIFF
--- a/lbrynet/dht/protocol.py
+++ b/lbrynet/dht/protocol.py
@@ -120,7 +120,7 @@ class KademliaProtocol(protocol.DatagramProtocol):
 
     @defer.inlineCallbacks
     def sendRPC(self, contact, method, args):
-        while True:
+        for _ in range(constants.rpcAttempts):
             try:
                 response = yield self._sendRPC(contact, method, args)
                 return response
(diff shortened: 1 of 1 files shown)


lbrynet/dht/protocol.py
```diff
@@ -118,7 +118,20 @@ def _migrate_outgoing_rpc_args(self, contact, method, *args):
...
+    @defer.inlineCallbacks
     def sendRPC(self, contact, method, args):
+        while True:
+            try:
+                response = yield self._sendRPC(contact, method, args)
+                return response
+            except TimeoutError:
+                if contact.contact_is_good:
+                    log.debug("RETRY %s ON %s", method, contact)
+                    continue
```
jackrobison: I think there should be some delay here.

It seems like this could get into nasty feedback loops where some portion of sent rpc calls keep spinning while others get through with a high enough frequency to keep the contact marked as good.

shyba: changed the infinite loop to obey the max rpcAttempts. So it will give up on high load + high packet loss. (sanity check)

Issue #929, opened 2017-09-29
Drop disable_max_key_fee


kauffj:
If `max_key_fee` is set, I have a max key fee. If it's not, I don't.

There's no need to use two values for this.

walidmujahid:
If no one has taken this task on already, I would like to add it to the tasks/issues list I will be working on.

tzarebczan:
This is a good one to start with, thanks for grabbing it!

walidmujahid:
https://github.com/lbryio/lbry/issues/1527

eukreign:
Dropped `--disable-max-key-fee` as a separately stored boolean and instead there is just `max_key_fee: null` in the daemon_settings.yml, since setting things null via command line is difficult there is new `--no-max-key-fee` which sets `max_key_fee: null`. Also, passing a new max key fee via cli is now using this format: `--max-key-fee 3.0 USD` (in daemon_settings.yml it remains unchanged for backwards compatibility)

https://github.com/lbryio/lbry/pull/1815

PR #1507, opened 2018-10-03
Drop "disable_max_key_fee"

walidmujahid:
From: https://github.com/lbryio/lbry/issues/929

- [x] Rmeove `disable_max_key_fee` and roll functionality into `max_key_fee`.
- [ ] Test `check_fee_and_convert`

### DIFF
--- a/tests/unit/lbrynet_daemon/test_Downloader.py
+++ b/tests/unit/lbrynet_daemon/test_Downloader.py
@@ -80,11 +80,10 @@ class GetStreamTests(unittest.TestCase):
         peer_finder = DummyPeerFinder()
         blob_manager = mock.Mock(spec=DiskBlobManager)
         max_key_fee = {'currency': "LBC", 'amount': 10, 'address': ''}
-        disable_max_key_fee = False
         data_rate = {'currency': "LBC", 'amount': 0, 'address': ''}
         getstream = Downloader.GetStream(
             sd_identifier, wallet, exchange_rate_manager, blob_manager, peer_finder, DummyRateLimiter(), prm,
-            storage, max_key_fee, disable_max_key_fee, timeout=3, data_rate=data_rate
+            storage, max_key_fee, timeout=3, data_rate=data_rate
         )
         getstream.download_manager = mock.Mock(spec=DownloadManager)
         return getstream
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -263,7 +263,6 @@ ADJUSTABLE_SETTINGS = {
     # TODO: writing json on the cmd line is a pain, come up with a nicer
     # parser for this data structure. maybe 'USD:25'
     'max_key_fee': (json.loads, {'currency': 'USD', 'amount': 50.0}),
-    'disable_max_key_fee': (bool, False),
     'min_info_rate': (float, .02),  # points/1000 infos
     'min_valuable_hash_rate': (float, .05),  # points/1000 infos
     'min_valuable_info_rate': (float, .05),  # points/1000 infos
--- a/lbrynet/extras/daemon/Daemon.py
+++ b/lbrynet/extras/daemon/Daemon.py
@@ -872,7 +872,6 @@ class Daemon(AuthJSONRPCServer):
                          [--download_timeout=<download_timeout>]
                          [--peer_port=<peer_port>]
                          [--max_key_fee=<max_key_fee>]
-                         [--disable_max_key_fee=<disable_max_key_fee>]
                          [--use_upnp=<use_upnp>]
                          [--run_reflector_server=<run_reflector_server>]
                          [--cache_time=<cache_time>]
@@ -895,7 +894,6 @@ class Daemon(AuthJSONRPCServer):
                                                           }.
                                                           In the CLI, it must be an escaped JSON string
                                                           Supported currency symbols: LBC, USD, BTC
-            --disable_max_key_fee=<disable_max_key_fee> : (bool) False
             --use_upnp=<use_upnp>            : (bool) True
             --run_reflector_server=<run_reflector_server>  : (bool) False
             --cache_time=<cache_time>  : (int) 150
@@ -927,7 +925,6 @@ class Daemon(AuthJSONRPCServer):
             'cache_time': int,
             'reflect_uploads': bool,
             'share_usage_data': bool,
-            'disable_max_key_fee': bool,
             'peer_search_timeout': int,
             'sd_download_timeout': int,
             'auto_renew_claim_height_delta': int
--- a/lbrynet/extras/daemon/Daemon.py
+++ b/lbrynet/extras/daemon/Daemon.py
@@ -368,7 +368,7 @@ class Daemon(AuthJSONRPCServer):
             self.streams[sd_hash] = GetStream(
                 self.file_manager.sd_identifier, self.wallet_manager, self.exchange_rate_manager, self.blob_manager,
                 self.component_manager.peer_finder, self.rate_limiter, self.payment_rate_manager, self.storage,
-                conf.settings['max_key_fee'], conf.settings['disable_max_key_fee'], conf.settings['data_rate'],
+                conf.settings['max_key_fee'], conf.settings['data_rate'],
                 timeout
             )
             try:
--- a/lbrynet/extras/daemon/Downloader.py
+++ b/lbrynet/extras/daemon/Downloader.py
@@ -32,7 +32,7 @@ log = logging.getLogger(__name__)
 
 class GetStream:
     def __init__(self, sd_identifier, wallet, exchange_rate_manager, blob_manager, peer_finder, rate_limiter,
-                 payment_rate_manager, storage, max_key_fee, disable_max_key_fee, data_rate=None, timeout=None,
+                 payment_rate_manager, storage, max_key_fee, data_rate=None, timeout=None,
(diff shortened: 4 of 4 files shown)


jackrobison:
Please rebase to master. The docs files are automatically generated from the docstrings - so they shouldn't have any changes to them directly. Otherwise, looks good to me :+1:

jackrobison:
@walidmujahid Upon thinking about it a little more I realized this setting is not redundant. I think it's fine to drop it as a separate setting, but the functionality should be rolled in to the max_key_fee setting (such as disabling by setting it to -1)

walidmujahid:
> @walidmujahid Upon thinking about it a little more I realized this setting is not redundant. I think it's fine to drop it as a separate setting, but the functionality should be rolled in to the max_key_fee setting (such as disabling by setting it to -1)

I see @jackrobison. I can roll the functionality accordingly. Do you want me to do so and then do the rebase to master?

> Thanks for the PR! Question: If a user currently has this setting in their daemon_settings.yml, will it clear it out for them?

I am not sure @tzarebczan 

> Can we send you some LBC in appreciation for your efforts?

I would not say no to that :-) After all issues are fixed and the PR is accepted, do you want me to get with you on this on the discord?

jackrobison:
>I see @jackrobison. I can roll the functionality accordingly. Do you want me to do so and then do the rebase to master?

@walidmujahid :+1:

walidmujahid:
@jackrobison I am not fully sure about the way I went about doing it. Could I get some feedback?

walidmujahid:
> Question: If a user currently has this setting in their daemon_settings.yml, will it clear it out for them?

@tzarebczan I looked into it and it does not seem to clear out automatically when the API server restarts. Rather, the daemon_settings.yml is updated whenever something is changed via `settings_set`.

tzarebczan:
Maybe not automatically clear out all, but for cases like this where we are removing a setting, it may be worth to avoid confusion. I'll let @jackrobison / @eukreign decide how we should handle this/other cases.

lbrynet/daemon/Downloader.py
```diff
@@ -99,12 +97,15 @@ def set_status(self, status, name):
 
     @defer.inlineCallbacks
     def check_fee_and_convert(self, fee):
-        max_key_fee_amount = self.convert_max_fee()
+        currency, amount = self.max_key_fee['currency'], self.max_key_fee['amount']
+        max_key_fee_amount = self.convert_max_fee(currency, amount)
         converted_fee_amount = self.exchange_rate_manager.convert_currency(fee.currency, "LBC",
                                                                            fee.amount)
```
eukreign: Doesn't the above two lines do the same thing but with different values? Why is `self.convert_max_fee()` called for one but `converted_fee_amount` calculated here? Can we drop `self.convert_max_fee` if it's not needed?

lbrynet/daemon/Downloader.py
```diff
@@ -89,8 +88,7 @@ def check_status(self):
         else:
             log.debug("Waiting for stream descriptor (%i seconds)", self.timeout_counter)
 
-    def convert_max_fee(self):
-        currency, amount = self.max_key_fee['currency'], self.max_key_fee['amount']
+    def convert_max_fee(self, currency, amount):
         return self.exchange_rate_manager.convert_currency(currency, "LBC", amount)
```
eukreign: Why these changes?

lbrynet/daemon/Downloader.py
```diff
@@ -99,12 +97,15 @@ def set_status(self, status, name):
...
+        currency, amount = self.max_key_fee['currency'], self.max_key_fee['amount']
+        max_key_fee_amount = self.convert_max_fee(currency, amount)
         converted_fee_amount = self.exchange_rate_manager.convert_currency(fee.currency, "LBC",
                                                                            fee.amount)
+        is_disabled_max_key_fee = True if amount < 0 else False
+
         if converted_fee_amount > (yield self.wallet.default_account.get_balance()):
             raise InsufficientFundsError('Unable to pay the key fee of %s' % converted_fee_amount)
-        if converted_fee_amount > max_key_fee_amount and not self.disable_max_key_fee:
+        if converted_fee_amount > max_key_fee_amount and not is_disabled_max_key_fee:
```
eukreign: This is really verbose and redundant:
```
is_disabled_max_key_fee = True if amount < 0 else False
```
above can be represented much more succinctly as:
```
is_disabled_max_key_fee = amount < 0
```
but since the `is_disabled_max_key_fee` isn't used anywhere else and is making a double negative `(not amount < 0)` you could make this much cleaner and easier to read by removing the `not` and flipping the less than to greater than:
```
 if converted_fee_amount > max_key_fee_amount and amount > 0:
```
unless I'm missing something in the logic...

walidmujahid: No, you are not missing anything. It makes complete sense.

lbrynet/daemon/Downloader.py
walidmujahid: >Doesn't the above two lines do the same thing but with different values?

By "above two lines", are you referring to the newly inserted lines, or are you referrring to the use of `self.convert_max_fee` and what `converted_fee_amount` is doing?

>Why is self.convert_max_fee() called for one but converted_fee_amount calculated here? 

Are you asking why `self.convert_max_fee` is not being used for `converted_fee_amount`? If so, I agree. I originally was not paying attention to that in order to keep as close to the way it was just with my changes included. 

That probably was not the best way to go about it. I can change it accordingly. 

>Can we drop self.convert_max_fee if it's not needed?

I would not say it is not needed. It can be used for both `max_key_fee_amount`  and `converted_fee_amount`.

lbrynet/daemon/Downloader.py
walidmujahid: I had something else in mind at the time but based on an earlier comment of yours, I would say that this change would allow for the method's use for `converted_fee_amount`.

lbrynet/daemon/Downloader.py
```diff
@@ -89,22 +88,21 @@ def check_status(self):
...
         log.info("Download lbry://%s status changed to %s" % (name, status))
         self.code = next(s for s in STREAM_STAGES if s[0] == status)
 
     @defer.inlineCallbacks
     def check_fee_and_convert(self, fee):
-        max_key_fee_amount = self.convert_max_fee()
-        converted_fee_amount = self.exchange_rate_manager.convert_currency(fee.currency, "LBC",
-                                                                           fee.amount)
+        max_key_fee_amount = self.convert_max_fee(self.max_key_fee)
+        converted_fee_amount = self.convert_max_fee(fee.currency, fee.amount)
```
eukreign: Can't see how this could work. Did you run this code?

walidmujahid: Fixed it. Wrong version.

tests/unit/lbrynet_daemon/test_Downloader.py
```diff
@@ -103,6 +106,17 @@ def test_init_exception(self):
         with self.assertRaises(AttributeError):
             yield getstream.start(stream_info, name, "deadbeef" * 12, 0)
 
+    @defer.inlineCallbacks
+    def test_getstream_convert_fee_method(self):
+        getstream = self.init_getstream_with_mocs()
+        name = 'test'
+        stream_info = None
+
+        yield self.assertTrue(moc_convert_fee("USD", 60))
```
eukreign: What is this line testing?

tests/unit/lbrynet_daemon/test_Downloader.py
```diff
@@ -103,6 +106,17 @@ def test_init_exception(self):
...
+        yield self.assertRaises(TypeError, getstream.convert_fee, "USD:60")
+        yield self.assertRaises(TypeError, getstream.convert_fee,
+                                "USD:60", "USD", 60)
```
eukreign: `assertTrue` and `assertRaises` do not return `Deferreds` so I'm not sure if these test even runs, did you try running the test? Put an `self.assertTrue(False)` at the end of the test, if the test doesn't fail then it's not getting run :-p

walidmujahid: Yep, I ran the test. It did indeed run locally for me and it seems Travis ran it too. 

Maybe I am not fully understanding what you are saying.

tzarebczan:
@walidmujahid is this ready to be reviewed again? I re-assigned @eukreign to take a peek.

walidmujahid:
> @walidmujahid is this ready to be reviewed again? I re-assigned @eukreign to take a peek.

Not yet @tzarebczan. I am still struggling a bit with the tests, but I hope to have done as soon as possible.

alyssaoc:
try to re-establish contact, if not close

walidmujahid:
Still working on tests @alyssaoc.

eukreign:
The code touched by this PR has changed significantly.

eukreign:
Dropped `--disable-max-key-fee` as a separately stored boolean and instead there is just `max_key_fee: null` in the daemon_settings.yml, since setting things null via command line is difficult there is new `--no-max-key-fee` which sets `max_key_fee: null`. Also, passing a new max key fee via cli is now using this format: `--max-key-fee 3.0 USD` (in daemon_settings.yml it remains unchanged for backwards compatibility)

https://github.com/lbryio/lbry/pull/1815

PR #1515, opened 2018-10-08
Added comments for the fuctions

KellerDover:


eukreign:
`setup.py` is a standard format used by all Python projects; it's not necessary to add such high level of commenting to it. Our `setup.py` is actually very simple compared to most other projects.

If you want to learn more about how `setup.py` works there is a great amount of detail here: https://docs.python.org/3.7/distutils/setupscript.html

Issue #1159, opened 2018-03-27
Standardize claim name naming.


seanyesmunt:
Currently, there are discrepancies between with the naming of the "claim name".

When calling `file_list` this name is returned as `claim_name`
When calling `claim_list_xxx` it is returned as `name`

This should be standardized as just `name`, to avoid confusion.

PR #1518, opened 2018-10-09
#1159 replaced all instances of claim_name with name to avoid confusion

jvastbinder:


eukreign:
Closing as possible spam. If this was a legitimate PR please explain why you are proposing this change.

eukreign:
This PR creates more problems than it solves. @jvastbinder did you even bother to look at the output of your global search and replace to see if it made any sense? The ticket you reference refers to a handful of commands; nowhere does it suggest to do a blind global search and replace over the entire code base (including database table columns (!!!)).

jvastbinder:
Thanks for the feedback. I can see how I overestimated the scope of the naming discrepancy. I'll rollback my changes and give my further actions more careful thought.

PR #1524, opened 2018-10-10
set wallet max_gap after sync

### LINKED ISSUE #1519: set gap on wallet after migration
After wallet is migrated from pre-torba to torba format and we generate the same number of addresses to match the old number of addresses, we need to run `account.get_max_gap()` and update the users gap setting in their wallet to those values so that if they blow away their `blockchain.db` and we have to recreated in the future we can do it correctly. Without setting the correct gap and not having the old wallet available anymore (after migration) the gap setting is the only way we can reliably recreate the wallet with the same number of addresses.

Specifically the algorithm should be to use `account.get_max_gap()` values (add 1 to the gaps returned by `get_max_gap()`) if they are larger than the default gap values, otherwise if the max_gap is less than the default values, then keep the default values.

Continuation of https://github.com/lbryio/lbry/issues/1459

shyba:
closes #1519

### DIFF
--- a/lbrynet/wallet/account.py
+++ b/lbrynet/wallet/account.py
@@ -115,6 +115,13 @@ class Account(BaseAccount):
                 indent=2
             ))
 
+    @defer.inlineCallbacks
+    def save_max_gap(self):
+        gap = yield self.get_max_gap()
+        self.receiving.gap = max(20, gap['max_receiving_gap'] + 1)
+        self.change.gap = max(6, gap['max_change_gap'] + 1)
+        self.wallet.save()
+
     def get_balance(self, confirmations=0, include_claims=False, **constraints):
         if not include_claims:
             constraints.update({'is_claim': 0, 'is_update': 0, 'is_support': 0})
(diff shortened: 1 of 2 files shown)


lbrynet/wallet/account.py
```diff
@@ -115,6 +116,13 @@ def maybe_migrate_certificates(self):
                 indent=2
             ))
 
+    @defer.inlineCallbacks
+    def save_max_gap(self):
+        gap = yield self.get_max_gap()
+        self.receiving.gap = max(20, gap['max_receiving_gap'])
```
tzarebczan: This should be gap + 1 on both receiving and sending

lbrynet/wallet/account.py
```diff
@@ -33,6 +33,7 @@ def get_certificate_private_key(self, ref: TXORef):
 
     @defer.inlineCallbacks
     def maybe_migrate_certificates(self):
+        yield self.save_max_gap()
```
eukreign: Can this be moved up one level and put next to where `maybe_migrate_certificate` is called?

PR #1539, opened 2018-10-15
gather varargs

shyba:


eukreign:
Thanks for tracking this down @shyba . I manually copied the `asyncio.gather` fix into asyncio branch.

PR #1564, opened 2018-10-30
Update setup.py

naveenuchiha:
### Fixes #


eukreign:
possible spam, please comment if this is not spam

PR #1570, opened 2018-10-31
added alt text

dbeezt:
### Description
Added missing alt text


eukreign:
@NetOperatorWibby this is an HTML only PR, can you confirm that this changes makes sense and is consistent with the rest of our docs/sites?

NetOperatorWibby:
@eukreign Looks good. Not sure why CI checks failed.

PR #1641, opened 2018-11-15
change links

ykris45:
i know this will be redirect to https://lbry.tech/playground, woulkd be nice if this link always up to date


jessopb:
Thanks for pointing that out :) Gonna make a single PR.

PR #1645, opened 2018-11-16
async iterative find (WIP)

jackrobison:
- [x] refactor iterative find to asyncio
- [x] refactor iterative find to be an async generator that will keep going until the caller is done with it or the network is traversed
- [ ] refactor the blob downloader client to iterate over the new iterative find generator (instead of only trying the first set of peers returned and then finding new peers from scratch upon a failure)


jackrobison:
Closed in favor of https://github.com/lbryio/lbry/tree/async-dht

Will be merged in 0.30.2

PR #1647, opened 2018-11-18
Minor fixes to files 

osilkin98:
### Description
Adds docstrings to a couple of methods that I was looking at. A Few of the ones I added seemed good enough to include into the master branch so I decided to commit them here. Also polishes up the mechanics of a couple of methods. 

### Fixes 
1. `lbrynet/blob/blob_file.py`: Verification of a base16 formatted string was done in a crude way, essentially taking in a string and for each character, calling a *separate* function, which checks if *that* character matches one of the 16 hexadecimal characters. This gives an average runtime of `O(n^2)`, which is extremely inefficient. The new method simply goes through and converts each character in the string to base16 and checks to see that the string passed in can convert to a base-16 number in the interval [0, 15], indicating that it is a single character. If a non-base16 character is passed in, e.g. 'F', the conversion throws ValueError, which when caught, returns False. The entire operation is `O(n)`. 

2. `lbrynet/blob/blob_file.py`: Adds docstrings and throws `TypeError` instead of `Exception` if `length` is needed and is of a non `int` type, which includes `None`.  Subsequent exception handling takes this into account

3. `lbrynet/p2p/HTTPBlobDownloader.py`: More docstrings are added.

4. `scripts/reseed_file.py`: Converts snippet from Python2 to Python3


lbrynet/blob/blob_file.py
```diff
@@ -16,19 +16,21 @@
...
 
 
-def is_valid_hashcharacter(char):
-    return char in "0123456789abcdef"
-
+def is_hex_character(char):
+    try:
+        return 16 > int(char, 16) >= 0
+    except ValueError:  # Raised if non base16 string given
+        return False
```
eukreign: This make the code about 7x slower:
```
Python 3.7.0 (default, Aug  6 2018, 20:07:46) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import timeit
>>> timeit.timeit('"f" in "0123456789abcdef"', number=10000)
0.0009198940824717283
>>> newway = """
... try:
...   16 > int('f', 16) >= 0
... except ValueError:
...   pass
... """
>>> timeit.timeit(newway, number=10000)
0.007640283089131117
```

lbrynet/blob/blob_file.py
```diff
@@ -16,19 +16,21 @@
...
     """Checks whether the blobhash is the correct length and contains only
-    valid characters (0-9, a-f)
+    hexadecimal characters: (0-9, a-f).
 
     @param blobhash: string, the blobhash to check
 
     @return: True/False
     """
-    return len(blobhash) == blobhash_length and all(is_valid_hashcharacter(l) for l in blobhash)
+    return len(blobhash) == blobhash_length and all(is_hex_character(l) for l in blobhash)
```
eukreign: Using `unhexlify` would be about 5x faster than manually looping over individual characters and calling the `is_hex_character` function.
```
Python 3.7.0 (default, Aug  6 2018, 20:07:46) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import timeit
>>> timeit.timeit('try:\n  unhexlify("beef")\nexcept Error:\n  pass', 'from binascii import unhexlify', number=10000)
0.0037124441005289555
>>> timeit.timeit('all(is_hex(c) for c in "beef")', 'def is_hex(c): return c in "0123456789abcdef"', number=10000)
0.024959808913990855
```

lbrynet/p2p/BlobManager.py
```diff
@@ -120,6 +127,7 @@ def delete_blobs(self, blob_hashes):
     @defer.inlineCallbacks
     def _completed_blobs(self, blobhashes_to_check):
         """Returns of the blobhashes_to_check, which are valid"""
+        # This iterative method might not be the best since get_blob raises TypeError
```
eukreign: Please open an issue explaining what needs to be fixed / review with a permalink to relevant lines of code instead of a comment here.

lbrynet/p2p/HTTPBlobDownloader.py
```diff
@@ -121,6 +122,12 @@ def _download_and_get_retry_delay(self):
 
     @defer.inlineCallbacks
     def _download_blob(self, blob):
+        """ Given a `BlobFile` object, attempt to download the given
+        blob from any of the given mirrors. up to self.max_failures times.
+
+        :param BlobFile blob: The BlobFile object which we want to download
+        :return:
+        """
```
eukreign: Instead of adding the comment, can you add type annotations to the method itself.

lbrynet/p2p/HTTPBlobDownloader.py
```diff
@@ -151,6 +158,11 @@ def _download_blob(self, blob):
                 self.writers.remove(writer)
 
     def download_blob(self, blob):
+        """ Tries to download the given BlobFile object
+
+        :param BlobFile blob:
+        :return:
+        """
```
eukreign: Instead of adding the comment, can you add type annotations to the method itself

scripts/reseed_file.py
```diff
@@ -67,7 +67,7 @@ def reseed_file(input_file, sd_blob):
             continue
         blob = yield blob_manager.get_blob(blob_info['blob_hash'], True)
         if not blob.verified:
-            print "Blob {} is not verified".format(blob)
+            print("Blob {} is not verified".format(blob))
```
eukreign: :+1:

lbrynet/p2p/BlobManager.py
```diff
@@ -39,9 +39,16 @@ def stop(self):
...
+        :param str blob_hash: String representing the desired blob's hash
+        :param length: Length of the blob to download. If None is specified, then
+          the blob is assumed to already be saved in the `self.blobs` dict.
+        :raises TypeError: If the blob isn't saved in `self.blobs` and `length` is
+          passed in as a non-`int` type (`None` by default)
         """
         if length is not None and not isinstance(length, int):
-            raise Exception("invalid length type: {} ({})".format(length, str(type(length))))
+            # Shouldn't this be replaced with TypeError
+            raise TypeError("invalid length type: {} ({})".format(length, str(type(length))))
```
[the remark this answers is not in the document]
osilkin98: ValueError would not be acceptable in this case, as the Exception is only raised if the length argument is not specified, or an incorrect type is provided, making the type of length be `None` or any non-`int` type, respectively. 

As noted by the [Python 3 exceptions](https://docs.python.org/3/library/exceptions.html):

> exception TypeError
Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.
>
>This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be. If an object is meant to support a given operation but has not yet provided an implementation, NotImplementedError is the proper exception to raise.
>
>Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

lbrynet/blob/blob_file.py
osilkin98: That's weird. I *know* the fastest way would be to simply to get the single `char` and actually have it index a base16 array (or check to see if it indexes properly without the array), but since that's not available in this case, I guess converting it causes more overhead than I thought.

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
eukreign: Okay, point taken, you can leave this as you have it.

lbrynet/blob/blob_file.py
osilkin98: In this case the entire line should just be replaced with the `unhexlify` instead of what was in the commit prior to mine.

lbrynet/p2p/BlobManager.py
```diff
@@ -39,9 +39,16 @@ def stop(self):
     def get_blob(self, blob_hash, length=None):
         """Return a blob identified by blob_hash, which may be a new blob or a
         blob that is already on the hard disk
+
+        :param str blob_hash: String representing the desired blob's hash
+        :param length: Length of the blob to download. If None is specified, then
+          the blob is assumed to already be saved in the `self.blobs` dict.
+        :raises TypeError: If the blob isn't saved in `self.blobs` and `length` is
+          passed in as a non-`int` type (`None` by default)
```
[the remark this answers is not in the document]
osilkin98: I can add type hinting in the parameter list, though the comments should be kept as a reference for anyone trying to make sense of it in the future.

lbrynet/p2p/BlobManager.py
osilkin98: That comment wasn't supposed to go into the to-be-pulled, it was supposed to be part of something completely different.

lbrynet/p2p/HTTPBlobDownloader.py
osilkin98: I can add both, Sphynx docstrings are much easier to read when docstrings are rendered with reStructuredText, which is why I kept it despite its redundancy.

lbrynet/p2p/HTTPBlobDownloader.py
osilkin98: I can do that, but again it works a lot better with reStructuredText as the docstrings so I don't see why not. It may be redundant but at least it renders well.

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
eukreign: Just keep it as succinct as possible. I prefer to make the method names, the arguments names and the code cleaner rather than to have to provide comments. Comments should be a last resort when the code is too complicated and can't be made simpler.

I've seen too many times where comments were written with good intentions, then someone changed the code and now the comments don't represent the code and so readers end up having to waste time reading incorrect comments and also studying the code anyways to understand what's going on. I believe it's more efficient and a greater win-win to just write clean and simple to understand code rather to have shit code that needs to be commented.

With that caveat, I'm fine if you want to leave some comment, just please keep it short and sweet and to the point and don't include anything obvious that can't be inferred from the method name, arguments names or a quick skim of the code.

lbrynet/blob/blob_file.py
eukreign: Both the `in` operation and the `int('f', 16)` operations are probably very fast, the difference is that the `in` is at the language / assembly level while `int()` requires calling a function:
```
$ python3.7
Python 3.7.0 (default, Aug  6 2018, 20:07:46) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dis import dis
>>> dis('"f" in "0123456789abcdef"')
  1           0 LOAD_CONST               0 ('f')
              2 LOAD_CONST               1 ('0123456789abcdef')
              4 COMPARE_OP               6 (in)
              6 RETURN_VALUE
>>> dis("16 > int('f', 16) >= 0")
  1           0 LOAD_CONST               0 (16)
              2 LOAD_NAME                0 (int)
              4 LOAD_CONST               1 ('f')
              6 LOAD_CONST               0 (16)
              8 CALL_FUNCTION            2
             10 DUP_TOP
             12 ROT_THREE
             14 COMPARE_OP               4 (>)
             16 JUMP_IF_FALSE_OR_POP    24
             18 LOAD_CONST               2 (0)
             20 COMPARE_OP               5 (>=)
             22 RETURN_VALUE
        >>   24 ROT_TWO
             26 POP_TOP
             28 RETURN_VALUE
```
You can see above how much more work is the `int()` approach.

lbrynet/blob/blob_file.py
eukreign: Sure, only catch is that `unhexlify()` requires `bytes` only and fails on if you pass it a `str` so changing the function to use `unhexlify()` will require careful testing to make sure any code using `is_valid_blobhash` only passes `bytes` and not `str`.

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: I don't think we're using sphynx doc generator. I imagine that if we do use a doc generator we'd want one that can understand Python 3 style type annotations though.

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: Unnecessary code and comments adds to the technical debt of the project; we want to reduce technical debut. Also, we're not using sphynx doc generator. I imagine that if we do use a doc generator we'd want one that can understand Python 3 style type annotations though.

lbrynet/blob/blob_file.py
osilkin98: I looked into it and `unhexlify()` wouldn't be suitable since it has the following problems:
1. It only accepts even length hexadecimal digits (probably because it's made for reading binary): 
    ```python
    >>> unhexlify(b'3fd4c')
    binascii.Error: Odd-length string
    ```
2. It is indifferent to uppercase/lowercase (both previous solutions accept only lowercase hex strings)
    ```python
    >>> unhexlify(b'3fd4c5') == unhexlify(b'3FD4C5')
    True
    ```

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
osilkin98: I agree completely. However, when the function body contains a lot of asynchronous object oriented method calls that aren't formatted in a neat and easily understandable way, anyone trying to read it and maintain it will have to spend way too much time trying to figure out what the hell is going on. 

As a matter of fact, a lot of the variable and method names *are* very well written, and I appreciate that ***very*** much. The problem comes in the fact that when your functions begin exceeding 40 lines of code without any comments, it becomes difficult to understand what's going on. The Google standard states that you should *try* to limit functions to 40 lines, and if you're going to write a long block of code, you should put a block comment explaining what's going on. In most cases a lot of these functions exceed well beyond the 80 lines mark without *any* explanation as to what is going on. 

### Where Comments are Redundant:
```python
def max(a, b):
    # return the larger value
    return a if a > b else b
```

### Where Comments are Useful:
```python
def FFT_vectorized(x):
    """A vectorized, non-recursive version of the Cooley-Tukey FFT"""
    x = np.asarray(x, dtype=float)
    N = x.shape[0]

    if np.log2(N) % 1 > 0:
        raise ValueError("size of x must be a power of 2")

    # N_min here is equivalent to the stopping condition above,
    # and should be a power of 2
    N_min = min(N, 32)
    
    # Perform an O[N^2] DFT on all length-N_min sub-problems at once
    n = np.arange(N_min)
    k = n[:, None]
    M = np.exp(-2j * np.pi * n * k / N_min)
    X = np.dot(M, x.reshape((N_min, -1)))

    # build-up each level of the recursive calculation all at once
    while X.shape[0] < N:
        X_even = X[:, :X.shape[1] / 2]
        X_odd = X[:, X.shape[1] / 2:]
        factor = np.exp(-1j * np.pi * np.arange(X.shape[0])
                        / X.shape[0])[:, None]
        X = np.vstack([X_even + factor * X_odd,
                       X_even - factor * X_odd])
```

lbrynet/blob/blob_file.py
eukreign: What's a "Python 3 STL"? Do you mean AST (Abstract Syntax Tree)?

`int` is not a reserved word and not part of any AST, it's a function like any other function and just happens to be built-in.

You can even treat it like a regular variable and override it:
```
$ python3.7
Python 3.7.0 (default, Aug  6 2018, 20:07:46) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> int = 33
>>> int
33
```

lbrynet/blob/blob_file.py
eukreign: 1. A hex string is supposed to be even. I don't get your point about this being an issue. Why would we want odd length hexadecimal strings?
2. Why is being indifferent to uppercase/lowercase a problem?

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
eukreign: I think I addressed this pretty clearly in my original comment, again, to reiterate:

> I believe it's more efficient and a greater win-win to just write clean and simple to understand code rather to have shit code that needs to be commented.

You seem to agree but miss the obvious: your criticism of the code is correct, so fix the code, don't compound the technical debt by also adding comments. :-D

I agree that methods longer than 40 lines are too long. Can you fix that? Instead of adding even more lines via comments. <-- the only exception to that is if there is a performance benefit of inlining complicated algorithms in one function, but this is very rarely necessary and in this narrow case a comment is acceptable (to say why the method is massive).

No code is sacred here, if you think something can be done better please do it better. I'm a big fan of less code and simpler code. I have a special place in my heart for PRs that delete code leaving things simpler and cleaner while still functioning as before.

Have you seen this? https://github.com/lbryio/lbry/wiki/Zen-of-LBRY-Development

lbrynet/p2p/HTTPBlobDownloader.py
osilkin98: Sphynx does understand Python 3 type annotations. The above isn't supposed to be a good example, just something I wrote for myself and figured wouldn't do any harm if included. 

That being said, Sphynx makes the parameters very easy to read when being rendered.
For something like the following block:
    ```python
    def get_blob(self, blob_hash, length=None):
        """Return a blob identified by blob_hash, which may be a new blob or a
        blob that is already on the hard disk

        :param str blob_hash: String representing the desired blob's hash
        :param length: Length of the blob to download. If None is specified, then
          the blob is assumed to already be saved in the `self.blobs` dict.
        :raises TypeError: If the blob isn't saved in `self.blobs` and `length` is
          passed in as a non-`int` type (`None` by default)
        """
    ```
The docstring would render as the following:
image

Although the typing of parameters in between `:param type parameter_name: Description` can be hard to read if you're not rendering. For that reason, most people prefer the following:
    ```python
    def get_blob(self, blob_hash, length=None):
        """Return a blob identified by blob_hash, which may be a new blob or a
        blob that is already on the hard disk

        :param blob_hash: String representing the desired blob's hash
        :type blob_hash: str
        :param length: Length of the blob to download. If None is specified, then
          the blob is assumed to already be saved in the `self.blobs` dict.
        :type length: int or None
        :raises TypeError: If the blob isn't saved in `self.blobs` and `length` is
          passed in as a non-`int` type (`None` by default)
        """
    ```
The rendering would still be the same:
image

lbrynet/p2p/HTTPBlobDownloader.py
osilkin98: >Unnecessary code and comments adds to the technical debt of the project; we want to reduce technical debt

I don't know what you mean by this. The entire project is written in Python, a language which sacrifices performance in exchange for readability. Not sure how an interpreter skipping comment lines is going to make it any slower.

lbrynet/blob/blob_file.py
osilkin98: I don't know, I assumed that since the `int(string, base)` only accepts lowercase strings to be parsed for base16 conversion, and that the first method only includes lowercase characters, the project's standard is lowercase hexadecimal characters. This isn't a problem with me, I personally want as much fault tolerance as possible.

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: If you google "technical debt" you'll get a ton of articles, here are a few:

- https://blog.professorbeekums.com/all-code-is-debt/
- https://www.bmc.com/blogs/technical-debt-explained-the-complete-guide-to-understanding-and-dealing-with-technical-debt/
- https://hackernoon.com/there-are-3-main-types-of-technical-debt-heres-how-to-manage-them-4a3328a4c50c

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: We are using [mypy](http://mypy-lang.org/) on [torba](https://github.com/lbryio/torba) and plan to use it in lbry repo very soon as well. This means using Python 3 style type annotations. If we use Python 3 style type annotations and also add sphynx doc strings with the same information it leads to huge technical debt.

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
osilkin98: No, I haven't actually. I asked about a standard in #dev, and was told that there is no set standard, and so I just defaulted to the Google standard. 

I wholeheartedly agree with you, though. I think that if you can reduce something and make it simpler and easier to understand, you should. Sometimes the best solutions are the simplest ones, while other times, they are more complicated and thus require a more thorough explanation. This is *why* I want to be adding comments in the first place. This code is extremely messy, and so it helps people to get a grasp and hopefully simplify the process in the first place.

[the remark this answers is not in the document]
eukreign: At a minimum we use `pylint` and this is actually what failed your PR. You have an unused import (which will get resolved if you use it in a python 3 style annotation).

The standards currently applied / will be applied soon:
1. `pylint` runs automatically on all PRs
2. `mypy` not yet integrated but will be soon (which means use Python 3 type annotations)
3. https://github.com/lbryio/lbry/wiki/Zen-of-LBRY-Development

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: That's not correct, `mypy` statically analyzes the entire code base and enforces that the types are what you say they are. It's not just "typehinting". We use it on torba already which is the wallet code for lbrynet (i gave links to mypy and torba in previous comment).

eukreign: We are starting to use it in lbrynet, but it's going to take some time to get all of the code updated before we can start enforcing it, this is a good example of a file that uses lots of type annotations: https://github.com/lbryio/lbry/blob/master/lbrynet/extras/wallet/transaction.py

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
osilkin98: Oh wow, that was stupid of me. I imported that since it references the class that is being passed. I'm just going to use MyPy type hinting since it actually makes use of the import that I did JUST for that comment. 

Thank you for bringing that to my attention, although this PR was very stupid, you've taught a lot in terms of what to do and what not to do. I'm not sure how to edit the PR to put in the correct changes, so I guess I'll just close this one and reopen another one that does it correctly. 

Also you were right, starting out with a small pull request was a good choice once I actually started looking under the hood and playing around with it, so thank you for the advice.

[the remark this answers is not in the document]
eukreign: Sounds good. Like I said, no code is sacred in lbry, feel free to improve anything you want. We believe in breaking things and moving forward quickly. Looking forward to your future PRs.

lbrynet/p2p/HTTPBlobDownloader.py
osilkin98: I'll definitely use it, but it doesn't change the fact that a lot of this code is already complicated, and rewriting something that wasn't written well is much harder logistically, than to simply add a docstring and fix it when you have the time to. Do you know if there's a UML diagram for the flow of the project? If not, there should be some made going forward, it would probably allow us to spot redundancies in design before anyone has to implement it.

lbrynet/p2p/BlobManager.py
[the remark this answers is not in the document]
osilkin98: reply me on slack

lbrynet/p2p/HTTPBlobDownloader.py
eukreign: Better to just fix it right away as you're learning it.

Haven't had time for UML diagrams, by the time a UML diagram is created the code will probably change and the UML diagram will be out of date.

Less code = quicker to understand = quicker to change = quicker to innovate and move forward.

The more docs, UML diagrams and other stuff you add the more technical debt you have and the more bogged down you get. Now making a simple code change requires also updating docs and updating UML diagram, etc. Or more likely, people will only make the code change and now the docs and UML diagram are out of date.

It's much more efficient to write clear and tight code that's easy to understand and easy to modify.

Another really good way to document an API is to write readable unit tests, if they clearly show inputs/outputs it does two things: 1) provide working examples and 2) is "living" documentation because if the code changes the tests have to change, otherwise tests fail. This is *much* better than static written docs which get stale when code changes.

osilkin98: If you're moving forward all the time and not looking ahead, you lose sight of the big picture. Is it tedious to draw out a diagram of what the system is? Maybe, but at least when you have one done, you realize "oh well we can get rid of Y and Z because they do the same thing as X, just slightly differently". 

It's good practice to move fast and break stuff, but if you're not thinking about what you're doing then you're just running around aimlessly. While it's great if you can reduce everything down into a purely functional format, that's not always going to be the case, and not everyone will follow that. Hence why you should take out a minute to document what your function is doing so you can think "well huh, this is overly complicated, I can make it simpler"

eukreign: We do brain storming on a white board, this includes sketching stuff roughly resembling UML diagram but much less formal. We did this for the new wallet architecture. But the final design changed enough that the original drawn diagrams don't really reflect reality.

I'm a visual person so totally support quick sketches and diagrams. I just don't want to waste time on formal UML diagrams with expectations that they will be kept updated, etc.

osilkin98: I completely understand that. Unfortunately I've never even seen the original so I proposed the UML thing as it would help anyone who doesn't have the privilege of seeing that in person.

eukreign: If we do anything like that again I'll be sure to post a photo into slack.

PR #1660, opened 2018-11-22
Removes `six` From Project

### LINKED ISSUE #1656: drop `six` dependency
We have some legacy code using `six`, someone with knowledge of Python but maybe not as much knowledge of LBRY can learn the codebase by doing some simple refactoring.

Basically, the following code needs to be updated to drop `six` in favor of pure Python 3.
```
[lbry]$ grep -n -r six * | grep -v Binary | grep -v \.html | grep -v \.toc | grep -v \.txt
lbrynet/extras/daemon/auth/server.py:3:from six.moves.urllib import parse as urlparse
lbrynet/extras/wallet/claim_proofs.py:1:import six
lbrynet/extras/wallet/claim_proofs.py:35:            to_hash += six.int2byte(child['character'])
lbrynet/schema/hashing.py:1:import six
lbrynet/schema/hashing.py:6:    if isinstance(x, six.text_type):
lbrynet/schema/hashing.py:16:    if isinstance(x, six.text_type):
lbrynet/schema/decode.py:4:import six
lbrynet/schema/decode.py:51:            if six.PY3 and isinstance(claim_value, six.binary_type):
lbrynet/schema/decode.py:58:        if six.PY3 and isinstance(claim_value, six.text_type):
lbrynet/schema/base.py:1:import six
lbrynet/schema/base.py:10:if six.PY2:
lbrynet/schema/base.py:16:elif six.PY3:
lbrynet/schema/address.py:1:import six
lbrynet/schema/address.py:15:    if six.PY3:
lbrynet/schema/validator.py:2:import six
lbrynet/schema/validator.py:23:    if isinstance(claim_id, six.binary_type):
setup.py:46:        'six'
tests/unit/cryptstream/test_cryptblob.py:11:from six import BytesIO
```

- [ ] refactor all code using `six` to be pure Python 3 code
- [ ] remove `six` dependency from `setup.py`

osilkin98:
### Description

Refactors code and removes compatibility library `six` from project. Effectively renders code only usable with `python3`  now. 

### Fixes #1656

### DIFF
--- a/lbrynet/extras/wallet/claim_proofs.py
+++ b/lbrynet/extras/wallet/claim_proofs.py
@@ -1,4 +1,3 @@
-import six
 import struct
 import binascii
 from torba.client.hash import double_sha256
@@ -32,7 +31,7 @@ def verify_proof(proof, rootHash, name):
                 if previous_child_character >= child['character']:
                     raise InvalidProofError("children not in increasing order")
             previous_child_character = child['character']
-            to_hash += six.int2byte(child['character'])
+            to_hash += bytes((child['character'], ))
             if 'nodeHash' in child:
                 if len(child['nodeHash']) != 64:
                     raise InvalidProofError("invalid child nodeHash")
--- a/lbrynet/schema/base.py
+++ b/lbrynet/schema/base.py
@@ -1,4 +1,3 @@
-import six
 from lbrynet.schema.schema import ADDRESS_CHECKSUM_LENGTH
 from lbrynet.schema.hashing import double_sha256
 from lbrynet.schema.error import InvalidAddress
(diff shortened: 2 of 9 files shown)


lbrynet/extras/wallet/claim_proofs.py
```diff
@@ -32,7 +31,7 @@ def verify_proof(proof, rootHash, name):
                 if previous_child_character >= child['character']:
                     raise InvalidProofError("children not in increasing order")
             previous_child_character = child['character']
-            to_hash += six.int2byte(child['character'])
+            to_hash += bytes((child['character'], ))
```
eukreign: Extraneous space after the comma?

lbrynet/schema/base.py
```diff
@@ -7,18 +6,11 @@
...
-    iseq, bseq, buffer = (
-        lambda s: s,
-        bytes,
-        lambda s: s.buffer,
-    )
+iseq, bseq, buffer = (
+    lambda s: s,
+    bytes,
+    lambda s: s.buffer,
+)
```
eukreign: Can you follow this refactor further? For example, there is no benefit to renaming `bytes` to `bseq`, instead it would be better to just use `bytes` where `bseq` currently appears. Same goes for `iseq` (which doesn't even do anything) and `buffer`.

eukreign: To clarify, `iseq` should be dropped altogether.

lbrynet/extras/wallet/claim_proofs.py
osilkin98: The comma is needed so that `child['character']` gets interpreted as a `tuple` rather than an `int`:
```python
>>> u = ((3))
>>> u
3
>>> v = ((3, ))
>>> v
(3,)
>>> bytes(u)
b'\x00\x00\x00'
>>> bytes(v)
b'\x03'
```
The space preceding the comma makes it look less cluttered.

lbrynet/schema/base.py
osilkin98: Done

lbrynet/extras/wallet/claim_proofs.py
eukreign: Can you please remove the space after the comma.

PR #1677, opened 2018-11-29
icons!

eukreign:
what title says is what it is


kauffj:
If this doesn't need to be merged immediately, @NetOperatorWibby is currently working on a project to source assets like this from the same origin so that they're not littered repeatedly throughout our repos.

@eukreign would it be tricky to source these from https://github.com/lbryio/design? The two of you should discuss.

eukreign:
@kauffj , i believe @tzarebczan really wants this in the current release. we can change it later. (in fact, all of these icons are from this same repo, the folder was deleted before and I just brought it back).

kauffj:
:+1:

PR #1687, opened 2018-12-01
Removes Faker from Dependencies

### LINKED ISSUE #1667: drop `faker` dependency
Drop `faker` dependency `setup.py` along with the entire `extras_require` section and refactor the one test case that uses it to just have raw static data directly.

- [ ] refactor https://github.com/lbryio/lbry/blob/ffae10d11f01520c253a2f2a1c668979ecd952be/tests/unit/lbrynet_daemon/test_Daemon.py#L150 to not use `Faker` and instead run the test in the debugger with the existing `Faker` taking note of the generated values, then simply statically include those values where the generated ones would have be returned.
- [ ] remove `faker` from `setup.py` along with the entire `extras_require` section
- [ ] update installation docs to just say `pip install -e .` as `.[test]` is no longer needed

osilkin98:
## Fixes issue #1667 

Removes the `faker` module from the project's dependencies. The module performs only one generation of test data for `unit.tests.lbrynet_daemon.test_Daemon.py` in the class `TestFileListSorting`. The module was responsible for generating attributes for 10 files of which certain information was known ahead of time.

As a result, we can drop it from `setup.py`, and remove the `extras_require` section with it. This reduces the installation command from `pip install -e .[test]` to `pip install -e .`.

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes)
- [x] Documentation changes
- [] Other - Please describe:


osilkin98:
That's perfectly fine @tzarebczan, I've just been working on impementing the comment system so it's no big deal

tests/unit/lbrynet_daemon/test_Daemon.py
```diff
@@ -269,43 +264,10 @@ def test_invalid_sort_produces_meaningful_errors(self):
             yield self.test_daemon.jsonrpc_file_list(sort=sort_options)
 
     def _get_fake_lbry_files(self):
-        return [self._get_fake_lbry_file() for _ in range(10)]
+        return list(map(lambda meta: self._get_fake_lbry_file(meta), FAKED_ATTRIBUTES))
```
eukreign: Please change to:
`[self._get_fake_lbry_file(meta) for meta in FAKED_ATTRIBUTES]`

tests/unit/lbrynet_daemon/_sample_file_list.py
```diff
@@ -0,0 +1,237 @@
+if __name__ == '__main__':
+    from sys import exit
+    exit()
+
```
eukreign: Can you remove above conditional and also move this entire `FAKED_ATTRIBUTES` list into the file that imports it, just all the way at the bottom.

eukreign:
@osilkin98 please see comments above, ping me on chat if you have more questions, etc.

tests/unit/lbrynet_daemon/_sample_file_list.py
osilkin98: Boom

tests/unit/lbrynet_daemon/test_Daemon.py
osilkin98: Done

PR #1704, opened 2018-12-10
DHT and blob exchange to use `asyncio`

jackrobison:


alyssaoc:
The buddy is Victor

jackrobison:
https://github.com/lbryio/lbry/pull/1769

PR #1709, opened 2018-12-11
Deletes unused test scripts

osilkin98:
## PR Type 
Removes scripts that perform similar functionality to that in `tests.unit`. 
I was looking through the folder and started refactoring legacy code when I realized that most of the stuff isn't doing anything that isn't provided by `tests.unit` and since they're not referenced anywhere within the actual project, I just went ahead and deleted them. 

## Why is this change necessary?
Reduces the overall weight of the project. The scripts are basically just copies of things found in `tests.unit` and don't belong in the project. There aren't any usages of them anywhere that I could find. 
If they're for testing individually for something the daemon does at runtime, they should either be moved off the git tree and into local storage or in a directory `tests/runtime-tests/`.


tzarebczan:
Thanks for the PR @osilkin98! We love ones that delete unncessary lines of code :)

eukreign:
I don't use these scripts but I think others might. @jackrobison @shyba please chime in if you want to keep these.

eukreign:
If we're using the scripts then we should keep this. They also serve as examples on how to use the SDK / APIs.

osilkin98:
@eukreign The scripts in this PR are very dated, using python 2 syntax and also don't provide any *good* example or insight as to how the internals work.

eukreign:
Can we delete them one at a time as equivalent functionality is added to daemon and/or updated versions of the scripts that do the same things are written?

osilkin98:
There are already equivalent scripts in the `tests` directory. I *do* agree that we should have examples, so maybe we could create a separate `examples` directory at the root level and put new ones there? These should still be deleted though, they're not very explanatory and I wouldn't say that they constitute as good examples.

PR #1717, opened 2018-12-14
fix directory cli args and update help message

jackrobison:


tzarebczan:
```2018-12-14 13:35:25,429 INFO     lbrynet.extras.wallet.manager:193: Wallet at c:\lbrynew3\wallets\default_wallet is empty, generating a default account.``` 
It tries to create my wallet in the data_dir, not sure why

Per our convo, don't forget to support legacy lbryum_wallet_dir in config file.

PR #1727, opened 2018-12-26
Add resolve caching

hackrush01:
**Known Problems:**
`valid_at_height` field is not cached as it is not reliably present in *non-cached* resolve calls
`claims_in_channel` field is not present. Some things will be modified for adding it and I'm not exactly sure what(maybe another column is needed to be added in the table) also might be breaking for app as it is currently a top level property of a certificate resolve and even a claim(which is wrong per discussion during meeting) so it will be moved in as a field of `certificate`

### DIFF
--- a/lbrynet/extras/wallet/dewies.py
+++ b/lbrynet/extras/wallet/dewies.py
@@ -30,4 +30,4 @@ def lbc_to_dewies(lbc: str) -> int:
 
 
 def dewies_to_lbc(dewies) -> str:
-    return satoshis_to_coins(dewies)
+    return dewies if isinstance(dewies, str) else satoshis_to_coins(dewies)
--- a/lbrynet/extras/wallet/manager.py
+++ b/lbrynet/extras/wallet/manager.py
@@ -10,6 +10,7 @@ from typing import Optional
 from twisted.internet import defer
 
 from lbrynet.schema.schema import SECP256k1
+from lbrynet.schema.uri import parse_lbry_uri
 from torba.client.basemanager import BaseWalletManager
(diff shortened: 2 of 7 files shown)


lbrynet/extras/wallet/dewies.py
```diff
@@ -30,4 +30,4 @@ def lbc_to_dewies(lbc: str) -> int:
 
 
 def dewies_to_lbc(dewies) -> str:
-    return satoshis_to_coins(dewies)
+    return dewies if isinstance(dewies, str) else satoshis_to_coins(dewies)
```
hackrush01: I'm skeptical about this change. This was needed because the claim was re-processed. In which case dewies would already be in string(desired) format.

lbrynet/extras/wallet/manager.py
```diff
@@ -268,16 +273,55 @@ def get_wallet_info_query_handler_factory(self):
...
+                                                                         certificate=result.get(
+                                                                             "certificate", None))
+        if is_channel:
+            result = {"certificate": result["claim"]}
+        return result
+
     async def resolve(self, *uris, **kwargs):
         page = kwargs.get('page', 0)
         page_size = kwargs.get('page_size', 10)
         check_cache = kwargs.get('check_cache', False)  # TODO: put caching back (was force_refresh parameter)
```
shyba: remove "TODO" comment, you did it! \o\

eukreign:
closing this for now, we can re-open with a new PR once the other major refactoring PRs are merged and it turns out we still need this caching

PR #1745, opened 2019-01-08
Removes twisted dependency from lbrynet.extras.daemon.analytics.py

osilkin98:
## PR Checklist

## PR Type
This removes the use of `treq` and `twisted` modules from being used in `analytics.py`. It also removes the old `LoopingCall` mechanism and replaces its manager component with a `dict` that stores the looping Tasks as a `name -> Task` key value pair, which makes use of the `looping_call` function from `lbrynet.extras.daemon.storage.py`. Modification is also made to the `looping_call` function in order to catch the task cancellation error when being wrapped as a task coroutine. 

Instead of `treq`, the `aiohttp` library is used, although instead of creating a `ClientSession` object, I use the `aiohttp.request` ContextManager to set the `Api`'s cookies as that of the response's.
 
Why is this change necessary?
To remove the godawful dependency that is `twisted`

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting)
- [x] Refactoring (no functional changes)
- [ ] Documentation changes
- [x] Other - Please describe: Removes unwanted dependencies

### DIFF
--- a/lbrynet/extras/daemon/Components.py
+++ b/lbrynet/extras/daemon/Components.py
@@ -696,7 +696,7 @@ class UPnPComponent(Component):
                     log.debug("set up upnp port redirects for gateway: %s", self.upnp.gateway.manufacturer_string)
         else:
             log.error("failed to setup upnp")
-        self.component_manager.analytics_manager.send_upnp_setup_success_fail(success, await self.get_status())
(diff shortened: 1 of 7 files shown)


jackrobison:
@osilkin98 can you rebase/squash this down to one or two commits on top of master?

PR #1794, opened 2019-01-22
refactor exchange rate manager to use asyncio

jackrobison:


lbrynet/extras/daemon/exchange_rate_manager.py
```diff
@@ -77,21 +74,23 @@ def _on_error(self, err):
...
-        d.addErrback(self._on_error)
-        return d
+    async def _update_price(self):
+        while True:
+            try:
+                response = await asyncio.wait_for(self._make_request(), self.REQUESTS_TIMEOUT)
+                self._save_price(self._subtract_fee(self._handle_response(response)))
+            except (asyncio.CancelledError, asyncio.TimeoutError, InvalidExchangeRateResponse) as err:
+                self._on_error(err)
+            await asyncio.sleep(self.EXCHANGE_RATE_UPDATE_RATE_SEC)
```
eukreign: Should this catch float conversion errors and KeyError. Most of the fee code involves accessing keys (without checking if server response is correct) and then blindly wrapping it in `float()`.

Also, why does it catch `CancelledError`?

jackrobison: fixed

PR #1820, opened 2019-01-28
[WIP] android build of lbrynet as an IPC service

eukreign:
- [ ] get PR to point where lbrynet daemon starts and continues running - 2pt
- [ ] expose API in service - 2pt
- [ ] sample app or automated test which calls at least one API method - 2pt
- [ ] generate an `.aar` instead of `.apk` file which can be included in `lbry-android` app - 2pt


lyoshenka:
closing. will reopen when we revisit this issue

PR #1841, opened 2019-02-01
General fixes downloader

shyba:


lbrynet/blob/blob_file.py
```diff
@@ -96,24 +96,24 @@ def callback(finished: asyncio.Future):
                 raise error
         return callback
 
-    async def save_verified_blob(self, writer):
+    async def save_verified_blob(self, writer, verified_bytes):
```
jackrobison: can you type this?

lbrynet/stream/stream_manager.py
```diff
@@ -204,7 +204,8 @@ def wait_for_stream_finished(self, stream: ManagedStream):
             self.wait_for_stream_finished(stream)
             return stream
         except asyncio.CancelledError:
-            await downloader.stop()
+            downloader.stop()
+            log.info("stopped stream")
```
jackrobison: log.debug?

PR #1862, opened 2019-02-05
adds announce tests on stream creation/assemble, refactor calls to set announce

shyba:


lbrynet/stream/assembler.py
```diff
@@ -74,24 +75,25 @@ def _decrypt_and_write():
...
         with open(self.output_path, 'wb') as stream_handle:
             self.stream_handle = stream_handle
             for blob_info in self.descriptor.blobs[:-1]:
                 while True:
                     try:
                         blob = await self.get_blob(blob_info.blob_hash, blob_info.length)
-                        await self._decrypt_blob(blob, blob_info, self.descriptor.key)
+                        if await self._decrypt_blob(blob, blob_info, self.descriptor.key):
+                            await self.blob_manager.blob_completed(blob)
                         break
```
jackrobison: Should `break` be in this `if`?

PR #1903, opened 2019-02-12
Fix chunked response parsing

shyba:


tests/unit/stream/test_downloader.py
```diff
@@ -80,3 +84,17 @@ def _mock_accumulate_peers(q1, q2):
...
+
+            def send_response(self, responses):
+                to_send = []
+                while responses:
+                    to_send.append(responses.pop())
+                for byte in BlobResponse(to_send).serialize():
+                    self.transport.write(bytes([byte]))
+        self.server.server_protocol_class = ChunkedServerProtocol
+        self.server.start_server(33333, '127.0.0.1')
+        await self._test_transfer_stream(10)
```
eukreign: Can this test have an assert somewhere?

shyba: _test_transfer_stream does it, but I will carry some to the test itself.
What's being really tested here is that it doesnt hang or explodes in errors (which was the fix). I will assert it transferred fine.

eukreign: For hang, is it possible to assert that it finishes in 1s or 2s, etc (and fail test if it takes longer)?

Issue #1931, opened 2019-02-18
Download may randomly stall


tzarebczan:
## The Issue
Run into this a few times now, but not sure why it happens yet. It may also be related to some reflector race condition issues @jackrobison / @lyoshenka are aware of. 

Here's a recent one that stopped during a ttfb run, and then completed after restart. 

```
2019-02-18 00:20:21,508 INFO     lbrynet.stream.stream_manager:474: download stream from singham-cameo-in-sooryavanshi-ajay-devgn#6717a2bd1ec9bb27d2a84b7343d617a344625c70
2019-02-18 00:20:21,906 INFO     lbrynet.blob_exchange.client:126: downloaded 538c5d7c from 34.239.118.206:3333
2019-02-18 00:20:21,914 INFO     lbrynet.stream.downloader:49: added head blob to search
2019-02-18 00:20:21,916 INFO     lbrynet.stream.stream_manager:302: got descriptor 538c5d7c71eade6088139ad8467d2e6ebd1dc1f1b16b85abe892eca130fff16364349a9ae8ce0cadd44da813d0d1abb1 for singham-cameo-in-sooryavanshi-ajay-devgn
2019-02-18 00:20:22,110 INFO     lbrynet.blob_exchange.client:126: downloaded 4537c503 from 34.239.118.206:3333
2019-02-18 00:20:22,240 INFO     lbrynet.blob_exchange.client:126: downloaded 4f29ac29 from 34.239.118.206:3333
2019-02-18 00:20:22,362 INFO     lbrynet.blob_exchange.client:126: downloaded c60eca73 from 34.239.118.206:3333
2019-02-18 00:20:22,457 INFO     lbrynet.blob_exchange.client:126: downloaded f5053222 from 34.239.118.206:3333
2019-02-18 00:20:22,550 INFO     lbrynet.blob_exchange.client:126: downloaded d8abeb1a from 34.239.118.206:3333
2019-02-18 00:20:22,654 INFO     lbrynet.blob_exchange.client:126: downloaded 128d76be from 34.239.118.206:3333
2019-02-18 00:20:22,843 INFO     lbrynet.blob_exchange.client:126: downloaded 8777ced4 from 34.239.118.206:3333
2019-02-18 00:20:22,993 INFO     lbrynet.blob_exchange.client:126: downloaded 3fef987c from 34.239.118.206:3333
[...]

PR #1934, opened 2019-02-18
Unban peers if no more peers to try and 10 seconds passed

shyba:
maybe related: #1931


tzarebczan:
Looks like it works here, @jackrobison please double check the code.

PR #1941, opened 2019-02-20
make dht split threshold configurable

shyba:
main use: seeders


shyba:
#1942 supersedes this

PR #1977, opened 2019-03-11
time to first bytes analytics

jackrobison:


lbrynet/stream/stream_manager.py
```diff
@@ -275,101 +276,6 @@ def stop(self):
         if delete_file and stream.output_file_exists:
             os.remove(stream.full_path)
 
-    def wait_for_stream_finished(self, stream: ManagedStream):
```
shyba: nice! simpler

PR #1984, opened 2019-03-13
Update to new protobuf schema

shyba:


lbrynet/extras/daemon/Components.py
```diff
@@ -252,23 +251,18 @@ def component(self):
 
     async def get_status(self):
         if self.wallet_manager and self.running:
-            local_height = self.wallet_manager.network.get_local_height()
-            remote_height = self.wallet_manager.network.get_server_height()
+            local_height = self.wallet_manager.ledger.headers.height
             best_hash = self.wallet_manager.get_best_blockhash()
             return {
                 'blocks': max(local_height, 0),
-                'blocks_behind': max(remote_height - local_height, 0),
```
jackrobison: why remove this?

lbrynet/extras/daemon/Daemon.py
```diff
@@ -2323,66 +2180,42 @@ def jsonrpc_claim_list_mine(self, account_id=None, page=None, page_size=None):
...
     @requires(WALLET_COMPONENT)
-    async def jsonrpc_claim_list(self, name):
+    async def jsonrpc_claim_search(self, name, channel_id=None, winning=False):
         """
-        List current claims and information about them for a given name
+        Search for claims on the blockchain.
 
         Usage:
-            claim_list (<name> | --name=<name>)
+            claim_search (<name> | --name=<name>) [--channel_id=<channel_id>] [--winning]
```
jackrobison: when would `winning` be used instead of `resolve(<name>)`? is `winning` needed?

lbrynet/extras/daemon/Components.py
eukreign: `get_server_height()` was stubbed out and not actually returning the server height. Now checking the code, the real network does have a `get_server_height()` so this can be added back. I'll do that.

lbrynet/extras/daemon/Daemon.py
eukreign: This is for paginating a huge channel. `resolve` will only return a small sampling of the latest claims in a channel. So for home page user would still do `resolve` but if they are on a screen that is showing all of the claims in a channel with pagination they would use claim search which will support filtering and better search functionality in the future.

jackrobison: When would they want non winning claims when listing those in a channel? What's the use-case for a channel owner to have two competing claims for the same name?

eukreign: Perhaps a very large channel with many contributors and if one group in the organization doesn't agree with another group in the organization and they fight over who's claim will show up. Both having the channel key to publish....

PR #2005, opened 2019-03-29
Fixed usage message for 'file delete'

eggplantbren:
## PR Checklist

- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

Just a small fix to a lbrynet usage message.


tzarebczan:
I think this is expected, since it has to be passed as a single term for the json rpc. Maybe we can make this clearer on the CLI response, but this would not be the solution.

eggplantbren:
Okay. Sorry about the bad PR. Please close

PR #2010, opened 2019-03-31
`resolve` implementation refactoring

shyba:


lbrynet/wallet/ledger.py
```diff
@@ -51,36 +50,20 @@ def __init__(self, *args, **kwargs):
...
-            except Exception as e:
-                return {'error': str(e)}
-        try:
-            resolver = self.resolver
-            resolutions = await self.network.get_values_for_uris(self.headers.hash().decode(), *uris)
-            return await resolver._handle_resolutions(resolutions, uris, page, page_size)
-        except Exception as e:
-            log.exception(e)
-            return {'error': str(e)}
+        return Resolver(self.headers, self.transaction_class, network=self.network, ledger=self)
```
eukreign: Since you're passing in the `ledger` (which has everything on it), do you need any of the other arguments? Seems like this entire call can just be `Resolver(self)`?

lbrynet/wallet/resolve.py
```diff
@@ -92,25 +88,16 @@ def __init__(self, claim_trie_root, height, transaction_class, hash160_to_addres
...
-                                                                             certificate,
-                                                                             raw)
-            else:
-                log.error("unknown response type: %s", claim_resolution_type)
+                    claim_response = _verify_proof(parsed_uri.name,
+                                                   claim_trie_root,
+                                                   claim_response,
+                                                   height, depth,
+                                                   transaction_class=self.transaction_class,
+                                                   ledger=self.ledger)
```
eukreign: Same as the other comment about already having the `ledger` which has `ledger.transaction_class` (in fact, that's where you got it from originally) so you don't need to pass it separately.

lbrynet/wallet/resolve.py
```diff
@@ -284,87 +253,56 @@ def prepare_claim_queries(start_position, query_size, channel_claim_infos):
...
+        output = tx.outputs[nOut]
         r = {
             'name': name,
-            'value': hexlify(value),
-            'claim_id': claim_id,
-            'txid': txid,
-            'nout': n,
-            'amount': amount,
-            'effective_amount': effective_amount,
+            'value': hexlify(output.script.values['claim']),
```
eukreign: Is this `value` actually used by any apps? I imagine we can just drop this from the dictionary. I think the protobuf stuff is isolated in lbrynet and doesn't need to be exposed to UI facing apps.

lbrynet/wallet/resolve.py
```diff
@@ -284,87 +253,56 @@ def prepare_claim_queries(start_position, query_size, channel_claim_infos):
...
-                        claim_sequence = result['claim_sequence']
-                        claim_script = claim_output.script
-                        decoded_name = claim_script.values['claim_name'].decode()
-                        decoded_value = claim_script.values['claim']
-                        if decoded_name == name:
-                            return _build_response(name, decoded_value, claim_id,
-                                                   tx.id, nOut, claim_output.amount,
-                                                   effective_amount, claim_sequence,
-                                                   claim_address, supports)
+                        if tx.outputs[nOut].script.values['claim_name'].decode() == name:
```
eukreign: You can just do `tx.output[nOut].claim_name` there is a `@property` on the Output object for this: 

https://github.com/lbryio/lbry/blob/63568185574afa1d53ac545638c38547177c26d2/lbrynet/wallet/transaction.py#L69

lbrynet/wallet/ledger.py
shyba: changing, good point

lbrynet/wallet/resolve.py
shyba: changing, good point

lbrynet/wallet/resolve.py
shyba: its hex while not decoded, then it becomes the value if we can decode or None and a new field 'hex' is added for displaying the hex value when its unable to decode. I will ping you in Slack to discuss some simplification for that, but yeah its complex and probably unused.

lbrynet/wallet/resolve.py
shyba: great, replacing it

PR #2026, opened 2019-04-12
adds api examples to docs

jessopb:
## PR Checklist

## PR Type

Adds api examples to lbry.tech sdk docs

Why is this change necessary?

Reduces development friction to see output examples of commands

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes)
- [x] Documentation changes
- [ ] Other - Please describe:


scripts/generate_json_api.py
```diff
@@ -66,43 +66,144 @@ class Examples(CommandTestCase):
...
+
+        await r(
+            'Check if address is mine',
+            'address', 'is_mine', an_address
+        )
+
+        # send/fund
+
+        transfer = await r(
+            'Transfer LBC',
```
eukreign: Can you make this more descriptive: "Send 2 LBC from default account to a specific account."

scripts/generate_json_api.py
```diff
@@ -66,43 +66,144 @@ class Examples(CommandTestCase):
...
+            'account', 'fund', f"--to_account=\"{account['id']}\"", "--amount=2.0", "--broadcast"
+        )
+
+        await self.on_transaction_dict(transfer)
+        await self.generate(1)
+        await self.on_transaction_dict(transfer)
+
+        await r(
+            'Get Default Account balance',
```
eukreign: random capitalization

scripts/generate_json_api.py
```diff
@@ -66,43 +66,144 @@ class Examples(CommandTestCase):
...
+
+        txlist = await r(
+            'List Transactions',
+            'transaction', 'list'
+        )
+
+        # await r(
+        #     'Show a transaction'
+        #     'transaction', 'show', f"\"{txlist[0]['txid']}\""
+        # )
```
eukreign: what's this?

scripts/generate_json_api.py
```diff
@@ -128,45 +230,104 @@ class Examples(CommandTestCase):
...
+        await r(
+            'Delete a blob',
+            'blob', 'delete', f"{bloblist[0]}"
+        )
+
+
+        # support
+
+        # destroy
```
eukreign: what's destroy?

scripts/generate_json_api.py
```diff
@@ -128,45 +230,104 @@ class Examples(CommandTestCase):
...
+
+        # NO DHT
+        # await r(
+        #     'Get a blob',
+        #     'blob', 'get', f"\"{bloblist[0]}\""
+        # )
+        # await r(
+        #     'Announce a blob',
+        #     'blob', 'announce', f"--blob_hash={bloblist[0]}"
+        # )
```
eukreign: If it doesn't work then just remove it.

jessopb: Roger

scripts/generate_json_api.py
jessopb: I couldn't get transaction show to work. I'll delete that.

scripts/generate_json_api.py
```diff
@@ -66,43 +66,144 @@ class Examples(CommandTestCase):
...
-            'Create an account.',
+            'Create an account',
             'account', 'create', '"generated account"'
         )
 
         await r(
-            'Remove an account.',
-            'account', 'remove', account['id']
+            'Add an account from seed',
+            'account', 'add', '"new account"', f"--seed=\"{account['seed']}\""
```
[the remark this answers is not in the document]
jessopb: They are two accounts. The goal here was to demonstrate account add and account create. The two accounts have different names.  account['id'] is the account named "new account" while the account being added is named "generated account". I `account remove` "new account" near the end after I use it for send/fund. Am I missing something?

scripts/generate_json_api.py
jessopb: # destroy was going to be things like abandon or delete type things. 

I wanted to organize the script into meaningful categories, but maybe that doesn't make sense.

scripts/generate_json_api.py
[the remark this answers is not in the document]
eukreign: Name of the account is not really used anywhere but the keys representing the account do matter (this is where the account_id comes from). In the example above, you create an account and then add the same account again, this creates two accounts with the same account_id. This should actually fail but it works to avoid a bricked wallet in case user manually edits their wallet file and adds the same account twice. Normally you should never add the same account twice.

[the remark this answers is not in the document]
jessopb: OH, you are very right. I will look at rearranging it.

jessopb:
I was unable to view the results of my last commit other than viewing the json file. A clean install run locally on 8080 now fails to view api docs and instead infinitely redirects/404s. I filed an issue on .tech about that.

Issue #2016, opened 2019-04-02
Raise peer and download timeouts


tzarebczan:
## The Issue
Lot's of users are still running into data blob timeouts on slower connections or with multiple downloads. We should consider raising them. Believe it's set to 1 minute or so on the reflector, but I can test this out. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

lyoshenka:
@tzarebczan this is fixed, right?

PR #2048, opened 2019-04-24
Adjust default max_connections_per_download and peer_connect_timeout settings

tzarebczan:
Address https://github.com/lbryio/lbry/issues/2016 for timeout issues until we can deploy more reflector peers. Would like @jackrobison to review.

## PR Checklist

- [x] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

Why is this change necessary?

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number: N/A

### DIFF
--- a/lbrynet/conf.py
+++ b/lbrynet/conf.py
@@ -480,7 +480,7 @@ class Config(CLIConfig):
     # protocol timeouts
     download_timeout = Float("Cumulative timeout for a stream to begin downloading before giving up", 30.0)
     blob_download_timeout = Float("Timeout to download a blob from a peer", 30.0)
-    peer_connect_timeout = Float("Timeout to establish a TCP connection to a peer", 2.0)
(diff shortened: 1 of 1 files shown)


tzarebczan:
From my testing with messing with timeout params like download/peer/blob timeout:
```It seems like changing the timeout params has little effect on time to first byte for me. I actually got one of my best runs with peer connect timeout of 6 and 2/3 min blob/download timeouts.

But without a proper dht network and available peers, it's a poor test. Aws should work everytime, and if it doesn't , we need to know why.```

tzarebczan:
So it should also be safe to increase blob_download_timeout....without this, it won't help the slower connections. 

I also recently learned we try to download multiple blobs at the same time - so this could be slowing down the download even more for users with poor connections. Something like this would need to be handled in the connection manager.

PR #2125, opened 2019-05-11
Remove search junctions, make search retry, some refactoring

shyba:


lbrynet/dht/blob_announcer.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, loop: asyncio.BaseEventLoop, node: 'Node', storage: 'SQLiteSt
         while batch_size:
             if not self.node.joined.is_set():
                 await self.node.joined.wait()
+            await asyncio.sleep(60, loop=self.loop)
+            if not self.node.protocol.routing_table.get_peers():
+                log.warning("No peers in DHT, announce round skipped")
+                continue
```
jackrobison: should the delay be before the `continue`?

lbrynet/dht/node.py
```diff
@@ -183,61 +185,38 @@ def get_iterative_value_finder(self, key: bytes, shortlist: typing.Optional[typi
...
-
     async def peer_search(self, node_id: bytes, count=constants.k, max_results=constants.k*2,
-                          bottom_out_limit=20) -> typing.List['KademliaPeer']:
-        accumulated: typing.List['KademliaPeer'] = []
-        async with self.peer_search_junction(node_id, max_results=max_results,
-                                             bottom_out_limit=bottom_out_limit) as junction:
-            async for peers in junction:
-                accumulated.extend(peers)
+                          bottom_out_limit=20, shortlist: typing.Optional[typing.List] = None
+                          ) -> typing.List['KademliaPeer']:
```
jackrobison: can you add annotations for the type contained by `shortlist`?

lbrynet/dht/node.py
```diff
@@ -183,61 +185,38 @@ def get_iterative_value_finder(self, key: bytes, shortlist: typing.Optional[typi
...
-                        if not (
-                                peer.address == self.protocol.external_ip
-                                and peer.tcp_port == self.protocol.peer_port
-                        )
-                    ])
+        tasks = []
+        async def __start_producing_task():
+            while True:
+                blob_hash = await search_queue.get()
+                tasks.append(asyncio.create_task(self._value_producer(blob_hash, result_queue)))
```
jackrobison: These tasks should be made with `self.loop.create_task`

lbrynet/dht/node.py
```diff
@@ -183,61 +185,38 @@ def get_iterative_value_finder(self, key: bytes, shortlist: typing.Optional[typi
...
+        tasks.append(asyncio.create_task(__start_producing_task()))
+        try:
+            await asyncio.wait(tasks)
```
jackrobison: loop argument on `wait`

lbrynet/dht/blob_announcer.py
shyba: It's before. `continue` will make it hit the delay

PR #2181, opened 2019-05-24
improve set_build.py, pin torba version

jackrobison:
- adds BUILD_COMMIT to lbrynet.build_type

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ dist: xenial
 language: python
 python: "3.7"
 
+env:
+  - TORBA=master
+
 jobs:
   include:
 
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -27,10 +30,9 @@ jobs:
         - bash <(curl -s https://codecov.io/bash)
 
     - name: "Integration Tests"
-      cache: false
       install:
         - pip install tox-travis coverage
-        - pushd .. && git clone --single-branch --branch master https://github.com/lbryio/torba.git && popd
+        - pip install git+https://github.com/lbryio/torba.git@${TORBA}#egg=torba
```
eukreign: this doesn't make sense, why did you change it from checking out the correct version to install with pip?

jackrobison: it installs the same version by default - `master` - determined by the `TORBA` travis variable

eukreign: install it into what?

PR #2194, opened 2019-06-03
comment client api

eukreign:


lbrynet/extras/daemon/Daemon.py
```diff
@@ -3415,34 +3417,23 @@ def jsonrpc_routing_table_get(self):
...
-            )
         comment = {
             'comment': comment,
             'claim_id': claim_id,
             'parent_id': parent_id,
-            'channel_id': channel_id,
-            'channel_name': channel_name,
-            'signature': signature
         }
+        channel = await self.get_channel_or_none(channel_account_id, channel_id, channel_name, for_signing=True)
```
osilkin98: Oh wow that's a million times better

PR #2215, opened 2019-06-07
get without fee address pays to claim address

shyba:


lbrynet/extras/daemon/Daemon.py
```diff
@@ -2598,8 +2598,8 @@ def jsonrpc_channel_list(self, account_id=None, page=None, page_size=None):
         elif old_txo.claim.is_signed and not clear_channel and not replace:
             channel = old_txo.channel
 
-        if 'fee_address' in kwargs:
-            self.valid_address_or_error(kwargs['fee_address'])
+        if 'fee_amount' in kwargs:
+            kwargs['fee_address'] = self.get_fee_address(kwargs, claim_address)
```
eukreign: mentioned this in slack as well but this introduces a bug where if you don't use `--replace` mode and just want to update the `fee_amount` this will overwrite the previously set `fee_address`

lbrynet/extras/daemon/Daemon.py
```diff
@@ -2598,8 +2598,8 @@ def jsonrpc_channel_list(self, account_id=None, page=None, page_size=None):
         elif old_txo.claim.is_signed and not clear_channel and not replace:
             channel = old_txo.channel
 
-        if 'fee_address' in kwargs:
-            self.valid_address_or_error(kwargs['fee_address'])
+        if 'fee_amount' in kwargs:
+            kwargs['fee_address'] = self.get_fee_address(kwargs, old_txo.claim.stream.fee.address or claim_address)
```
eukreign: If user is just updating the fee_address and not touching the fee_amount, this change would never validate that the fee_address is valid address.

lbrynet/extras/daemon/Daemon.py
```diff
@@ -2598,8 +2598,8 @@ def jsonrpc_channel_list(self, account_id=None, page=None, page_size=None):
         elif old_txo.claim.is_signed and not clear_channel and not replace:
             channel = old_txo.channel
 
-        if 'fee_address' in kwargs:
-            self.valid_address_or_error(kwargs['fee_address'])
+        if {'fee_currency', 'fee_amount', 'fee_address'}.intersection(kwargs.keys()):
```
eukreign: You don't need `.keys()`, this works just as well: `{'fee_currency', 'fee_amount', 'fee_address'}.intersection(kwargs)`

lbrynet/schema/attrs.py
```diff
@@ -204,24 +204,28 @@ class Fee(Metadata):
...
+            raise Exception('In order to set a fee currency, please specify a fee amount')
         if address is not None:
             self.address = address
-        if currency is not None and amount is not None:
-            currency = currency.lower()
+        if currency or amount:
+            currency = currency.lower() if currency is not None else self.currency.lower()
             assert currency in ('lbc', 'btc', 'usd'), f'Unknown currency type: {currency}'
-            setattr(self, currency, Decimal(amount))
+            setattr(self, currency, Decimal(amount or self.amount))
```
eukreign: I think this can be cleaned up more. You check `currency` multiple times, including the inline. Are there really that many different cases? Maybe just check the good case and do `setattr` if everything needed is there, then have an `else` that handles the bad case. I think that might result in clearer code.

lbrynet/schema/claim.py
```diff
@@ -206,6 +206,7 @@ def to_dict(self):
 
     def update(self, file_path=None, height=None, width=None, duration=None, **kwargs):
         if kwargs.pop('clear_fee', False):
+            # clear_fee is set, ignore defaults
```
eukreign: This comment goes with code you removed, comment should be removed too.

Issue #1789, opened 2019-01-21
Support multi account awareness on all claim related calls


tzarebczan:
## The Issue
Currently, if a wallet has multiple accounts, the channel signatures are used from all available accounts. Other claim related calls (publish, claim_list_mine, channel_list, etc), should also be aware of claim data across all accounts. Not sure if this needs to be a special mode or not, but that's how it's working on spee.ch now and we have issues with updating claims because they may not be in the default wallet. 

My original idea is that we would have an inclusive/exclusive account wallet mode.  Inclusive would be for setups similar to spee.ch, where it's a single owner for all accounts and they are treated as one wallet essentially. Exclusive would be for LBRY on the Web, where each account has a distinct owner, and no channel/claim/tx information is ever shared.  

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

tzarebczan:
Closing for https://github.com/lbryio/lbry-sdk/issues/2264

PR #2220, opened 2019-06-10
Channel list shows channels across all accounts unless account id is passed

### LINKED ISSUE #1409: channel_export/channel_import

hackrush01:
fixes #1409, #1789

### DIFF
--- a/lbry/lbry/extras/daemon/Daemon.py
+++ b/lbry/lbry/extras/daemon/Daemon.py
@@ -1694,11 +1694,11 @@ class Daemon(metaclass=JSONRPCServerType):
 
         Returns: {Paginated[Output]}
         """
-        my_account = filtering_account = self.get_account_or_default(account_id)
+        my_account = self.get_account_or_default(account_id)
         return maybe_paginate(
             my_account.get_claims,
             my_account.get_claim_count,
-            page, page_size, account=[filtering_account]
+            page, page_size
         )
 
     @requires(WALLET_COMPONENT)
@@ -2145,38 +2145,31 @@ class Daemon(metaclass=JSONRPCServerType):
         return tx
 
     @requires(WALLET_COMPONENT)
-    def jsonrpc_channel_list(self, my_account_id=None, account_ids: list = None, page=None, page_size=None):
+    def jsonrpc_channel_list(self, my_account_ids=None, account_ids: list = None, page=None, page_size=None):
         """
         List my channel claims.
 
         Usage:
-            channel_list [<my_account_id> | --my_account_id=<my_account_id>]
+            channel_list [<my_account_ids> | --my_account_ids=<my_account_ids>]
                          [<account_ids> | --account_ids=<account_ids>]
                          [--page=<page>] [--page_size=<page_size>]
 
         Options:
-            --my_account_id=<my_account_id>  : (str, list) ids of my accounts
-            --account_ids=<account_ids>      : (str, list) ids of the account to use for filtering
-            --page=<page>                    : (int) page to return during paginating
-            --page_size=<page_size>          : (int) number of items on page during pagination
+            --my_account_ids=<my_account_ids>  : (str, list) ids of my accounts
+            --account_ids=<account_ids>        : (str, list) ids of the account to use for filtering
+            --page=<page>                      : (int) page to return during paginating
+            --page_size=<page_size>            : (int) number of items on page during pagination
 
         Returns: {Paginated[Output]}
         """
 
-        if my_account_id is None:
-            my_accounts = self.get_accounts_or_all(None)
-        else:
-            my_accounts = [self.get_account_or_error(my_account_id)]
-
-        if account_ids is None:
-            filtering_accounts = self.get_accounts_or_all(None)
-        else:
-            filtering_accounts = [self.get_account_or_error(account_id) for account_id in account_ids]
+        my_accounts = self.get_accounts_or_all(my_account_ids)
+        filtering_accounts = self.get_accounts_or_all(account_ids)
 
         return maybe_paginate(
             self.ledger.db.get_channels,
             self.ledger.db.get_channel_count,
-            page, page_size, account=filtering_accounts,
+            page, page_size, filtering_accounts=filtering_accounts,
             my_accounts=my_accounts
         )
 
@@ -2757,11 +2750,11 @@ class Daemon(metaclass=JSONRPCServerType):
 
         Returns: {Paginated[Output]}
         """
-        my_account = filtering_account = self.get_account_or_default(account_id)
+        my_account = self.get_account_or_default(account_id)
         return maybe_paginate(
             my_account.get_streams,
             my_account.get_stream_count,
-            page, page_size, account=[filtering_account]
+            page, page_size
         )
 
     @requires(WALLET_COMPONENT, EXCHANGE_RATE_MANAGER_COMPONENT, BLOB_COMPONENT,
@@ -2851,11 +2844,11 @@ class Daemon(metaclass=JSONRPCServerType):
 
         Returns: {Paginated[Output]}
         """
-        my_account = filtering_account = self.get_account_or_default(account_id)
+        my_account = self.get_account_or_default(account_id)
         return maybe_paginate(
             my_account.get_supports,
             my_account.get_support_count,
-            page, page_size, account=[filtering_account]
+            page, page_size
         )
 
     @requires(WALLET_COMPONENT, conditions=[WALLET_IS_UNLOCKED])
@@ -3024,11 +3017,11 @@ class Daemon(metaclass=JSONRPCServerType):
 
         Returns: {Paginated[Output]}
         """
-        my_account = filtering_account = self.get_account_or_default(account_id)
+        my_account = self.get_account_or_default(account_id)
         return maybe_paginate(
             my_account.get_utxos,
             my_account.get_utxo_count,
-            page, page_size, account=[filtering_account]
+            page, page_size
         )
 
     @requires(WALLET_COMPONENT)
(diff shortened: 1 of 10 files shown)


eukreign:
Looks like you're on the right track, the only concern I have so far is that you seem to be conflating `my_account` and `account` (you've renamed the variable in one case and you don't pass it to `get_txos()` anymore).

`my_account` is used to determine TXOs which are under your control as opposed to `accounts` which is used to determine which accounts you are filtering by. There is a subtle distinction there, make sure that as you refactor the code you are mindful to keep this part working.

As an example, if you have accounts A, B, C. And user is filtering on account B and some of the TXs include where something moved from account A -> B, the code has to mark the account A TXOs as "yours" as well. This gets set here:

https://github.com/lbryio/torba/blob/master/torba/client/basedatabase.py#L483

You may need to do an additional PR on torba which adds multi-account filtering/checking support on torba as well so that it checks `is_my_account` using a list of accounts instead of the current single `my_account`.

tzarebczan:
This would be nice to get in for 0.38 if it's close to being ready.

lbry/lbry/wallet/account.py
```diff
@@ -111,28 +111,28 @@ def get_utxo_count(self, **constraints):
...
-        return self.ledger.db.get_channel_count(account=self, **constraints)
+        return self.ledger.db.get_channel_count(my_accounts=[self], **constraints)
 
     def get_supports(self, **constraints):
-        return self.ledger.db.get_supports(account=self, **constraints)
+        return self.ledger.db.get_supports(my_accounts=[self], **constraints)
 
     def get_support_count(self, **constraints):
-        return self.ledger.db.get_support_count(account=self, **constraints)
+        return self.ledger.db.get_support_count(my_accounts=[self], **constraints)
```
eukreign: All of these above should be `accounts=[self]`

lbry/lbry/extras/daemon/Daemon.py
```diff
@@ -2138,26 +2138,32 @@ def jsonrpc_channel_new(self):
         return tx
 
     @requires(WALLET_COMPONENT)
-    def jsonrpc_channel_list(self, account_id=None, page=None, page_size=None):
+    def jsonrpc_channel_list(self, my_account_ids=None, account_ids: list = None, page=None, page_size=None):
```
eukreign: As discussed, should have no API changes.

lbry/lbry/extras/daemon/Daemon.py
```diff
@@ -3560,17 +3566,17 @@ def get_fee_address(self, kwargs: dict, claim_address: str) -> str:
...
+        channels = await self.ledger.db.get_channels(my_accounts=all_accounts, **{f'claim_{key}': value}, limit=1)
+        if len(channels) == 1:
+            if for_signing and channels[0].private_key is None:
+                raise Exception(f"Couldn't find private key for {key} '{value}'. ")
+            return channels[0]
+        elif len(channels) > 1:
+            raise ValueError(
+                f"Multiple channels found with channel_{key} '{value}', "
+                f"pass a channel_id to narrow it down."
+            )
```
eukreign: Did this method have to change? What's the functional difference between this and the previous version? Seems like previous version should still work.

lbry/lbry/wallet/database.py
```diff
@@ -52,26 +52,37 @@ def txo_to_row(self, tx, address, txo):
             row['claim_name'] = txo.claim_name
         return row
 
-    async def get_txos(self, **constraints) -> List[Output]:
-        my_account = constraints.get('my_account', constraints.get('account', None))
-
-        txos = await super().get_txos(**constraints)
+    async def get_txos(self, my_accounts, filtering_accounts=None, **constraints) -> List[Output]:
+        constraints.pop('account', None)
```
eukreign: Why is this just getting removed? Can it just not be passed in altogether if it's just going to be pop'ed off?

torba/torba/client/baseledger.py
```diff
@@ -422,8 +422,9 @@ def process_status_update(self, update):
                         check_db_for_txos.append(txi.txo_ref.id)
 
                 referenced_txos = {
-                    txo.id: txo for txo in await self.db.get_txos(txoid__in=check_db_for_txos)
-                }
+                    txo.id: txo for txo in await self.db.get_txos(
+                        my_accounts=None, txoid__in=check_db_for_txos
+                    )}
```
eukreign: `my_accounts` used to default to `None` in `get_txos()`, can it just go back to defaulting to `None`? Then this change will no longer be necessary.

torba/torba/client/basedatabase.py
```diff
@@ -480,14 +482,16 @@ def _transaction(conn: sqlite3.Connection, tx: BaseTransaction, address, txhash,
...
+            txo.is_my_account = row[7] in my_account_address
             txos.append(txo)
         return txos
 
     async def get_txo_count(self, **constraints):
         constraints.pop('offset', None)
         constraints.pop('limit', None)
         constraints.pop('order_by', None)
+        constraints.pop('my_accounts', None)
+        constraints['account'] = constraints.pop('filtering_accounts', constraints.pop('account', None))
```
eukreign: This looks wrong, is `account` a list or not? How can it hold both `filtering_accounts` which seems like a list and `account` which seems like a single account?

torba/torba/client/basedatabase.py
```diff
@@ -461,15 +461,17 @@ def _transaction(conn: sqlite3.Connection, tx: BaseTransaction, address, txhash,
...
+        if my_accounts is not None:
+            my_account_address = [my_account.public_key.address for my_account in my_accounts if
+                                  isinstance(my_account, BaseAccount)]
+
         if 'order_by' not in constraints:
             constraints['order_by'] = ["tx.height=0 DESC", "tx.height DESC", "tx.position DESC"]
         rows = await self.select_txos(
             "tx.txid, raw, tx.height, tx.position, tx.is_verified, txo.position, chain, account",
-            **constraints
+            account=filtering_accounts, **constraints
```
eukreign: account is singular, filtering_accounts is plural, this seems like it will be a problem?

lbry/lbry/extras/daemon/Daemon.py
hackrush01: This addresses the ToDo that you told me to complete yesterday

hackrush01: addressed*

Not sure if this will still work the new changes. Working on that.

hackrush01: https://github.com/lbryio/lbry-sdk/pull/2220/commits/42064a09cc63c0f357f641b5f725b86cd5e0dffa#diff-dd8ed5f4b40f4021a557728289646ad7L3569

lbry/lbry/wallet/database.py
hackrush01: This was there as fallback, in case some function is passing `account` which was not caught during the programming, I probably put in for debug. But can remove.

torba/torba/client/basedatabase.py
hackrush01: Some of the transaction list command can only use `account` as they have not yet been changed to work with my and filtering account. Hence this was needed

torba/torba/client/basedatabase.py
hackrush01: SQL only accepts account as a filter currently, so instead of changing the SQL I used this approach

torba/torba/client/basedatabase.py
hackrush01: Also, `[account]` is not a single account. But changing it to `[accounts]` breaks SQL. So I think the case is that "account" is the name of a column in an SQL file which is used for filtering.

hackrush01: Yes, looks like the `pubkey_address` table inside blockchain.db has a column names "account".

torba/torba/client/basedatabase.py
hackrush01: Reason same as above.

lbry/tests/integration/test_claim_commands.py
```diff
@@ -470,17 +470,17 @@ class ChannelCommands(CommandTestCase):
         account2_id, account2 = new_account['id'], self.daemon.get_account_or_error(new_account['id'])
 
         # before sending
-        self.assertEqual(len(await self.daemon.jsonrpc_channel_list()), 3)
```
eukreign: This should still be valid, you can call channel_list() and get 3, right? Why did this one get removed?

lbry/tests/integration/test_claim_commands.py
```diff
@@ -470,17 +470,17 @@ class ChannelCommands(CommandTestCase):
...
-        self.assertEqual(len(await self.daemon.jsonrpc_channel_list(account_id=account2_id)), 0)
+        self.assertEqual(len(await self.daemon.jsonrpc_channel_list(self.account.id)), 3)
+        self.assertEqual(len(await self.daemon.jsonrpc_channel_list(account2_id)), 0)
 
         other_address = await account2.receiving.get_or_create_usable_address()
         tx = await self.out(self.channel_update(claim_id, claim_address=other_address))
 
         # after sending
-        self.assertEqual(len(await self.daemon.jsonrpc_channel_list()), 2)
```
eukreign: This one is still valid, it'll just be 3 instead of 2.

lbry/tests/integration/test_claim_commands.py
```diff
@@ -617,7 +692,7 @@ class StreamCommands(ClaimTestCase):
         channels = await self.out(self.daemon.jsonrpc_channel_list(account1_id))
         self.assertEqual(len(channels), 1)
         self.assertEqual(channels[0]['name'], '@spam')
-        self.assertEqual(channels, await self.out(self.daemon.jsonrpc_channel_list()))
+        self.assertEqual(channels, await self.out(self.daemon.jsonrpc_channel_list(account1_id)))
```
eukreign: Does this test make sense to you? After the change it's just checking that `self.daemon.jsonrpc_channel_list(account1_id) == self.daemon.jsonrpc_channel_list(account1_id)`, is that helpful?

torba/torba/client/basedatabase.py
eukreign: Then other stuff needs to be fixed or refactored. It just doesn't make sense that you have a variable that's clearly one item `account` and you are setting it to a list of accounts `filtering_accounts`.

tzarebczan:
@hackrush01 do you have everything you need to try to finish this up for the next release?

lyoshenka:
fyi this is also blocking ytsync syncing channels that have been sent to creators, because the sdk cannot publish to a channel for which it no longer has the claim (even though it does have the channel cert)

eukreign:
superseded by #2386 which includes this functionality for all commands

PR #2225, opened 2019-06-11
[WIP] show status while downloading headers

shyba:
found a bug while doing this, work in progress


shyba:
#2268 covers that, closing

PR #2241, opened 2019-06-18
separate try/except for media server

lyoshenka:
This PR puts the media server port binding into a separate try/except so when it fails to bind the port, the logs correctly say what happened.


lyoshenka:
thanks @jessopb @EnigmaCurry for catching this

lbrynet/extras/daemon/Daemon.py
```diff
@@ -436,7 +440,7 @@ def ensure_download_dir(self):
             await self.stop()
         except Exception as e:
             await self.analytics_manager.send_server_startup_error(str(e))
-            log.exception('Failed to start lbrynet-daemon')
+            log.exception('Failed to start lbrynet')
```
eukreign: the other two raise/except raise systemexit, should this also raise systemexit?

PR #2249, opened 2019-06-19
add coin selection strategy config

lyoshenka:
Adds coin selection strategy config value so niko can upload

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,2 +1,4 @@
+# ignore everything
 *
+# except the .gitignore file
 !.gitignore
--- a/.idea/lbry-sdk.iml
+++ b/.idea/lbry-sdk.iml
@@ -2,9 +2,9 @@
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$">
-      <sourceFolder url="file://$MODULE_DIR$/torba/torba" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/torba" isTestSource="false" />
       <sourceFolder url="file://$MODULE_DIR$/torba/tests" isTestSource="false" />
-      <sourceFolder url="file://$MODULE_DIR$/lbry/lbry" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/lbry" isTestSource="false" />
       <sourceFolder url="file://$MODULE_DIR$/lbry/tests" isTestSource="false" />
       <sourceFolder url="file://$MODULE_DIR$/lbry/scripts" isTestSource="false" />
     </content>
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,12 +65,12 @@ jobs:
         - docker pull lbry/pyinstaller34_32bits:py371
       script:
         - pushd lbry && python scripts/set_build.py && popd
-        - docker run -v "$(pwd):/src/lbry" lbry/pyinstaller34_32bits:py371 lbry/scripts/wine_build.sh
-        - sudo zip -j dist/lbrynet-windows.zip dist/lbrynet.exe
+        - docker run -v "$(pwd):/src/lbry" lbry/pyinstaller34_32bits:py371 lbry/lbry/scripts/wine_build.sh
+        - sudo zip -j lbry/dist/lbrynet-windows.zip lbry/dist/lbrynet.exe
       deploy:
         provider: releases
         api_key: $GITHUB_OAUTH_TOKEN
-        file: dist/lbrynet-windows.zip
+        file: lbry/dist/lbrynet-windows.zip
         skip_cleanup: true
         overwrite: true
         draft: true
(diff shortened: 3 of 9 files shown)


tests/unit/test_conf.py
```diff
@@ -225,3 +226,22 @@ def test_max_key_fee_from_args(self):
...
+            default_none = OneOfString(["a"], "invalid default", None)
+        with self.assertRaises(ValueError):
+            invalid_default = OneOfString(["a"], "invalid default", "b")
+
+        valid_default = OneOfString(["a"], "valid default", "a")
+
+        c = TestConfig()
+        with self.assertRaises(ValueError):
+            c.one_of_string = "d"
+
```
jackrobison: can you add some non error case tests?

tests/unit/test_conf.py
```diff
@@ -225,3 +226,22 @@ def test_max_key_fee_from_args(self):
         args = parser.parse_args(['--max-key-fee', '1.0', 'BTC'])
         c = Config.create_from_arguments(args)
         self.assertEqual(c.max_key_fee, {'amount': 1.0, 'currency': 'BTC'})
+
+    def test_one_of_string(self):
+        parser = argparse.ArgumentParser()
+        Config.contribute_to_argparse(parser)
+
+        with self.assertRaises(ValueError):
+            no_vaid_values = OneOfString([], "no valid values", None)
```
jackrobison: these can just be like `self.assertRaises(ValueError, OneOfString, [], "no valid values", None)`

tests/unit/test_conf.py
```diff
@@ -225,3 +226,22 @@ def test_max_key_fee_from_args(self):
...
+
+        with self.assertRaises(ValueError):
+            default_none = OneOfString(["a"], "invalid default", None)
+        with self.assertRaises(ValueError):
+            invalid_default = OneOfString(["a"], "invalid default", "b")
+
+        valid_default = OneOfString(["a"], "valid default", "a")
```
jackrobison: this isn't used / the value isn't checked

lyoshenka: this is a non-error test case. it tests that if the default is in the valid values list, it does not raise

lbrynet/conf.py
```diff
@@ -192,6 +193,21 @@ def contribute_to_argparse(self, parser: ArgumentParser):
...
+        self.valid_values = valid_values
+        if not self.valid_values:
+            raise ValueError(f"No valid values provided")
+        if self.default not in self.valid_values:
+            raise ValueError(f"Default value must be one of: " + ', '.join(self.valid_values))
+
+    def validate(self, val):
+        super().validate(val)
+        if val not in self.valid_values:
+            raise ValueError(f"Setting '{self.name}' must be one of: " + ', '.join(self.valid_values))
```
eukreign: You can put expressions in f'strings:
```f"Setting '{self.name}' must be one of: {', '.join(self.valid_values)}"```

File: lbrynet/conf.py [hunk 5]
```diff
@@ -562,6 +578,9 @@ class Config(CLIConfig):
     streaming_get = Toggle("Enable the /get endpoint for the streaming media server. "
                            "Disable to prevent new streams from being added.", True)
 
+    coin_selection_strategy = OneOfString(STRATEGIES, "Strategy to use when selecting UTXOs for a transaction",
+                                          "branch_and_bound")
```
eukreign: This default would only run `branch_and_bound` and not the others. I think default should be `None` or `"auto"` which lets torba decide which strategy is best (currently doing the 3 strategies it does in specific order).

tests/unit/test_conf.py
eukreign: no, don't use context-less `self.assertRaises`, using `with self.assertRaises` context manager is the modern/better.
without context manager you don't get any type checking since you're passing arguments which go to `OneOfString` to `assertRaises` instead.

tests/unit/test_conf.py
eukreign: You've already "tested" the passing case by adding OneOfString to the main `TestConfig` and you'll actually use it to set some value below so that's enough I think.

tests/unit/test_conf.py
```diff
@@ -225,3 +226,32 @@ def test_max_key_fee_from_args(self):
...
+
+        self.assertEqual("hello", OneOfString(["hello"], "valid default", "hello").default)
```
eukreign: ^ would be better to do: `self.assertEqual("hello". TestConfig().one_of_string)` since that's testing how it's actually used everywhere

lyoshenka:
Thanks for the comments. I'll fix it up.

I had a really hard time figuring out how to test all this stuff because I had trouble following the logic. For example, I could not figure out where functions like `validate()` get called. From testing a bit using the command-line, it feels like they don't get called at all. For example, running `LBRY_DOWNLOAD_TIMEOUT=whatever lbrynet start` works just fine, even though `download_timeout` is supposed to be validated as an integer. On the other hand, `lbrynet start --download-timeout=whatever` does error, but it's not the expected `Setting must be a decimal`error, but instead a `could not convert string to float` error.

Maybe you can walk me through some of this stuff eventually. It's not urgent.

PR #2274, opened 2019-06-28
Adds venv targets to Makefile

osilkin98:
Adds targets for creating and cleaning a venv, along with installing the project inside it


tzarebczan:
This would only work on Linux/Mac, not Windows.

eukreign:
Just use `virtualenvwrapper` scripts (install via global/user `pip`), it does all of this for you; that's what I use. No reason to re-invent the wheel here.

I use it like this:
```
cd lbry-sdk
mkvirtualenv -a `pwd` -p python3 lbry
make install
```

If you follow the `virtualenvwrapper` installation instructions you should be able to do `workon lbry` anywhere and it'll take you to `lbry-sdk` directory and set you up with the correct virtual environment.

If find `virtualenvwrapper` indispensable.

https://virtualenvwrapper.readthedocs.io/en/latest/

lyoshenka:
Isn't it a lot easier for new devs to run `make install` (a very standard install path) then to go install virtualenvwrapper and then run an undocumented command? Seems like this PR is pretty useful for people who are not that familiar with all the python tools out there.

Alternatively, if we do want to require virtualenvwrapper, could you please add that to the readme?

eukreign:
The README includes instructions for someone who doesn't already have a preferred way of managing their virtual environment on how to get started and end up with working environment.

The current `make install` is generic and will work with any virtual environment management technique.

I don't think we want to prescribe a specific way, even on SDK team there are three different virtual environment managment tools used.

osilkin98:
Could just change the target to `virtualenv` so as to be verbose about what you’re actually using

PR #2287, opened 2019-07-02
use utils.aiohttp for comments

tzarebczan:
Fix certificate errors
Need to test locally still...


tzarebczan:
Confirmed this is working, merging.

PR #2297, opened 2019-07-08
let daemon be shut down before it has fully started

### LINKED ISSUE #2231: SDK does not shut down while failing to connect to wallet server
## The Issue
If you call lbrynet stop while the sdk is trying to connet/reconnect to a wallet server (i.e. lack of connectivity, can't reach), the daemon won't actually stop. Ctrl-C works though.

To test: set your wallet server to an invalid IP/port. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

lyoshenka:
There is more that can be improved here. Starting and stopping the daemon is still somewhat inconsistent. The daemon can shut down by raising SystemExit, or it shut down by raising GracefulExit. Is that right, or should there only be one way to do it?

fixes #2231

### DIFF
--- a/lbry/lbry/conf.py
+++ b/lbry/lbry/conf.py
@@ -362,6 +362,9 @@ class ConfigFileAccess:
         del self.data[key]
 
 
+TBC = typing.TypeVar('TBC', bound='BaseConfig')
+
+
 class BaseConfig:
 
     config = Path("Path to configuration file.", metavar='FILE')
@@ -418,7 +421,7 @@ class BaseConfig:
         }
 
     @classmethod
-    def create_from_arguments(cls, args):
+    def create_from_arguments(cls, args) -> TBC:
         conf = cls()
         conf.set_arguments(args)
         conf.set_environment()
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -221,6 +221,61 @@ def ensure_directory_exists(path: str):
         pathlib.Path(path).mkdir(parents=True, exist_ok=True)
 
 
+def run_daemon(args: list, conf: Config):
+    default_formatter = logging.Formatter("%(asctime)s %(levelname)-8s %(name)s:%(lineno)d: %(message)s")
+    file_handler = logging.handlers.RotatingFileHandler(
+        conf.log_file_path, maxBytes=2097152, backupCount=5
+    )
+    file_handler.setFormatter(default_formatter)
+    log.addHandler(file_handler)
+    logging.getLogger('torba').addHandler(file_handler)
+
+    if not args.quiet:
+        handler = logging.StreamHandler()
+        handler.setFormatter(default_formatter)
+        log.addHandler(handler)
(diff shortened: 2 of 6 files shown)


lbry/lbry/conf.py
```diff
@@ -362,6 +362,9 @@ def __delitem__(self, key):
         del self.data[key]
 
 
+TBC = typing.TypeVar('TBC', bound='BaseConfig')
```
eukreign: What's this for?

lyoshenka: this is how the internet told me to do typing for a @classmethod

eukreign: doesn't seem right, especially for a `@classmethod`, the `cls` variable points to the class on which the class method appears (it's bound to it)

I think you'd only need to do this if you wanted to pass the Config to a function somewhere else, but you didn't want to bind it to `BaseConfig` or to the specific sublcass of `BaseConfig`. I think everywhere in LBRY it should require the specific `Config` object that describes LBRY configs, so you can just use that directly without needing a `TypeVar`.

see here for correct usage:
https://www.python.org/dev/peps/pep-0484/#type-variables-with-an-upper-bound

lyoshenka: a) how can I say the function returns a `Config` object when I'm in the `BaseConfig` class? I don't really understand that part.
b) if everything uses Config and not BaseConfig, why have both classes? can we merge the two?

btw here's the link I got that from: https://github.com/python/typing/issues/58#issuecomment-326240794. It's endorsed by gvanrossum :wink:

eukreign: I got the opposite impression from what Guido says in that thread:
"We're not going to support this. The issue seems fairly academic."

a) I think for the "Factory" pattern specifically, you will need the `TypeVar` after all, but in the PR you have it marked on the `cls` variable too, that's what I was mostly objecting to. The `cls` variable already has the correct type info, if you hover over it in PyCharm it should show correct type.

b) Makes it easier to test in isolation and also if we wanted to have another Config type, for example, if we wanted to re-use this code in the wallet server (which currently has its own crappy config management implementation).

So, in summary, I would only have the `TBC` set as return type of `def create_from_arguments(cls, args) -> TBC:` and not on the `cls` variables since those are typed correctly already (also, you can't call those class methods and pass the `cls` as an argument, the `@classmethod` takes care of that, adding the type info to `cls` is just noise and doesn't affect the API or internal operation of `@classmethod`).

PR #2305, opened 2019-07-09
Fix startup error if download dir not set on Linux

### LINKED ISSUE #2291: Startup error in get_linux_directories
When starting LBRY-0.33.2 the first time on Ubuntu 19.04, I get the following message during startup:

> The daemon may have encountered an unexpected error, or another daemon instance is already running.

The log file does not exist, but the console has the following output:

    Starting SDK
    Sandbox listening on port 5278.
    Daemon: [17050] Failed to execute script cli
    
    Daemon: Traceback (most recent call last):
      File "lbrynet/extras/cli.py", line 322, in <module>
      File "lbrynet/extras/cli.py", line 229, in main
      File "lbrynet/conf.py", line 400, in create_from_arguments
      File "lbrynet/conf.py", line 570, in __init__
      File "lbrynet/conf.py", line 582, in set_default_paths
      File "lbrynet/conf.py", line 621, in get_linux_directories
    AttributeError: 'NoneType' object has no attribute 'group'

I tried manually creating the `~/.config/LBRY/user-dirs.dirs` according to the code that tries to read it, but the exception is still the same.

shauser:
Fixes #2291

### DIFF
--- a/lbry/lbry/conf.py
+++ b/lbry/lbry/conf.py
@@ -644,15 +644,18 @@ def get_darwin_directories() -> typing.Tuple[str, str, str]:
 
 
 def get_linux_directories() -> typing.Tuple[str, str, str]:
+    download_dir = None
     try:
         with open(os.path.join(user_config_dir(), 'user-dirs.dirs'), 'r') as xdg:
-            down_dir = re.search(r'XDG_DOWNLOAD_DIR=(.+)', xdg.read()).group(1)
-        down_dir = re.sub('\$HOME', os.getenv('HOME') or os.path.expanduser("~/"), down_dir)
-        download_dir = re.sub('\"', '', down_dir)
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for the PR @shauser , congrats on the first contribution. We'll get a review on this. 

Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the PR?

tzarebczan:
@eukreign can this get a review please?

eukreign:
@shauser how do i reproduce this? under what circumstances would `user-dirs.dirs` file not be there?

if the `XDG_DOWNLOAD_DIR` has the same value, why do we even check the file?

would be interesting if someone could look into some background on this issue, i don't want to just apply this fix without understanding what problem it's fixing

lyoshenka:
I'm going to close this because we haven't heard back from @shauser. Stephan, we would love it if you came back to finish this. Please reopen when you do.

PR #2310, opened 2019-07-15
set file status to completed only when all bytes were written

### LINKED ISSUE #2293: completed status and blob completed incorrect in streaming mode
streaming `what` on a fresh install with save files/blobs off - completed shows true, and blobs completed shows a value. No blobs exist on disk. 

This is pretty bad for the desktop app streaming solution because we only want to show completed files when they are actually completed - the file is saved, or all the blobs are saved (on disk too). 
```
(lbry-venv) C:\Users\thoma\Documents\lbry>lbrynet file list --claim_name=what
[
  {
    "blobs_completed": 2,
    "blobs_in_stream": 76,
    "blobs_remaining": 74,
    "channel_claim_id": null,
    "channel_name": null,
    "claim_id": "19b9c243bea0c45175e6a6027911abbad53e983e",
    "claim_name": "what",
    "completed": true,
    "confirmations": 208412,
```

shyba:
closes #2293

### DIFF
--- a/lbry/lbry/blob/blob_manager.py
+++ b/lbry/lbry/blob/blob_manager.py
@@ -69,10 +69,10 @@ class BlobManager:
     def is_blob_verified(self, blob_hash: str, length: typing.Optional[int] = None) -> bool:
         if not is_valid_blobhash(blob_hash):
             raise ValueError(blob_hash)
-        if blob_hash in self.blobs:
-            return self.blobs[blob_hash].get_is_verified()
         if not os.path.isfile(os.path.join(self.blob_dir, blob_hash)):
             return False
+        if blob_hash in self.blobs:
+            return self.blobs[blob_hash].get_is_verified()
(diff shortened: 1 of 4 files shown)


tzarebczan:
Looks correct for status, but the other bug I mentioned was that the blobs show completed but they actually aren't on disk. If that's not a quick fix, no worries, we can keep that issue open separately. 

```(lbry-venv) C:\Users\thoma\Documents\lbry>lbrynet file list --claim_name=what
[
  {
    "blobs_completed": 1,
    "blobs_in_stream": 76,
    "blobs_remaining": 75,
    "channel_claim_id": null,
    "channel_name": null,
    "claim_id": "19b9c243bea0c45175e6a6027911abbad53e983e",
    "claim_name": "what",
    "completed": false,
```

tzarebczan:
This LGTM!

PR #2326, opened 2019-07-17
return partial fetched results from interrupted queries

jackrobison:


tzarebczan:
Are we still doing this? Can we close it?

PR #2329, opened 2019-07-18
Fix effective amount calculation

tzarebczan:
Fixes https://github.com/lbryio/lbry-sdk/issues/2318


tzarebczan:
Closing pending further conversations. I will try to submit a PR on the reader side to use support + bid for effective amount in the meantime

PR #2332, opened 2019-07-21
Adds `comment abandon` command

osilkin98:
## Changes
Adds the command `comment_abandon` with the following signature:
```python
def comment_abandon(comment_id: str) -> dict:
    return {
        comment_id: {
            "deleted": bool
        }
    }
```

## Functionality
Accepts a `comment_id` and attempts to query the channel that the comment belongs to. 
If the user owns the channel, a signature is then made with the following encoding:
```
`comment_id` + `channel_claim_hash`
```
This signature is then validated by the comment server, returning the outcome of the request accordingly. 

###### TODO:
- Add unit tests

### DIFF
--- a/lbry/lbry/extras/daemon/Daemon.py
+++ b/lbry/lbry/extras/daemon/Daemon.py
@@ -3497,40 +3497,33 @@ class Daemon(metaclass=JSONRPCServerType):
             response['is_claim_signature_valid'] = is_comment_signed_by_channel(response, channel)
         return response
 
-    async def jsonrpc_comment_abandon(self, comment_id, channel_name=None, channel_id=None, channel_account_id=None):
+    async def jsonrpc_comment_abandon(self, comment_id):
         """"
         Delete a comment published under your channel identity
 
         Usage:
             comment_delete  (<comment_id> | --comment_id=<comment_id>)
-                            (--channel_id=<channel_id> | --channel_name=<channel_name>)
-                            [--channel_account_id=<channel_account_id>...]
+
 
         Options:
             --comment_id=<comment_id>                   : (str) The ID of the comment to be deleted.
-            --channel_id=<channel_id>                   : (str) The ID of the channel that posted the comment.
-            --channel_name=<channel_name>               : (str) The Name of the channel that posted the comment..
-            --channel_account_id=<channel_account_id>   : (str) one or more account ids for accounts to look in
-                                                          for channel certificates, defaults to all accounts.
+
         Returns:
         """
         abandon_comment_body = {'comment_id': comment_id}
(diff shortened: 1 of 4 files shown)


lbry/lbry/extras/daemon/Daemon.py
```diff
@@ -33,7 +33,8 @@
 from lbry.extras.daemon.ComponentManager import RequiredCondition
 from lbry.extras.daemon.ComponentManager import ComponentManager
 from lbry.extras.daemon.json_response_encoder import JSONResponseEncoder
-from lbry.extras.daemon.comment_client import jsonrpc_post, sign_comment, is_comment_signed_by_channel
+from lbry.extras.daemon.comment_client import jsonrpc_post, sign_comment, sign_abandon_comment
+from lbry.extras.daemon.comment_client import is_comment_signed_by_channel
```
eukreign: either just import `comment_client` as a module and use dot notation to access the functions later in the code:
```
from lbry.extras.daemon import comment_client
```
or split the import list across several lines:
```
from lbry.extras.daemon.comment_client import (
    jsonrpc_post, sign_comment, sign_abandon_comment, is_comment_signed_by_channel
)
```

lbry/lbry/extras/daemon/comment_client.py
```diff
@@ -41,17 +41,30 @@ def sign_comment(comment: dict, channel: Output):
...
 
 
-async def jsonrpc_post(url: str, method: str, **params) -> any:
+def sign_abandon_comment(body: dict, channel: Output):
+    pieces = [body['comment_id'].encode(), channel.claim_hash]
+    digest = sha256(b''.join(pieces))
+    signature = channel.private_key.sign_digest_deterministic(digest, hashfunc=hashlib.sha256)
+    body.update({
+        'signature': binascii.hexlify(signature).decode()
+    })
```
eukreign: Is it possible to refactor this so that creating a comment and abandoning a comment use the same signing function? It seems like it should be possible to make this API symmetrical.

lbry/lbry/extras/daemon/comment_client.py
```diff
@@ -41,17 +41,30 @@ def sign_comment(comment: dict, channel: Output):
...
+
+
+async def jsonrpc_post(url: str, method: str, params: dict = None, **kwargs) -> any:
+    params = dict() if not params else params
```
eukreign:
```
params = params or {}
```

lbry/lbry/extras/daemon/comment_client.py
osilkin98: done

PR #2333, opened 2019-07-22
index common tags during block updates

jackrobison:
-remove hard coded common tags
-don't strip apostrophes from tags


tzarebczan:
What's the status of this PR? Been sitting out here for a while.

eukreign:
closing this for now, one of the major issues with this is that each index adds ~100MB per tag in storage requirement. we're going to manually maintain the list of tags that need indexing to keep this in check.

PR #2350, opened 2019-07-26
add granular balances

shyba:
backwards-incompatible: `account balance` no longer returns an integer. It now returns a dictionary with granular balances. The integer it used to return is now under `available` key on that new dictionary format.


tzarebczan:
Tested this on Windows + spee.ch wallet and it took  ~5 seconds to run vs 1 second previously. The app polls every 2 seconds, so would  not work with the app. The spee.ch wallet is an extreme case, but it shows that this does take a bunch more resources to run. My recommendation was to not show any reserved data (only show balance), unless the param is passed (then it would do reserved + breakdown). 

Websockets would help with the app polling issue.

PR #2383, opened 2019-08-11
small fix to annoying formatting

osilkin98:


eukreign:
this PR fails to take into account operator precedence:
```
>>> "col " + "ASC" if True else "DESC"
'col ASC'
>>> "col " + "ASC" if False else "DESC"
'DESC'
```

PR #2384, opened 2019-08-11
Faster file list

shyba:
check blob hashes from completed_blob_hashes instead of heavier function calls


shyba:
that's a big slowdown, maybe not the only one but def one

Issue #2264, opened 2019-06-25
Treat multiple accounts as one for multi-device / wallet scenarios


tzarebczan:
Building off of https://github.com/lbryio/lbry-sdk/issues/1789, but this would include the ability to treat a wallet as inclusive, and return all channels/transactions/balances summed up across all accounts. 

To support cross device syncing for existing users on multiple devices, one approach would be to allow multiple accounts to be treated as one from an account based perspective. The other approach would be a way to allow them to merge one account into another by sending all transactions/claim/tips/certificates. 

Not sure we want to support this yet, but with the increased influx of users, there are tons more in this boat. This would also apply to users who used different emails and then later merge their accounts (not sure how this would work yet).  The other option would not to allow them to combine accounts and give them explicit instructions on how to do so (most people will not know how to follow / this won't be easy if there's publishing/channels/etc across the 2 devices). 

If we allow syncing on lbry.tv, any APIs with the account id param would need to support multiple account ids.

tzarebczan:
The app team will be adding wallet sync on Desktop soon so we'll need users to be able to see transactions/content/etc across multiple accounts.

tzarebczan:
Missing account send support. This would include publishes across multiple accounts utxos and other claim/support related activities.

tzarebczan:
We also need to support other account-related commands like encrypt/decrypt/lock/unlock, right? If I have a wallet with 2 encrypted accounts with the same password, it should unlock both. Same would go for the other commands.

lyoshenka:
is encryption done at the account level? I thought it encrypts your whole wallet. it seems better to do it that way - simpler, and it will match how sync works.

tzarebczan:
Today it is, but with us going more towards the wallet level route for lbry.tv and other wallets, it makes sense to as you said. See https://github.com/lbryio/lbry-sdk/issues/2382 for a discussion around scenarios for current users / passwords.

tzarebczan:
Besides account_send, we also need to support stream abandon and channel abandon, and any other API calls dealing with transactions/claims.

lyoshenka:
the only thing left is balance and abandon rpc methods

eukreign:
balance and abandon work across accounts

PR #2386, opened 2019-08-12
most commands now work across all accounts by default

eukreign:
backwards-incompatible: `channel_create`, `channel_update`, `publish`, `stream_create`, `stream_update` and `support_create` have a new `--funding_account_ids` argument with a new default behavior that all accounts are used to fund those transactions and you must explicitly limit accounts by using the `--funding_account_ids` argument.
backwards-incompatible: `claim_list`, `channel_list`, `stream_list`, `support_list`, `transaction_list` and `utxo_list` now list results across all accounts, you mass pass `account_id` argument to limit results to a specific account.


eukreign:
closes #2264, #2220

PR #2388, opened 2019-08-12
Throttle concurrent blob downloads

jackrobison:
pause other blob exchange transports for a given blob during a download once one of them has written 64k bytes. If this download subsequently fails, resume the other transports.


lyoshenka:
uTP will handle this

PR #2419, opened 2019-08-26
Adds spend_address param to support_abandon

osilkin98:
Allows you to specify a lbry wallet address to `support abandon` which receives the LBC released from the support transaction.


lyoshenka:
@osilkin98 what's the next step here? does it need a review?

tzarebczan:
Lex is aware of this one and wanted to do a deeper use case review.

osilkin98:
Alright, I'll leave this one open just in case.

lyoshenka:
Lex and Oleg discussed offline. Instead of adding this param, a better approach would be to add a new rpc call for the use case of sending a channel with a bunch of claims/supports to another address. `spend_address` has some corner cases that are hard to handle without that.

PR #2427, opened 2019-08-28
added pagination to `file_list`

### LINKED ISSUE #1650: Add pagination to file_list RPC command
`file_list` results should be paginated, just like other `*_list` RPC commands. 

Might be blocked by https://github.com/lbryio/lbry/issues/1228.

osilkin98:
fixes #1650


lbry/lbry/extras/daemon/Daemon.py
```diff
@@ -1544,15 +1545,32 @@ def jsonrpc_file_list(self, sort=None, reverse=False, comparison=None, **kwargs)
...
+            offset = page_size * (page-1)
+            return {
+                'total_items': total_items,
+                'total_pages': int((total_items + (page_size-1)) / page_size),
+                'page': page,
+                'page_size': page_size,
+                'items': file_list[offset:offset+page_size]
+            }
+
+        return file_list
```
eukreign: The method should return the same return structure (either `Paginated` or `List` but not both).

osilkin98: I fixed by defaulting `page` and `page_size` to `1` and `10` respectively.

osilkin98:
I cannot add it myself but this PR needs the `backwards-incompatible` tag on it.

tzarebczan:
FYI: We have some inconsistencies for pagination across other apis - i.e. transaction/support list will show all transactions, and not paginate by default, but claim search does paginate.

tzarebczan:
Lex also mentioned that we aren't using the same pagination process as other APIs - is that possible here or it requires deeper changes that @shyba was looking into?

osilkin98:
@tzarebczan Should it use pagination only when `page` and `page_size` are passed in, or just either one (with the other defaulted)? 

> Lex also mentioned that we aren't using the same pagination process as other APIs - is that possible here or it requires deeper changes that @shyba was looking into?

This will have to be done in a separate PR.

tzarebczan:
I'd say only use it when it's passed - this would be backward compatible and we can make larger breaking changes when we reformat the entire output to use claim data.

tzarebczan:
Not sure if adding pagination before doing that / doing it the right way is useful though...

osilkin98:
@tzarebczan Good call, existing code won't need to be refactored as done in later commits.

jackrobison:
I don't think we should introduce multiple response formats that are contingent on the args provided to an api call - this will only add handling code for clients of the api for the different cases and push back the BC break a little further. If pagination is to become the default, then it should always return the dict of paginated results. Updating to handle paginated results should be fairly simple for api clients - this smaller BC break shouldn't be blocked by future changes to claims within file_list.

lyoshenka:
Feels like the solution is to pretend they requested the first page if they did not include any pagination options. Its ok to break BC here.

lyoshenka:
@osilkin98 do you know what to do next? do you want to talk to jack or lex?

osilkin98:
> Feels like the solution is to pretend they requested the first page
Should the `page_size` be 10 or just the entire thing?

osilkin98:
What are we thinking with this?

osilkin98:
Because they were easier to regex than using brackets.

eukreign:
@osilkin98 using direct dictionary access is better because it fails fast if the key is not there instead of returning None, also it's inconsistent with the rest of the code base which uses plain Python dictionary access except in specific circumstances where it's an expected case that the key might not be present.

can you please change the `.get()` to `[]`

osilkin98:
Alright, give me a bit.

PR #2446, opened 2019-09-05
Strips comment before signing

osilkin98:
Strips whitespace off the comment *before* it gets sent to the server.


lyoshenka:
@osilkin98 please rebase and then merge

Issue #2050, opened 2019-04-24
Cannot set max_key_full to null via cli


tzarebczan:
## The Issue
You can set the max_key_fee to null via json rpc or starting the sdk with `--disable_max_key_fee`, but it's not possible to do so via the cli:

```
C:\Users\thoma\Desktop>lbrynet settings set max_key_fee null
"Max key fee is made up of two values: \"AMOUNT CURRENCY\"."

C:\Users\thoma\Desktop>lbrynet settings set max_key_fee ''
"Max key fee is made up of two values: \"AMOUNT CURRENCY\"."
```

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

tzarebczan:
Fixed in #2450

PR #2450, opened 2019-09-08
Enable to set max_key_fee to null via cli

mirgee:
This fix is meant to address #2050. Instead letting argparse check the number of arguments, the arguments are cast to a list whose length is checked in `_parse_list()`. It was tested via both CLI and yaml, and I extended some of the unit tests in `test_conf.py`.

### DIFF
--- a/lbry/lbry/conf.py
+++ b/lbry/lbry/conf.py
@@ -148,7 +148,10 @@ class MaxKeyFee(Setting[dict]):
 
     @staticmethod
     def _parse_list(l):
-        assert len(l) == 2, 'Max key fee is made up of two values: "AMOUNT CURRENCY".'
+        if len(l) == 1 and (l[0] == 'null' or not l[0]):
+            return None
+        assert len(l) == 2, ('Max key fee is made up of either two values: '
+                             '"AMOUNT CURRENCY", or "null" (to set no limit)')
         try:
             amount = float(l[0])
         except ValueError:
@@ -159,8 +162,8 @@ class MaxKeyFee(Setting[dict]):
         return {'amount': amount, 'currency': currency}
 
     def deserialize(self, value):
-        if value is None:
-            return
+        if not value:
+            return None
         if isinstance(value, dict):
             return {
                 'currency': value['currency'],
@@ -176,7 +179,7 @@ class MaxKeyFee(Setting[dict]):
         parser.add_argument(
             self.cli_name,
             help=self.doc,
-            nargs=2,
+            nargs='+',
             metavar=('AMOUNT', 'CURRENCY'),
             default=NOT_SET
         )
@@ -527,7 +530,8 @@ class Config(CLIConfig):
         "peers are not found or are slow", 2.0
     )
     max_key_fee = MaxKeyFee(
-        "Don't download streams with fees exceeding this amount", {'currency': 'USD', 'amount': 50.0}
+        "Don't download streams with fees exceeding this amount. When set to "
+        "null, the amount is unbounded.", {'currency': 'USD', 'amount': 50.0}
     )
 
     # reflector settings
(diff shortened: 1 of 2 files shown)


tzarebczan:
Thanks for the PR! We'll get a review on this soon. Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the contribution?

lbry/lbry/conf.py
```diff
@@ -148,7 +148,10 @@ def validate(self, value):
 
     @staticmethod
     def _parse_list(l):
-        assert len(l) == 2, 'Max key fee is made up of two values: "AMOUNT CURRENCY".'
+        if len(l) == 1 and (l[0] == 'null' or not l[0]):
+            return None
+        assert len(l) == 2, ('Max key fee is made up of either two values: '
+                             '"AMOUNT CURRENCY", or "null" (to set no limit)')
```
eukreign:
```
        assert len(l) == 2, \
            'Max key fee is made up of either two values: "AMOUNT CURRENCY", or "null" (to set no limit)'
```
or
```
        assert len(l) == 2, (
            'Max key fee is made up of either two values: '
            '"AMOUNT CURRENCY", or "null" (to set no limit)'
        )
```
is the format we normally use.

lbry/lbry/conf.py
```diff
@@ -159,8 +162,8 @@ def _parse_list(l):
         return {'amount': amount, 'currency': currency}
 
     def deserialize(self, value):
-        if value is None:
-            return
+        if not value:
+            return None
```
eukreign: Why this change?

lbry/tests/unit/test_conf.py
```diff
@@ -222,6 +227,14 @@ def test_max_key_fee_from_args(self):
         c = Config.create_from_arguments(args)
         self.assertEqual(c.max_key_fee, None)
 
+        args = parser.parse_args(['--max-key-fee', 'null'])
+        c = Config.create_from_arguments(args)
+        self.assertEqual(c.max_key_fee, None)
+
+        args = parser.parse_args(['--max-key-fee', ''])
+        c = Config.create_from_arguments(args)
+        self.assertEqual(c.max_key_fee, None)
```
eukreign: I think `null` as an argument is enough, supporting a blank is ambiguous and might have unintended edge cases.

lbry/lbry/conf.py
mirgee: To support the empty string case, which we don't want to do, see below; I'll revert it.

lbry/lbry/conf.py
```diff
@@ -148,7 +148,12 @@ def validate(self, value):
 
     @staticmethod
     def _parse_list(l):
-        assert len(l) == 2, 'Max key fee is made up of two values: "AMOUNT CURRENCY".'
+        if len(l) == 1 and l[0] == 'null':
```
eukreign: I think this could be simplified to:
```
if l == ['null']:
```

lbry/tests/unit/test_conf.py
```diff
@@ -207,6 +207,11 @@ def test_max_key_fee_from_yaml(self):
                 c.max_key_fee = None
             with open(config, 'r') as fd:
                 self.assertEqual(fd.read(), 'max_key_fee: null\n')
+            c = Config.create_from_arguments(
+                types.SimpleNamespace(config=config)
+            )
+            with open(config, 'r') as fd:
+                self.assertEqual(c.max_key_fee, None)
```
eukreign: Can you explain what's going on here? I don't understand what this is testing. (why the new `c` config instance and why the assert for None, also why that assert happens in a file context manager but the `fd` descriptor isn't used)

mirgee: This is trying to test `max_key_fee` is set to `None` that when config is created from `settings.yaml` where `max_key_fee: null\n` (the context manager shouldn't be there). It doesn't test my changes (or our code, for that matter), so I will remove it.

PR #2452, opened 2019-09-10
Fix sqlite misuse

jackrobison:
fixes https://github.com/lbryio/lbry-sdk/issues/2164


torba/torba/client/basedatabase.py
```diff
@@ -583,10 +584,12 @@ def get_utxo_count(self, **constraints):
 
     async def add_keys(self, account, chain, keys):
         await self.db.executemany(
-            "insert into pubkey_address (address, account, chain, position, pubkey) values (?, ?, ?, ?, ?)",
-            ((pubkey.address, account.public_key.address, chain,
-              position, sqlite3.Binary(pubkey.pubkey_bytes))
-             for position, pubkey in keys)
+            "insert into pubkey_address values (?, ?, ?, ?, ?, NULL, 0)",
```
eukreign: It's good to leave the list of columns, that way if the table structure changes in any way this code will fail quickly, without the list of columns and depending on the changes in the data structure, this could might work but save things to wrong columns.

tzarebczan:
Confirmed this fixes the issue in the wallet I ran into it with previously (https://github.com/lbryio/lbry-sdk/issues/2164)

Issue #2027, opened 2019-04-12
--config argument silently falls back to defaults when the file does not exist


jackrobison:
This should log a warning if not prevent startup.

mirgee:
@jackrobison Also, since the config path doesn't change, calling `lbrynet settings set` fails.

I'd vote for failing if the path is invalid or the directory is not writeable, and write any changes the client makes to that directory. Then, the config path can be taken care of as a part of #877. What do you think?
 
We can be 'friendly' and e.g. allow the user to choose their own path to the config file and if doesn't exist, warn them, fall back to defaults, and write any change they make to the specified path (if it exists and is writeable); but I think failing immediately when there is a chance of unexpected behavior is much safer.

PR #2454, opened 2019-09-11
Log inaccessible downloads / wallet path, switch to default if possible

### LINKED ISSUE #877: Daemon fails to start if it cannot write/access download directory
### Steps to Reproduce:

- Stop daemon
- Change the permissions on the download directory to 000 (e.g. `chmod 000 ~/Downloads`)
- Start daemon

### Error Trace:

```
2017-09-22 18:18:58,200 INFO     lbrynet.daemon.Daemon:471: Loading databases
2017-09-22 18:18:58,205 ERROR    lbrynet.daemon.DaemonControl:100: Failed to start lbrynet-daemon
  File "site-packages/lbrynet/daemon/DaemonControl.py", line 97, in start_server_and_listen
  File "site-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
  File "site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages/lbrynet/daemon/DaemonServer.py", line 53, in start
  File "site-packages/twisted/internet/defer.py", line 1297, in _inlineCallbacks
  File "site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
  File "site-packages/lbrynet/daemon/Daemon.py", line 251, in setup
  File "site-packages/twisted/python/threadpool.py", line 246, in inContext
  File "site-packages/twisted/python/threadpool.py", line 262, in <lambda>
  File "site-packages/twisted/python/context.py", line 118, in callWithContext
  File "site-packages/twisted/python/context.py", line 81, in callWithContext
  File "site-packages/lbrynet/daemon/Daemon.py", line 473, in _setup_data_directory
OSError: [Errno 13] Permission denied: '/home/miroslav/Videos/lbry'
2017-09-22 18:18:58,210 INFO     lbrynet.daemon.DaemonServer:57: Stopping the reactor
```

_Copied from original issue: lbryio/lbry-app#495_

mirgee:
fixes #877. If any of download or wallet directories are not writable, then an attempt to set corresponding path to the default is made, and if unsuccessful, the app is shut down. Either case is logged into the log file. This was tested manually, and original unit tests seem to be passing.

In accordance with Pythonic EAFP, to test if directory is writable, I just write to it. It does not look pretty, but it has the benefit that it all the other cases, e.g. there is not enough space on the disk. Just checking the permissions might look nicer.

Should I also cover the case that even the log file is not accessible and logs should be just sent to stdout? Or is it OK to just throw an exception and not start?

Please, let me know if this is desired behavior, and if this is even a step in good direction before I start writing tests.

### DIFF
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -276,14 +276,15 @@ def run_daemon(args: argparse.Namespace, conf: Config):
     if hasattr(loop, 'shutdown_asyncgens'):
         loop.run_until_complete(loop.shutdown_asyncgens())
 
+
 def prepare_directory(dir_path: str):
     if not os.path.isdir(dir_path):
         pathlib.Path(dir_path).mkdir(parents=True, exist_ok=True)
-    testfile = tempfile.TemporaryFile(dir=dir_path)
-    testfile.close()
+    tempfile.TemporaryFile(dir=dir_path).close()
+    return dir_path
 
 
-def change_to_default_if_inaccessible(parsed_path: str, path_obj: Path, conf: Config):
+def change_to_default_if_inaccessible(parsed_path: str, path_obj: Path):
     assert path_obj.metavar in ('DIR', 'FILE'), "Path must be either to a directory or a file"
     handled_errors = (OSError, IOError, PermissionError)
     is_file = path_obj.metavar == 'FILE'
@@ -291,15 +292,14 @@ def change_to_default_if_inaccessible(parsed_path: str, path_obj: Path, conf: Co
     dir_path = os.path.basename(parsed_path) if is_file else parsed_path
     default_dir_path = os.path.basename(default_path) if is_file else default_path
     try:
-        prepare_directory(dir_path)
+        return prepare_directory(dir_path)
     except handled_errors:
         log.warning("Failed to access %s, setting to %s", dir_path, default_dir_path)
-        setattr(conf, path_obj.name, default_dir_path)
-        try:
-            prepare_directory(default_dir_path)
-        except handled_errors as e:
-            log.critical("Failed to access both %s and %s", dir_path, default_dir_path)
-            raise e
+    try:
+        return prepare_directory(default_dir_path)
+    except handled_errors:
+        log.critical("Failed to access both %s and %s", dir_path, default_dir_path)
+        raise
 
 
 def main(argv=None):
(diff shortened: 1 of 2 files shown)


mirgee:
This fix also potentially takes care of #2027.

lbry/lbry/extras/cli.py
```diff
@@ -276,14 +249,69 @@ def __exit():
...
+    if conf.share_usage_data:
+        loggly_handler = get_loggly_handler()
+        loggly_handler.setLevel(logging.ERROR)
+        log.addHandler(loggly_handler)
+
+
+def prepare_directory(dir_path: str):
+    if not os.path.isdir(dir_path):
+        pathlib.Path(dir_path).mkdir(parents=True, exist_ok=True)
+    testfile = tempfile.TemporaryFile(dir=dir_path)
```
jackrobison: this can be reduced down to just `tempfile.TemporaryFile(dir=dir_path).close()`

lbry/lbry/extras/cli.py
```diff
@@ -276,14 +249,69 @@ def __exit():
...
 def main(argv=None):
     argv = argv or sys.argv[1:]
     parser = get_argument_parser()
     args, command_args = parser.parse_known_args(argv)
 
     conf = Config.create_from_arguments(args)
-    for directory in (conf.data_dir, conf.download_dir, conf.wallet_dir):
-        ensure_directory_exists(directory)
+
+    change_to_default_if_inaccessible(conf.data_dir, Config.data_dir, conf)
```
jackrobison: I think it would be nicer to have the new function return the path to be used rather than having it update the attribute on the config object itself, like

```conf.data_dir = change_to_default_if_inaccessible(conf.data_dir, Config.data_dir)```

This way you don't need the `setattr` in this function.

File: lbry/lbry/extras/cli.py [hunk 3]
```diff
@@ -276,14 +249,69 @@ def __exit():
...
+
+def change_to_default_if_inaccessible(parsed_path: str, path_obj: Path, conf: Config):
+    assert path_obj.metavar in ('DIR', 'FILE'), "Path must be either to a directory or a file"
+    handled_errors = (OSError, IOError, PermissionError)
+    is_file = path_obj.metavar == 'FILE'
+    default_path = path_obj.default
+    dir_path = os.path.basename(parsed_path) if is_file else parsed_path
+    default_dir_path = os.path.basename(default_path) if is_file else default_path
+    try:
+        prepare_directory(dir_path)
```
jackrobison: if `prepare_directory` returns the directory path when it doesn't fail, you can just `return prepare_directory(dir_path)` here, and not need the nested try/except (only logging the warning in the first except, and the critical in the second non-nested except before re-raising)

lbry/lbry/extras/cli.py
```diff
@@ -276,14 +249,69 @@ def __exit():
...
+    except handled_errors:
+        log.warning("Failed to access %s, setting to %s", dir_path, default_dir_path)
+        setattr(conf, path_obj.name, default_dir_path)
+        try:
+            prepare_directory(default_dir_path)
+        except handled_errors as e:
```
jackrobison: unless you're using the caught exception for something other than just re-raising, you can just 
```
except handled_errors:
    log.critical(...)
    raise
```

mirgee:
@jackrobison Don't merge yet please, I will add tests.

eukreign:
@mirgee are you still working on this?

mirgee:
@eukreign Ready for review

eukreign:
@lyoshenka i'm still reviewing / working on this one

mirgee:
@lyoshenka It has been ready on my part for some time.

eukreign:
@mirgee thanks for the effort you put into this pull request and my apologies it took so long to make a decision on it.

I think the code is good but the solution is not intuitive: The original problem was that `lbrynet` does not start when downloads directory is unreadable and the expected behavior would be that it does start and provide some kind of message to the user. As of current version of `lbrynet` it does start with invalid downoad directory, so the original issue is resolved. This PR introduces a new behavior where `lbrynet` would ignore user supplied directory and use some other directory instead (the "default" directory); I think that poses security and usability issues if `lbrynet` can circumvent user specified paths.

Issue #1832, opened 2019-02-01
numeric cli arguments not handled properly


tzarebczan:
## The Issue
I published a claim under lbry://1112, but when running `lbrynet file list --claim_name=1112` (or ```--claim_name="1112"``` the result does not return. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

tzarebczan:
This is fixed in #2474

PR #2474, opened 2019-09-20
Handle claim_name as a string

mirgee:
This is a simple change to address #1832. Currently, when claim name is a digit, e.g. '1234' (docopt automatically converts to string), it is normalized to integer. Since it has been published as a string, the comparison then fails and the claim is not retrievable via either `lbrynet file list --claim_name=1234` or `lbrynet file list --claim_name="1234"` (tested for both).

### DIFF
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -45,7 +45,7 @@ async def execute_command(conf, method, params, callback=display):
 def normalize_value(x, key=None):
     if not isinstance(x, str):
         return x
-    if key in ('uri', 'channel_name', 'name', 'file_name', 'download_directory'):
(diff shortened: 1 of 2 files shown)


eukreign:
@mirgee thanks! any chance you could add a unit test to PR?

mirgee:
@eukreign Sure; I did add a check to `test_cli.py`, do you have something else in mind?

eukreign:
> @eukreign Sure; I did add a check to `test_cli.py`, do you have something else in mind?

Looks good, thanks, I'll merge when it's green.

tzarebczan:
@mirgee thanks again for another PR ! Shoot us an email so we can show you some [appreciation](https://lbry.com/faq/appreciation)

Issue #2084, opened 2019-05-06
Invalid characters on file name


tzarebczan:
## The Issue
Recently ran into a file that was uploaded by a mac user, where I assume a ? is a valid character in a file name. But on windows, the file will fail to save because it's invalid. Should we not allow this to begin with or strip it out automatically on download?

```2019-05-06 08:59:45,048 INFO     lbrynet.stream.managed_stream:333: save file for lbry://whycantbitcoinbeshutdownorregulated#bdb94799b6bafd444985f7ec8e8ea85af9982fdc (sd hash 23d1e1...) -> c:\testing\lbry036\downloads\Why can't Bitcoin be shut down or regulated?.mov
2019-05-06 08:59:45,048 ERROR    lbrynet.stream.managed_stream:359: unexpected error encountered writing file for stream 23d1e1afe89821db541513fa786d5bd69d777bae67d49c0746ec5277db764699097cee8f91b3fcce0444dec705163bc3
Traceback (most recent call last):
  File "c:\users\thoma\documents\lbry\lbrynet\stream\managed_stream.py", line 338, in _save_file
    with open(output_path, 'wb') as file_write_handle:
OSError: [Errno 22] Invalid argument: "c:\\testing\\lbry036\\downloads\\Why can't Bitcoin be shut down or regulated?.mov"
```
## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

mirgee:
@tzarebczan I think that disallowing it on upload is bad UX and unncessary. It is easy to forget that we sanitize something, so I am for implementing sanitization during both upload and download. If I was to choose one, I would go with the upload - since AFAIK the client side can be bypassed, we shouldn't broadcast streams with potentially invalid metadata in the first place. Is there some concensus on this?

mirgee:
@tzarebczan On the other hand, the [spec](https://lbry.tech/spec#metadata-validation) reads:

[...]

PR #2476, opened 2019-09-23
Sanitize file names before saving

mirgee:
This PR is to address #2084. I added a setter to `StreamManager.file_name` property which sanitizes it the input value. Getter sanitizes `suggested_file_name` if necessary. The sanitization strips trailing whitespaces, replaces spaces with underscores and removes all non-alphanumeric characters except for dashes and dots. I added a unit test and will add an integration test if you agree with this approach.

### DIFF
--- a/lbry/lbry/stream/descriptor.py
+++ b/lbry/lbry/stream/descriptor.py
@@ -89,7 +89,7 @@ class StreamDescriptor:
         self.blob_dir = blob_dir
         self.stream_name = stream_name
         self.key = key
-        self.suggested_file_name = sanitize_file_name(suggested_file_name)
+        self.suggested_file_name = suggested_file_name
         self.blobs = blobs
         self.stream_hash = stream_hash or self.get_stream_hash()
         self.sd_hash = sd_hash
@@ -251,9 +251,10 @@ class StreamDescriptor:
             blobs.append(blob_info)
         blobs.append(
             BlobInfo(len(blobs), 0, binascii.hexlify(next(iv_generator)).decode()))  # add the stream terminator
+        file_name = os.path.basename(file_path)
+        suggested_file_name = sanitize_file_name(file_name)
         descriptor = cls(
-            loop, blob_dir, os.path.basename(file_path), binascii.hexlify(key).decode(), os.path.basename(file_path),
-            blobs
+            loop, blob_dir, file_name, binascii.hexlify(key).decode(), suggested_file_name, blobs
         )
         sd_blob = await descriptor.make_sd_blob(old_sort=old_sort, blob_completed_callback=blob_completed_callback)
         descriptor.sd_hash = sd_blob.blob_hash
--- a/lbry/lbry/stream/descriptor.py
+++ b/lbry/lbry/stream/descriptor.py
@@ -15,6 +15,17 @@ from lbry.error import InvalidStreamDescriptorError
 
 log = logging.getLogger(__name__)
 
+_exprs = [
+    '[<>:"/\\\|\?\*]+',  # Illegal characters
+    '[\\x00-\\x1F]+',  # All characters in range 0-31
+    '[ \t]*(\.)+[ \t]*$',  # Dots at the end
+    '(^[ \t]+|[ \t]+$)',  # Leading and trailing whitespace
+    '^CON$', '^PRN$', '^AUX$',  # Illegal names
+    '^NUL$', '^COM[1-9]$', '^LPT[1-9]$'
+]
+
+RES = re.compile('(' + '|'.join((expr for expr in _exprs)) + ')')
+
 
 def format_sd_info(stream_name: str, key: str, suggested_file_name: str, stream_hash: str,
                    blobs: typing.List[typing.Dict]) -> typing.Dict:
@@ -48,16 +59,6 @@ def file_reader(file_path: str):
 
 
 def sanitize_file_name(dirty_name: str):
-    exprs = [
-        '[<>:"/\\\|\?\*]+',                         # Illegal characters
-        '[\\x00-\\x1F]+',                           # All characters in range 0-31
-        '[ \t]*(\.)+[ \t]*$',                       # Dots at the end
-        '(^[ \t]+|[ \t]+$)',                        # Leading and trailing whitespace
-        '^CON$', '^PRN$', '^AUX$',                  # Illegal names
-        '^NUL$', '^COM[1-9]$', '^LPT[1-9]$']
-
-    RES = re.compile('(' + '|'.join((expr for expr in exprs)) + ')')
-
     file_name, ext = os.path.splitext(dirty_name)
     file_name = re.sub(RES, '', file_name)
     ext = re.sub(RES, '', ext)
--- a/lbry/lbry/stream/descriptor.py
+++ b/lbry/lbry/stream/descriptor.py
@@ -4,6 +4,7 @@ import binascii
 import logging
 import typing
 import asyncio
+import re
 from collections import OrderedDict
 from cryptography.hazmat.primitives.ciphers.algorithms import AES
 from lbry.blob import MAX_BLOB_SIZE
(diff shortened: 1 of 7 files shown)


tzarebczan:
Thanks for the PR @mirgee . We'll have to review if this is the right approach or not. Not sure we want to stripe all the characters you mentioned.

mirgee:
@jackrobison So the the sanitization should be applied apply only to suggested file name when a stream is created / updated, and we should attempt to save the file using the suggested file name (and fail if it contains illegal characters)? Or try saving using `file_name` then try suggested file name in case of failure?

jackrobison:
@mirgee 

If a file is being created for a download, the default file name should be the suggested file name (the downloader should also sanitize this on their end, as old streams pre-sanitization will exist and weird streams can be constructed manually). 

The file name shouldn't be sanitized for the publisher, but the suggested file name in the sd blob they generate should be sanitized. When the publisher runs `file_list` they should see the path to the file they published - it shouldn't silently get renamed or duplicated with a sanitized name. If they deleted this file and then re-downloaded/saved it, the behavior should be the same as for someone downloading/saving it for the first time (it should be the sanitized suggested file name unless explicitly provided as an argument). 

If a file name is provided manually as an argument to `file_save` or `get` when downloading/saving I don't think it should ever be sanitized as this would be unexpected.

lbry/lbry/stream/managed_stream.py
```diff
@@ -381,7 +381,7 @@ def _write_decrypted_blob(handle: typing.IO, data: bytes):
             os.mkdir(self.download_directory)
         self._file_name = await get_next_available_file_name(
             self.loop, self.download_directory,
-            file_name or self.descriptor.suggested_file_name
+            file_name or self.file_name
```
jackrobison: `file_name or sanitize_file_name(self.file_name)`

lbry/lbry/stream/descriptor.py
```diff
@@ -65,7 +89,7 @@ def __init__(self, loop: asyncio.AbstractEventLoop, blob_dir: str, stream_name:
         self.blob_dir = blob_dir
         self.stream_name = stream_name
         self.key = key
-        self.suggested_file_name = suggested_file_name
+        self.suggested_file_name = sanitize_file_name(suggested_file_name)
```
jackrobison: this one isn't necessary

lbry/lbry/stream/managed_stream.py
mirgee: In `download_stream_from_uri()`, `ManagedStream` is first instantiated with input `file_name` and then `save_file()` is called without `file_name` argument. If the the user calls tries to call `jsonrpc_get` with some file name, they will get the sanitized suggested file name. Changing this on the callers side makes the code less intuitive and readble, so I will make this line `file_name or self._file_name or sanitize_file_name(self.descriptor.suggested_file_name)`.
https://github.com/lbryio/lbry-sdk/blob/master/lbry/lbry/stream/stream_manager.py#L416

lbry/lbry/stream/descriptor.py
```diff
@@ -46,6 +47,29 @@ def file_reader(file_path: str):
...
+def sanitize_file_name(dirty_name: str):
+    exprs = [
+        '[<>:"/\\\|\?\*]+',                         # Illegal characters
+        '[\\x00-\\x1F]+',                           # All characters in range 0-31
+        '[ \t]*(\.)+[ \t]*$',                       # Dots at the end
+        '(^[ \t]+|[ \t]+$)',                        # Leading and trailing whitespace
+        '^CON$', '^PRN$', '^AUX$',                  # Illegal names
+        '^NUL$', '^COM[1-9]$', '^LPT[1-9]$']
+
+    RES = re.compile('(' + '|'.join((expr for expr in exprs)) + ')')
```
eukreign: This makes it seem like a constant but it actually gets recompiled every time `sanitize_file_name` is called. Perhaps move this out of the function so that the compilation only happens once?

lbry/lbry/stream/descriptor.py
```diff
@@ -14,6 +15,17 @@
...
+    '[<>:"/\\\|\?\*]+',  # Illegal characters
+    '[\\x00-\\x1F]+',  # All characters in range 0-31
+    '[ \t]*(\.)+[ \t]*$',  # Dots at the end
+    '(^[ \t]+|[ \t]+$)',  # Leading and trailing whitespace
+    '^CON$', '^PRN$', '^AUX$',  # Illegal names
+    '^NUL$', '^COM[1-9]$', '^LPT[1-9]$'
+]
+
+RES = re.compile('(' + '|'.join((expr for expr in _exprs)) + ')')
+
```
eukreign: Can you embed `_exprs` into the `re.compile()` expression? Also please rename `RES` to something that makes sense. General convention in Python is to use concise but descriptive variable names; especially for public (not internal to a function) variables.

lbry/lbry/stream/descriptor.py
```diff
@@ -46,6 +58,19 @@ def file_reader(file_path: str):
             offset += bytes_to_read
 
 
+def sanitize_file_name(dirty_name: str):
+    file_name, ext = os.path.splitext(dirty_name)
+    file_name = re.sub(RES, '', file_name)
+    ext = re.sub(RES, '', ext)
+
+    if not file_name:
+        log.warning('Unable to suggest a file name for %s', dirty_name)
```
eukreign: In what situation would this happen? Should it maybe return the `dirty_name` in this case? Should the function throw an error?

lbry/lbry/testcase.py
```diff
@@ -207,10 +210,13 @@ def cleanup():
                 pass
 
         self.addCleanup(cleanup)
```
eukreign: Are you sure this is necessary? Files are automatically closed in Python.

lbry/tests/unit/stream/test_stream_descriptor.py
```diff
@@ -78,6 +78,11 @@ def _write_sd(self):
         self.sd_dict['blobs'][-2]['length'] = 0
         await self._test_invalid_sd()
 
+    def test_sanitize_file_name(self):
+        test_cases = [' t/-?t|.g.ext ', 'end_dot .', '.file\0\0', 'test n\16ame.ext', 'COM8', 'LPT2', '']
+        expected = ['t-t.g.ext', 'end_dot', '.file', 'test name.ext', '', '', '']
+        actual = [sanitize_file_name(tc) for tc in test_cases]
+        self.assertListEqual(actual, expected)
```
eukreign: It's difficult to read and debug tests written like this. Can you spell out each test on its own line?

Example from other tests:
https://github.com/lbryio/lbry-sdk/blob/9bb0ca2f0d7ef428d17ee8bdec89d0bb17f99dca/lbry/tests/unit/schema/test_tags.py#L12-L16

https://github.com/lbryio/lbry-sdk/blob/9bb0ca2f0d7ef428d17ee8bdec89d0bb17f99dca/torba/tests/client_tests/unit/test_utils.py#L10-L13

lbry/lbry/testcase.py
mirgee: For tests purposes, I don't see why it would be strictly necessary, since a tempfile won't need reopened after the test finishes, we don't care about system resources as much... But maybe I am missing something? What if CPython implementation changes? Is there a reason to not close the file deterministically?

lbry/lbry/stream/descriptor.py
mirgee: When after removing all illegal characters, the file name is still empty. We may return `dirty_name` (the client sanitizes file names as well) and/or throw an error. I like that returning an empty string makes it very easy to pick between `file_name` and `sanitized_file_name` using `or`, but maybe throwing and error signals the intent more clearly.

lbry/lbry/testcase.py
eukreign: CPython is not removing the garbage collector so files have always been automatically closed and will always be automatically closed into the foreseeable future. Unless the file is not getting cleaned up?

Generally I strongly dislike having code that doesn't do anything. If this code doesn't do anything (because Python cleans up the file automatically) I don't see the point of keeping this technical debt around.

lbry/lbry/stream/descriptor.py
eukreign: What if the filename is `CON`? Why do we bother to sanitize it if we're then going to allow that filename (by doing `file_name or sanitized_file_name` as you suggest)?

I'm leaning more towards throwing an error. @jackrobison @tzarebczan what do you guys think?

lbry/lbry/testcase.py
mirgee: The tests fail without the cleanup - the tempfile is deleted before it can be read...

eukreign: What if you change it to `self.file` instead? Then the variable will stick around until after the test is fully done and cleaned up.

mirgee: That works, do you think it's cleaner?

lbry/lbry/stream/descriptor.py
mirgee: Because maybe we should at least try to save the file using the original file name before failing, at least when downloading?

jackrobison: I don't think a weird name should cause the download to fail. If the name gets sanitized down to nothing, it should probably default to `_`, or something more descriptive like `lbry_download`

Issue #1739, opened 2019-01-03
Clearer verbose instructions


tzarebczan:
## The Issue
When running lbrynet, there are no instructions on how to run verbose mode. To run it correctly, it needs to be executed as: `lbrynet start --verbose=lbrynet`. Is the lbrynet part required? What are other options there?  At a minimum, this should be included in the lbrynet help / lbrynet response. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

mirgee:
@tzarebczan Currently, the `--verbose` param does not behave as per doc - it is not possible to change the loglevel for any other module than whole lbrynet (setting it to anything changes it to lbrynet) and not setting value does not change loglevel at all. What we should do is probably to make lbrynet the default (when `--verbose` is set), and allow used to select a (proper) module?

tzarebczan:
@shyba do you have an opinion here? I like lbrynet as default - even that's not clear today. Then we can create a separate issue to allow tracking/configuration of certain modules

shyba:
yes, we broke that at some point and I didn't realize. Thanks for the issue!
@tzarebczan agree too. I think `--verbose` should make it all go to DEBUG, so when we dont have something specific we can still see it all there. The selective verbose is very helpful too, since some modules are very noisy and sometimes we are investigating logs only under a specific module.

PR #2478, opened 2019-09-23
Make `--verbose` param behave as per documentation

mirgee:
Addressing #1739, this PR makes the `--verbose` param behave as per documentation: when it is specified as flag, i.e.  `--verbose`,  debug level is set for `lbry` logger as default, and when a list of strings is passed, debug level is set to each as a logger. This does not check that the loggers are valid modules.

### DIFF
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -242,11 +242,14 @@ def run_daemon(args: list, conf: Config):
 
     loop = asyncio.get_event_loop()
 
-    log.setLevel(logging.DEBUG if args.verbose == [] else logging.INFO)
-    for module in args.verbose or []:
-        logging.getLogger(module).setLevel(logging.DEBUG)
-    if isinstance(args.verbose, list):
+    log.setLevel(logging.INFO)
+    if args.verbose is not None:
         loop.set_debug(True)
+        if len(args.verbose) > 0:
+            for module in args.verbose:
+                logging.getLogger(module).setLevel(logging.DEBUG)
+        else:
+            log.setLevel(logging.DEBUG)
 
     if conf.share_usage_data:
         loggly_handler = get_loggly_handler()
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -189,7 +189,7 @@ def get_argument_parser():
     )
     start.add_argument(
         '--verbose', nargs="*",
-        help=('Enable debug output. Optionally specify loggers for which debug output '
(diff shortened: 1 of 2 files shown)


tzarebczan:
Thanks for the PR @mirgee ! We'll get a review on it soon. 

As usual, can we send you some [appreciation](https://lbry.com/faq/appreciation) for the contribution?

lbry/lbry/extras/cli.py
```diff
@@ -242,11 +242,11 @@ def run_daemon(args: list, conf: Config):
...
 
-    if args.verbose:
-        log.setLevel(logging.DEBUG)
+    log.setLevel(logging.DEBUG if args.verbose == [] else logging.INFO)
+    for module in args.verbose or []:
+        logging.getLogger(module).setLevel(logging.DEBUG)
+    if isinstance(args.verbose, list):
         loop.set_debug(True)
-    else:
-        log.setLevel(logging.INFO)
```
eukreign: The conditionals and logic here is a bit convoluted and hard to follow, can you simplify/clean this up?

mirgee: Hopefully cleared it up. I can add comments or a few new variables, if you think it's necessary...

lbry/lbry/extras/cli.py
```diff
@@ -242,11 +242,12 @@ def run_daemon(args: list, conf: Config):
...
-    if args.verbose:
-        log.setLevel(logging.DEBUG)
+    modules = args.verbose
+    if modules is not None:
         loop.set_debug(True)
-    else:
-        log.setLevel(logging.INFO)
+        log.setLevel(logging.DEBUG if modules == [] else logging.INFO)
+        for module in modules:
+            logging.getLogger(module).setLevel(logging.DEBUG)
```
eukreign: i think it's still convoluted, also adding more variables usually makes things more complicated (and if you need to add comments then it's even more complicated :-p)

what about (no additional variables and no inline `if` statement):
```
    if args.verbose is not None:
        loop.set_debug(True)
        if len(args.verbose) > 0:
            log.setLevel(logging.INFO)
            for module in args.verbose:
                logging.getLogger(module).setLevel(logging.DEBUG)
        else:
            log.setLevel(logging.DEBUG)
```

mirgee: I agree - it's that I didn't know why it was not readable and thought naming things would help. This could work.

eukreign:
@mirgee can you rebase this and we'll merge

PR #2482, opened 2019-09-25
normalize tag names (remove duplicates, blank spaces and illegal chars)

### LINKED ISSUE #2399: normalize tag names
Channel @animeforancaps has Anime and anime tags. We should normalize when saving and reading. When reading, if 2 normalized values overlap, return only 1.

mirgee:
fixes #2399. Now, tags are normalized when read, and when claims are created / updated. Duplicate tags (after normalization) are not inserted. Added some unit and integration tests. Maybe `normalize_tag()` and `clean_tags` can be moved to `attrs.py`. (I will merge to rerun the build once the pylint fix is on master).

### DIFF
--- a/lbry/lbry/schema/attrs.py
+++ b/lbry/lbry/schema/attrs.py
@@ -544,7 +544,3 @@ class TagList(BaseMessageList[str]):
         tag = normalize_tag(tag)
         if tag and tag not in self.message:
             self.message.append(tag)
-
-    def extend(self, tags: List[str]):
-        for tag in tags:
-            self.append(tag)
--- a/lbry/lbry/schema/attrs.py
+++ b/lbry/lbry/schema/attrs.py
@@ -540,17 +540,11 @@ class TagList(BaseMessageList[str]):
     __slots__ = ()
     item_class = str
 
-    @classmethod
-    def create_normalized(cls, message):
-        message[:] = clean_tags(message)
-        return cls(message)
-
     def append(self, tag: str):
         tag = normalize_tag(tag)
-        if tag not in self._message:
-            self._message.append(tag)
+        if tag and tag not in self.message:
+            self.message.append(tag)
 
     def extend(self, tags: List[str]):
-        tags = clean_tags(tags)
-        tags = list(set(tags).difference(set(self._message)))
-        self._message.extend(tags)
+        for tag in tags:
+            self.append(tag)
--- a/lbry/lbry/schema/attrs.py
+++ b/lbry/lbry/schema/attrs.py
@@ -540,16 +540,17 @@ class TagList(BaseMessageList[str]):
     __slots__ = ()
     item_class = str
 
-    def __init__(self, message):
+    @classmethod
+    def create_normalized(cls, message):
         message[:] = clean_tags(message)
-        self.message = message
+        return cls(message)
 
     def append(self, tag: str):
         tag = normalize_tag(tag)
-        if tag not in self.message:
-            self.message.append(tag)
+        if tag not in self._message:
+            self._message.append(tag)
 
     def extend(self, tags: List[str]):
         tags = clean_tags(tags)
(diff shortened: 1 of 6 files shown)


tzarebczan:
Thanks again for another PR! I know @eukreign had some reservations about doing this on the SDK side, but maybe he still can be convinced. We'll get a review on it soon.

eukreign:
The main reason I am against this is that it then makes it impossible to ever see the original claim (as it is on the blockchain).

It's fine to make it difficult to create claims with duplicate tags, but if you are loading a claim from the blockchain you should get values as close to what's on the blockchain as possible.

@mirgee can you change the PR so that it still shows the original tag list as created by the claim author on the blockchain? optionally, in the json response encoder or some other higher level we can add a way for users to request original vs. cleaned claim tag data.

lbry/lbry/schema/tags.py
```diff
@@ -10,4 +10,9 @@ def normalize_tag(tag: str):
 
 
 def clean_tags(tags: List[str]):
-    return [tag for tag in set(normalize_tag(tag) for tag in tags) if tag]
+    clean = []
+    for ind, tag in enumerate(tags):
+        norm_tag = normalize_tag(tag)
+        if norm_tag and norm_tag not in clean[:ind]:
```
eukreign: We use `i` or `idx` or `index` as the convention for index variables.

lbry/lbry/schema/base.py
```diff
@@ -119,3 +119,6 @@ def __getitem__(self, item) -> I:
 
     def __delitem__(self, key):
         del self._message[key]
+
+    def __eq__(self, values: List[str]) -> bool:
+        return self._message == values
```
eukreign: Where is the equality method used?

lbry/lbry/schema/attrs.py
```diff
@@ -533,3 +534,17 @@ class LocationList(BaseMessageList[Location]):
...
+    item_class = str
+
+    def append(self, tag: str):
+        tag = normalize_tag(tag)
+        if tag and tag not in self.message:
+            self.message.append(tag)
+
+    def extend(self, tags: List[str]):
+        for tag in tags:
+            self.append(tag)
```
eukreign: How is this different from the base `BaseMessageList.extend()`?

eukreign:
@mirgee see feedback above, also this PR needs to be rebased, it has conflicts on `test_claim_commands.py`

lbry/lbry/schema/base.py
mirgee: Only in tests I think, to compare tags with lists of strings more simply.

mirgee:
@eukreign Ready

lbry/lbry/schema/tags.py
```diff
@@ -10,4 +10,9 @@ def normalize_tag(tag: str):
 
 
 def clean_tags(tags: List[str]):
-    return [tag for tag in set(normalize_tag(tag) for tag in tags) if tag]
+    clean = []
+    for idx, tag in enumerate(tags):
+        norm_tag = normalize_tag(tag)
+        if norm_tag and norm_tag not in clean[:idx]:
+            clean.append(norm_tag)
+    return clean
```
eukreign: What does this do differently from the original?

mirgee: Sorry, missed this. This preserves the order of the input, which is no longer necessary after changing `extend()`.

eukreign: So, it can be reverted to the original?

mirgee: Yes. To have the unit tests on the same place, how about moving the two methods from `tags.py` to `attrs.py`, and the tests from `test_tags.py` to `test_schema.py`? I don't think that the integration tests duplicate the unit tests (except in terms of coverage) - they test the same functionality of different parts. But I suppose you would prefer to remove the unit test?

PR #2484, opened 2019-09-25
change trending window to 6 hours

lyoshenka:
@eukreign is this all I need to change, or does `TRENDING_DATA_POINTS` need to be changed as well. the comment on `TRENDING_DATA_POINTS` was not clear to me.

Does this change require a resync for the wallet server?

### DIFF
--- a/lbry/lbry/wallet/server/db/trending.py
+++ b/lbry/lbry/wallet/server/db/trending.py
@@ -1,6 +1,6 @@
 from math import sqrt
 
-TRENDING_WINDOW = 650  # number of blocks, ~24hr period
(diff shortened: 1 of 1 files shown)


eukreign:
> @eukreign is this all I need to change, or does `TRENDING_DATA_POINTS` need to be changed as well. the comment on `TRENDING_DATA_POINTS` was not clear to me.

`TRENDING_DATA_POINTS` is frequency and `TRENDING_DATA_POINTS` is how much data to track, they can be modified independently.

> Does this change require a resync for the wallet server?

Yes, this requires resync.

PR #2489, opened 2019-09-27
DHT object cache revamp

shyba:


lbry/lbry/dht/peer.py
```diff
@@ -135,36 +139,27 @@ def peer_is_good(self, peer: 'KademliaPeer'):
...
+    protocol_version: typing.Optional[int] = field(default=1, compare=False)
+
+    def __post_init__(self):
+        if self._node_id is not None:
+            if not len(self._node_id) == constants.hash_length:
+                raise ValueError("invalid node_id: {}".format(hexlify(self._node_id).decode()))
+        if self.udp_port is not None and not 0 <= self.udp_port <= 65536:
             raise ValueError("invalid udp port")
-        if tcp_port and not 0 <= tcp_port <= 65536:
+        if self.tcp_port and not 0 <= self.tcp_port <= 65536:
```
jackrobison: this should check if tcp_port is not None

PR #2497, opened 2019-10-02
bug fixes related to cross-account / wallet operations

eukreign:


tzarebczan:
There's still more to be done here right?

PR #2500, opened 2019-10-04
Made period usage constant

Mehvix:


eukreign:
Thanks for the PR but unfortunately these are outdated docs. Our docs website https://lbry.tech/api/sdk uses just the [api.json](https://github.com/lbryio/lbry-sdk/blob/master/lbry/docs/api.json) file which is generated by [generate_json_api.py](https://github.com/lbryio/lbry-sdk/blob/master/lbry/scripts/generate_json_api.py). We will delete the html files in the docs directory soon.

PR #2506, opened 2019-10-04
start of error codes

lyoshenka:
next steps:

- [x] rename BaseLBRYException to something like "ErrorCodeException". call it based on what it does. don't use "LBRY" in name
- [ ] get all exception codes into a single place where it's clear what they all are. maybe group them smartly. then generate exception classes from the list. do it like https://github.com/lbryio/lbry-sdk/wiki/Error-Codes but in code


eukreign:
replaced by #2611

PR #2498, opened 2019-10-02
code reformatting and update unit tests to use more specific asserts

hugovk:
* Upgraded Python syntax with [`pyupgrade --py37-plus`](https://github.com/asottile/pyupgrade)

* Upgraded unit tests to use more useful asserts

* `sudo` is no longer needed in `.travis.yml`
   * https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

PR #2508, opened 2019-10-05
use more specific asserts in torba tests

StrikerRUS:
Use `assertListEqual`/`assertTupleEqual`/`assertDictEqual` instead of `asserEqual`.
This PR doesn't conflict with #2498, in which `assertTrue`/`assertFalse` replaces `assertEqual`.

### DIFF
--- a/torba/tests/client_tests/integration/test_network.py
+++ b/torba/tests/client_tests/integration/test_network.py
@@ -18,7 +18,7 @@ class NetworkTests(IntegrationTestCase):
         self.assertEqual(self.ledger.network.remote_height, initial_height + 1)
 
     async def test_server_features(self):
-        self.assertEqual({
+        self.assertDictEqual({
             'genesis_hash': self.conductor.spv_node.coin_class.GENESIS_HASH,
             'hash_function': 'sha256',
             'hosts': {},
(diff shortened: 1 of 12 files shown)


eukreign:
I would like to keep the no-spaces `20*CENT` version of amounts because it acts as a sort of domain specific language in the tests. Can you change the amount back to no-spaces and also please rebase the PR.

StrikerRUS:
@eukreign Thanks for your review! I wanted just to make those lines consistent with 
https://github.com/lbryio/lbry-sdk/blob/077281a77a82cfdfca0ec02922d4e582233bae43/torba/tests/client_tests/unit/test_coinselection.py#L66-L69

I'm sorry about that! I've just reverted back my changes related to spaces around `*` operator. Please check it.

StrikerRUS:
I've just commited the same changes for `integration` folder in `torba`.

eukreign:
@StrikerRUS okay, looks good now, before I merge this can you explain the pros/cons of the plain assertEqual vs the specific ones? what benefits does this PR bring over what was there before?

For example:
```python
        self.assertEqual(network.client.server, ('127.0.0.1', 1337))
        self.assertTupleEqual(network.client.server, ('127.0.0.1', 1337))
```

Is there a case where a plain `assertEqual` would not catch something that `assertTupleEqual` does catch?

StrikerRUS:
@eukreign Sure! The obvious advantage of using specific tests that you cannot change any API without changing the corresponding test. Also, such asserts increase readability a lot and remove costs of the need to detect appropriate assert inside simple `assertEqual`.

For instance, you'll not be able to simply change `self.assertEqual(foo(), ('one', 'two'))` to `self.assertEqual(foo(), ('one'))`

tzarebczan:
@StrikerRUS Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the contribution? 

Also, we're giving away some [Hacktoberfest](https://lbry.com/news/hacktoberfest-2019) bonuses and goodies for this month.

PR #2510, opened 2019-10-06
made return more explicit

OrginalS:


eukreign:
Our tests are not required to follow code quality best practices because a lot of the time breaking the rules leads to more readable / easier to follow tests. For code in `lbry/lbry/*` and `torba/torba/*` we do perform code quality checks.

PR #2513, opened 2019-10-06
README.md updated

modihere:
The steps for running the SDK in Windows and Linux have been elaborated a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,16 @@ Our [releases page](https://github.com/lbryio/lbry-sdk/releases) contains pre-bu
 
 ## Usage
 
-Run `lbrynet start` to launch the API server.
+#### For systems running Windows:
+
+Check into the folder where the binary is extracted and run `lbrynet start` to launch the API server.
+
+#### For systems running Linux:
+
+Check into the folder where the binary is extracted and run `./lbrynet start` to launch the API server.
+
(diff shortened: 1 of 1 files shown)


eukreign:
Thanks for the PR but we already have detailed instructions for Windows/Linux installation in [INSTALL.md](https://github.com/lbryio/lbry-sdk/blob/master/INSTALL.md). If there is something missing please add it to the [INSTALL.md](https://github.com/lbryio/lbry-sdk/blob/master/INSTALL.md) file.

modihere:
Hey, i went through the install.md file and i guess the steps for installing from source is given there not the steps for running the binary file(pardon if i have missed). So i added the steps, even though they are very basic ones, for running the binary in README.md.

modihere:
> Thanks for the PR but we already have detailed instructions for Windows/Linux installation in [INSTALL.md](https://github.com/lbryio/lbry-sdk/blob/master/INSTALL.md). If there is something missing please add it to the [INSTALL.md](https://github.com/lbryio/lbry-sdk/blob/master/INSTALL.md) file.

Hey, i went through the install.md file and i guess the steps for installing from source is given there not the steps for running the binary file(pardon if i have missed). So i added the steps, even though they are very basic ones, for running the binary in README.md.

PR #2514, opened 2019-10-07
Update README.md

albin-jose:
Added PR Welcome badge, License badge and Table of contents


eukreign:
thanks for this PR but at the moment the main README is rather short, I don't think it needs a table of contents since everything mostly fits on a single screen; perhaps later if it gets to be longer than a screen we can add this.

PR #2517, opened 2019-10-07
Update INSTALL.md

Bharat123rox:
Separate instructions from commands, update some dated hyperlinks and suggest more relevant hyperlinks


eukreign:
We want people to specifically use python 3.7 that's why the link points to that version directly.

PR #2547, opened 2019-10-15
Use loggly on spv servers for error logging

jackrobison:
Enabled by setting the `LOGGLY_TOKEN` env variable


lyoshenka:
from jack: needs to be updated to not share the loop between processes

PR #2548, opened 2019-10-15
added unit test for keyboard interrupt occurring on startup

cbaumler:
Add a unit test to cover the case where the user performs a keyboard interrupt from the CLI.

### DIFF
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -276,6 +276,7 @@ def run_daemon(args: argparse.Namespace, conf: Config):
         pass
     finally:
         loop.run_until_complete(daemon.stop())
+        logging.shutdown()
 
     if hasattr(loop, 'shutdown_asyncgens'):
         loop.run_until_complete(loop.shutdown_asyncgens())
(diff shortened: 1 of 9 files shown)


tzarebczan:
Thanks for the PR! What will this prevent/fix? 

Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the contribution? 

Also, we're giving away some [Hacktoberfest](https://lbry.com/news/hacktoberfest-2019) bonuses and goodies for this month.

cbaumler:
@tzarebczan I noticed that the unit tests weren't achieving code coverage of this code in `cli.py`:
```
try:
    loop.run_until_complete(daemon.start())
    loop.run_forever()
except (GracefulExit, KeyboardInterrupt, asyncio.CancelledError):
    pass
finally:
    loop.run_until_complete(daemon.stop())
````

I added a unit test to cover the case of stopping the Daemon when a keyboard interrupt is issued. When running the unit test, I discovered an issue with shutting down the daemon runner where an exception would occur, because the `on_shutdown` signal needed to be frozen.

eukreign:
@cbaumler thanks for the PR, turns out the `freeze()` wasn't necessary because `shutdown()` isn't necessary either (actually according to docs only `cleanup()` should be called and i've confirmed that `cleanup()` eventually sends the `shutdown` signal).

PR #2555, opened 2019-10-16
refactored logging configuration and added more tests

lyoshenka:
makes setup_logging() params clearer. now you know which parts of `args` and `conf` are actually used and what they do

### DIFF
--- a/lbry/lbry/extras/cli.py
+++ b/lbry/lbry/extras/cli.py
@@ -7,10 +7,10 @@ import asyncio
 import argparse
 import logging
 import logging.handlers
-from docopt import docopt
 
 import aiohttp
 from aiohttp.web import GracefulExit
+from docopt import docopt
 
 from lbry import __version__ as lbrynet_version
 from lbry.extras.daemon.loggly_handler import get_loggly_handler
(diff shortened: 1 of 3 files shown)


lyoshenka:
i think the travis error is not real

lbry/lbry/extras/cli.py
```diff
@@ -143,7 +144,7 @@ def _granular_action_groups(self):
 
     def error(self, message):
         self.print_help(argparse._sys.stderr)
-        self.exit(2, '\n'+message+'\n')
+        self.exit(2, '\n' + message + '\n')
```
jackrobison: this can be an f-string

lbry/lbry/extras/cli.py
```diff
@@ -221,43 +222,44 @@ def ensure_directory_exists(path: str):
...
     logging.getLogger('aioupnp').setLevel(logging.WARNING)
     logging.getLogger('aiohttp').setLevel(logging.CRITICAL)
 
     log.setLevel(logging.INFO)
-    if args.verbose is not None:
+    if verbose is not None:
         loop.set_debug(True)
-        if len(args.verbose) > 0:
-            for module in args.verbose:
+        if len(verbose) > 0:
```
jackrobison: this if isn't needed

eukreign:
i'm not ignoring this PR, it's just conflicting with a higher priority one (one that's breaking travis: #2559 ) in the same area of code, i'd like to get that one merged first before looking at this one.

lyoshenka:
ok. i addressed jack's comments. merge when ready

eukreign:
@lyoshenka need rebasing

PR #2601, opened 2019-10-31
Persist and restore routing table

### LINKED ISSUE #2490: store known dht peers so we don't have to go through our seed nodes every time the daemon restarts
- save the peers we know about in a file
- try to connect to them when the dht starts up

mirgee:
This will fix #2490. The peers, their addresses (and node_ids), are stored in and restored from the local SQLite storage, along with the bucket ranges. The restored peers are pinged and are added to the routing table on response - the seeds are not persisted and tried only if no peers were persisted (I will change this so that they will be tried also if no good peers are left). More tests will be added.


lbry/lbry/dht/node.py
```diff
@@ -245,3 +227,71 @@ def accumulate_peers(self, search_queue: asyncio.Queue,
...
+                    self.protocol.routing_table.buckets_with_contacts()
+                )
+            await asyncio.sleep(1, loop=self.loop)
+
+    async def _persist_routing_table(self):
+        peers = []
+        resolved_addresses = await self._resolve_addresses(self._known_node_urls)
+        for peer in self.protocol.routing_table.get_peers():
+            if NodeURL(node_id=peer.node_id, address=peer.address, udp_port=peer.udp_port) in resolved_addresses:
+                self.protocol.routing_table.remove_peer(peer)
```
jackrobison: the seed nodes shouldn't be removed from the routing table

lbry/lbry/dht/protocol/protocol.py
```diff
@@ -242,6 +242,9 @@ def maybe_ping(self, peer: 'KademliaPeer'):
                     break
             await asyncio.sleep(1, loop=self._loop)
 
+    def is_empty(self):
+        return len(self._pending_contacts) == 0
```
jackrobison: what's this for?

lbry/lbry/dht/protocol/routing_table.py
```diff
@@ -175,6 +175,17 @@ def __init__(self, loop: asyncio.AbstractEventLoop, peer_manager: 'PeerManager',
...
+        if len(self.buckets) > 1:
+            raise RuntimeError('Routing table already in use')
+        if len(ranges) == 0:
+            return
+        self.buckets: typing.List[KBucket] = [
+            KBucket(
+                self._peer_manager, range_min=ri[0], range_max=ri[1], node_id=self._parent_node_id
+            ) for ri in ranges
+        ]
+
```
jackrobison: this can be in `__init__`, then the new function isn't necessary:


```
self._buckets = [
    KBucket(
        self._peer_manager, range_min, range_max, self._parent_node_id
    ) for range_min, range_max in bucket_ranges or [(0, 2**constants.hash_bits)]
]
```

lbry/lbry/dht/protocol/routing_table.py
```diff
@@ -317,3 +328,6 @@ def buckets_with_contacts(self) -> int:
             if len(bucket):
                 count += 1
         return count
+
+    def get_nonempty_bucket_ranges(self) -> typing.List[typing.Tuple[int, int]]:
```
jackrobison: what's this for?

lbry/lbry/dht/node.py
```diff
@@ -245,3 +227,71 @@ def accumulate_peers(self, search_queue: asyncio.Queue,
                          asyncio.Queue, asyncio.Task]:
         q = peer_queue or asyncio.Queue(loop=self.loop)
         return q, self.loop.create_task(self._accumulate_search_junction(search_queue, q))
+
+    async def _maintain_connection(self, seed_peers: typing.List['KademliaPeer'], timeout: float = None):
+        elapsed = 0
+        while True:
+            if timeout and elapsed > timeout:
```
jackrobison: This timeout should be removed - it will cause join_network to give up.

jackrobison: also, if join_network were to have a timeout the top level call could instead be wrapped in `asyncio.wait_for`

lbry/tests/unit/database/test_SQLiteStorage.py
```diff
@@ -247,3 +253,13 @@ class ContentClaimStorageTests(StorageTest):
         current_claim_info = await self.storage.get_content_claim(stream_hash)
         # this should still be the previous update
         self.assertDictEqual(current_claim_info, update_info)
+
+
+class UpdatePeersTest(StorageTest):
+    async def test_update_get_peers(self):
+        node_id = hashlib.sha384(os.urandom(32)).digest()
```
jackrobison: generally it's best to avoid using random numbers in tests since they're not repeatable and can be slow if generating lots of them

lbry/lbry/dht/node.py
mirgee: The timeout is put here so that both timed (for fixed seeds) and infinite (for restored peers) loops are callable from `join_network`.

lbry/lbry/dht/node.py
```diff
@@ -245,3 +227,71 @@ def accumulate_peers(self, search_queue: asyncio.Queue,
...
+                self.joined.set()
+                log.info(
+                    "Joined DHT, %i peers known in %i buckets", len(self.protocol.routing_table.get_peers()),
+                    self.protocol.routing_table.buckets_with_contacts()
+                )
+            await asyncio.sleep(1, loop=self.loop)
+
+    async def _persist_routing_table(self):
+        peers = []
+        resolved_addresses = await self._resolve_addresses(self._known_node_urls)
```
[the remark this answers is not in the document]
mirgee: OK, but I thought this change was mainly to avoid pinging the seeds on startup?

lbry/lbry/dht/protocol/protocol.py
mirgee: Nothing anymore, forgot to remove it

lbry/lbry/dht/protocol/routing_table.py
mirgee: Yes, I initially put it here when I was doing validation of restored buckets (which I decided to move to db and then we abandonded persisting buckets) and didn't want to have logic / exceptions in the constructor, and forgot to remove it...

lbry/lbry/dht/protocol/routing_table.py
mirgee: This was again needed for the buckets, sorry...

lbry/lbry/dht/node.py
mirgee: And the timeout is not for the whole call, just the state of waiting for peers to respond...

lbry/lbry/dht/node.py
```diff
@@ -121,56 +134,25 @@ def stop(self) -> None:
...
         self.joined.set()
 
-    def start(self, interface: str, known_node_urls: typing.List[typing.Tuple[str, int]]):
-        self._join_task = self.loop.create_task(
-            self.join_network(
-                interface=interface, known_node_urls=known_node_urls
-            )
-        )
+    def start(self, interface: str = None):
+        interface = interface or self.protocol.external_ip
```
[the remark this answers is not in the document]
mirgee: What do you mean by the default interface?

lbry/lbry/dht/node.py
jackrobison: The `join_network` loop keeps running and will cause the node to re-join if it loses connectivity and the routing table is emptied. The refresh loop could have still been running, and thus the persisted peer can have been cleared out as well by the time connectivity is restored.

The `peers` list can be populated at the start of the loop with the persisted peers, or if there aren't any with the fixed peers. Then if by the end of the loop the join wasn't yet successful the fixed peers can be added to the `peers` list if they weren't already added. This would cause the join network loop to try persisted peers and possibly the fixed peers until the next cycle of the refresh loop, and afterward fall back to only the fixed peers if the join still hasn't been successful.

lbry/lbry/dht/node.py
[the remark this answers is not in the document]
jackrobison: The purpose is to avoid relying on our seed nodes for startup, and also to speed up joining the network. The seeds shouldn't be relied upon or favored over viable alternate nodes, but they don't need to be avoided.

lbry/lbry/dht/node.py
```diff
@@ -28,6 +29,8 @@ def __init__(self, loop: asyncio.AbstractEventLoop, peer_manager: 'PeerManager',
         self.joined = asyncio.Event(loop=self.loop)
         self._join_task: asyncio.Task = None
         self._refresh_task: asyncio.Task = None
+        self._storage = storage
+        self._known_node_urls = known_node_urls or []
```
mirgee: (I know `known_node_urls` don't have to be passed on initialization, will rewrite my tests first to match the old signature of `start` first)

lbry/lbry/extras/daemon/storage.py
```diff
@@ -273,6 +273,14 @@ class SQLiteStorage(SQLiteMixin):
                 timestamp integer,
                 primary key (sd_hash, reflector_address)
             );
+
+            create table if not exists peer (
```
jackrobison: this can be cleaner with `node_id` as the primary key and a `unique (address, udp_port)` constraint

lbry/lbry/extras/daemon/storage.py
```diff
@@ -744,3 +752,20 @@ def get_streams_to_re_reflect(self):
             "where r.timestamp is null or r.timestamp < ?",
             int(self.time_getter()) - 86400
         )
+
+    # # # # # # # # # # dht functions # # # # # # # # # # #
+    async def get_peers(self):
```
jackrobison: I think this should be named somewhat more descriptively, like `get_persisted_kademlia_peers`

lbry/lbry/extras/daemon/storage.py
```diff
@@ -744,3 +752,20 @@ def get_streams_to_re_reflect(self):
...
+        query = 'select node_id, address, udp_port, tcp_port from peer'
```
jackrobison: with the primary key change, this is `select * from peer`

lbry/lbry/extras/daemon/storage.py
jackrobison: also please add a return type annotation for this

lbry/lbry/extras/daemon/storage.py
```diff
@@ -744,3 +752,20 @@ def get_streams_to_re_reflect(self):
...
+        return [(binascii.unhexlify(n), a, u, t) for n, a, u, t in await self.db.execute_fetchall(query)]
+
+    async def update_peers(self, peers: typing.List['KademliaPeer']):
```
jackrobison: `save_kademlia_peers`

lbry/lbry/dht/node.py
```diff
@@ -121,56 +126,55 @@ def stop(self) -> None:
...
-
-    def start(self, interface: str, known_node_urls: typing.List[typing.Tuple[str, int]]):
-        self._join_task = self.loop.create_task(
-            self.join_network(
-                interface=interface, known_node_urls=known_node_urls
-            )
-        )
+        restored_peers = peers_from_urls(await self._storage.get_peers()) if self._storage else []
+
+        fixed_peers = peers_from_urls([
```
jackrobison: this should be in the loop to allow reconnecting from the seeds if the internet connection drops/comes back. if `seed_peers` is empty, it can be extended with the fixed seeds.

lbry/lbry/dht/node.py
```diff
@@ -121,56 +126,55 @@ def stop(self) -> None:
...
-                 self.protocol.routing_table.buckets_with_contacts())
-        self.joined.set()
-
-    def start(self, interface: str, known_node_urls: typing.List[typing.Tuple[str, int]]):
-        self._join_task = self.loop.create_task(
-            self.join_network(
-                interface=interface, known_node_urls=known_node_urls
-            )
-        )
+        restored_peers = peers_from_urls(await self._storage.get_peers()) if self._storage else []
```
jackrobison: within the loop, this is how `seed_peers` should be initialized

lbry/lbry/dht/node.py
```diff
@@ -121,56 +126,55 @@ def stop(self) -> None:
...
+
+        fixed_peers = peers_from_urls([
+            (None, await resolve_host(address, udp_port, 'udp'), udp_port, None)
+            for address, udp_port in known_node_urls or []
+        ])
+
+        seed_peers = restored_peers or fixed_peers
+        fallback = False
+        while seed_peers:
```
jackrobison: This can be `while True` and the set_joined outside the loop can be removed - also set_joined won't be duplicated / doesnt have to be a function

jackrobison: This way the join task doesn't end immediately if there are no persisted peers and no fixed seeds are found or given.

lbry/lbry/dht/node.py
mirgee: Currently, the `seed_peers` are already changed if the search fails (and can easily extend), but doing it conditionally within the loop potentially saves DNS lookup, so agreed.

lbry/lbry/dht/node.py
mirgee: I also don't like the duplication, and I assume it is / was there only because some tests (unnecessarily?) rely on it.

Node with no seeds can still join if another node pings it, correct? Then infinite loop makes sense... (But no fixed peers are worth logging a warning, IMO.)

lbry/lbry/extras/daemon/storage.py
mirgee: Agreed.

lbry/lbry/extras/daemon/storage.py
mirgee: Being explicit instead of instead of implicits makes it easier to add a new column, change order and it makes it clear what is being queried.

lbry/lbry/dht/node.py
mirgee: Why initialize seeds within the loop?

lbry/lbry/extras/daemon/storage.py
```diff
@@ -744,3 +752,20 @@ def get_streams_to_re_reflect(self):
...
+    # # # # # # # # # # dht functions # # # # # # # # # # #
+    async def get_persisted_kademlia_peers(self) -> typing.List[typing.Tuple[bytes, str, int, int]]:
+        query = 'select node_id, address, udp_port, tcp_port from peer'
+        return [(binascii.unhexlify(n), a, u, t) for n, a, u, t in await self.db.execute_fetchall(query)]
+
+    async def save_kademlia_peers(self, peers: typing.List['KademliaPeer']):
+        def _save_kademlia_peers(transaction: sqlite3.Connection):
+            transaction.execute('delete from peer').fetchall()
+            transaction.executemany(
+                'insert into peer(node_id, address, udp_port, tcp_port) values (?, ?, ?, ?)', (
```
jackrobison: There's an unneeded extra layer of parenthesis here

lbry/tests/unit/dht/test_node.py
```diff
@@ -84,3 +86,66 @@ class TestNodePingQueueDiscover(AsyncioTestCase):
             # teardown
             for n in nodes.values():
                 n.stop()
+
+
+class TestTemporarilyLosingConnetction(AsyncioTestCase):
+
+    async def test_losing_connection(self):
+        loop = asyncio.get_event_loop()
```
jackrobison: AsyncioTestCase has `self.loop`

lbry/tests/unit/dht/test_node.py
```diff
@@ -84,3 +86,66 @@ class TestNodePingQueueDiscover(AsyncioTestCase):
...
+            advance = dht_mocks.get_time_accelerator(loop, loop.time())
+            await advance(1000)
+
+            # Join the network, assert that at least the known peers are in RT
+            self.assertTrue(node.joined.is_set())
+            self.assertTrue(len(node.protocol.routing_table.get_peers()) >= num_seeds)
+
+            # Refresh, so that the peers are persisted
+            await advance(4000)
+            self.assertTrue(len(await node._storage.get_persisted_kademlia_peers()) > num_seeds)
```
jackrobison: This can be asserted to be false before advancing the clock

lbry/tests/unit/dht/test_node.py
```diff
@@ -84,3 +86,66 @@ class TestNodePingQueueDiscover(AsyncioTestCase):
...
+
+            # We lost internet connection - all the peers stop responding
+            popped_protocols = []
+            for peer_address in peer_addresses[:-1]:
+                popped_protocols.append(dht_network.pop(peer_address))
+
+            # The peers are cleared on refresh from RT and storage
+            await advance(4000)
+            self.assertFalse(node.protocol.routing_table.get_peers())
+            self.assertFalse(await node._storage.get_persisted_kademlia_peers())
```
jackrobison: ```self.assertListEqual([], await node._storage.get_persisted_kademlia_peers())```

mirgee:
@eukreign Take a look, please

lbry/tests/unit/dht/test_node.py
```diff
@@ -84,3 +86,67 @@ class TestNodePingQueueDiscover(AsyncioTestCase):
             # teardown
             for n in nodes.values():
                 n.stop()
+
+
+class TestTemporarilyLosingConnetction(AsyncioTestCase):
```
eukreign: `Connetction` is misspelled.

lbry/tests/unit/dht/test_node.py
```diff
@@ -84,3 +86,67 @@ class TestNodePingQueueDiscover(AsyncioTestCase):
...
+
+    async def test_losing_connection(self):
+        loop = self.loop
+        loop.set_debug(False)
+
+        peer_addresses = [
+            (f'127.0.0.1', 40000+i) for i in range(10)
```
eukreign: plain string, doesn't need `f`

PR #2630, opened 2019-11-21
GitLab CI

### LINKED ISSUE #2469: Switch to Gitlab CI
### v1

- [x] build wallet server docker image (this might be doable on [docker hub](https://hub.docker.com/repository/docker/lbry/lbrynet/builds) instead of gitlab)
- [x] check that mac builds run on older macs
- [x] are build results publicly-accessible? can outsiders see why builds fail? 
- [x] make sure secret variables are not public-accessible (and test this, victor says it sometimes fails)
- [x] make sure forks and prs from forks dont build
- [x] does gitlab email the committer when build breaks
- [x] release to github on tags

### later

- [ ] code signing for osx and windows (we don't have code signing now)
- [ ] create image with all prereqs installed to speed up linux builds

from tom: executable sizes are different by a few MB vs Travis builds. Windows/Mac is smaller, Linux is larger.

lyoshenka:
- add `.gitlab-ci.yml` to enable builds on gitlab
- update some build code to work with gitlab
- move build files into toplevel `docker` directory

fixes #2469


lyoshenka:
id prefer not to use something generic like `scripts` or `tools` because it becomes a pile of unrelated files (e.g. see our current scripts dir). do you like any of these: `release`, `deploy`, `ci`?

lyoshenka:
I went with 'ci'

eukreign:
I still think `scripts` would be better place for it specifically because it brings less attention to it. Unless the plan is that ci will constantly be broken we'll have to make changes to it and it will require community attention (and thus anyone ever opening our repo should be able to quickly find the CI scripts): I'd rather it be hidden and out of the way most of the time (buried in scripts directory with everything else).

If we want to keep the root directory clean and get people directly to the important stuff, having a `ci` directory feels like an eye sore.

lbry/lbry/conf.py
```diff
@@ -630,9 +630,13 @@ def log_file_path(self):
...
 def get_windows_directories() -> typing.Tuple[str, str, str]:
-    from lbry.winpaths import get_path, FOLDERID, UserHandle  # pylint: disable=import-outside-toplevel
+    from lbry.winpaths import get_path, FOLDERID, UserHandle, \
+        PathNotFoundException  # pylint: disable=import-outside-toplevel
 
-    download_dir = get_path(FOLDERID.Downloads, UserHandle.current)
+    try:
+        download_dir = get_path(FOLDERID.Downloads, UserHandle.current)
+    except PathNotFoundException:
+        download_dir = os.getcwd()
```
eukreign: This seems like a substantial change, perhaps it should be its own PR so we can get it properly documented in the release? (bonus: we can release 0.48.1 right away with that PR to test that gitlab releases work)

docker/set_build.py
```diff
@@ -10,26 +10,26 @@
...
     log.debug("configuring build type file: %s", build_type_path)
-    travis_commit = os.environ['TRAVIS_COMMIT'][:6]
-    build_type = get_build_type(os.environ.get('TRAVIS_TAG', None))
-    log.debug("setting build type=%s, build commit=%s", build_type, travis_commit)
+    commit_hash = os.getenv('CI_COMMIT_SHA', os.getenv('TRAVIS_COMMIT', None))[:6]
+    build_type = get_build_type(os.getenv('CI_COMMIT_TAG', os.getenv('TRAVIS_TAG', None)))
+    log.debug("setting build type=%s, build commit=%s", build_type, commit_hash)
     with open(build_type_path, 'w') as f:
-        f.write(f"BUILD = \"{build_type}\"\nBUILD_COMMIT = \"{travis_commit}\"\n")
+        f.write(f"BUILD = \"{build_type}\"\nBUILD_COMMIT = \"{commit_hash}\"\n")
```
eukreign: You don't need to pass `None` as second argument to `getenv()`  since that's the default already.

For any of the environment variables that are actually required it's better to access them via `os.environ` (throws `KeyError` if variable is missing) instead of `os.getenv()` (always returns None if variable is missing) as this will produce much more useful errors when they are missing.

For example `os.getenv('TRAVIS_COMMIT', None)[:6]` would fail with an error about slicing `None` (not very useful) instead of with an error that the variable is missing (much more useful).

docker/Dockerfile.wallet_server
```diff
@@ -53,5 +54,5 @@ ENV BANDWIDTH_LIMIT=1000000000000000000000000000000000000000000
 ENV MAX_SESSIONS=1000000000
 ENV MAX_SEND=1000000000000000000
 ENV EVENT_LOOP_POLICY=uvloop
-COPY ./lbry/scripts/wallet_server_entrypoint.sh /entrypoint.sh
+COPY ./build/wallet_server_entrypoint.sh /entrypoint.sh
```
eukreign: Is this the correct path? Didn't you rename it to `./docker`?

.gitlab-ci.yml
```diff
@@ -0,0 +1,179 @@
...
+build:mac:
+  extends: .build
+  tags: [macos] # makes gitlab use the mac runner
+  variables:
+    OS: mac
+    GIT_DEPTH: 5
+    VENV: /tmp/gitlab-lbry-sdk-venv
+  before_script:
+#    - brew upgrade python || true
+    - python3 --version | grep -q '^Python 3\.7\.' # dont upgrade python on every run. just make sure we're on the right Python
```
eukreign: how does it "make sure"? does the build fail if grep doesn't find the expression?

docker/Dockerfile.wallet_server
lyoshenka: yes, ty

docker/set_build.py
lyoshenka: good idea. can you use os.getenv and os.environ together? i think i need to make this a bit more verbose to really do what i want

.gitlab-ci.yml
lyoshenka: exactly. the build will fail if any command fails. grep will fail if it cannot find what it's looking for

lbry/lbry/conf.py
lyoshenka: This change is not that substantial. This code is only called in one place, in `cli.py` on line 290:
```
conf = Config.create_from_arguments(args)
```
This is five lines into `main()`. Previous to this change, the `PathNotFoundException` that's raised would bubble all the way up and immediately crash the sdk. So the only people affected by this change are people for whom the sdk does not work at the moment.

I'm happy to move this into a separate PR, but this code is here because the binary does not run on our windows CI runner without it (probably because the runner is a server and does not have a Downloads folder set). Do you really want me to separate out this code just so we can merge it ahead of the rest of this PR?

docker/set_build.py
eukreign: My comment is mostly an FYI, I'd like to drop this script altogether; so as long as it works for now that's probably good enough.

lbry/lbry/conf.py
eukreign: The changed behavior is that before it used to crashed with some clue of why it crashed (can't find download directory) and now it will silently work and perhaps place files where the user doesn't want / won't know to look for.

Maybe it's fine to hide this change in this PR but in general fixing something from previously crashing to silently working (but doing potentially unexpected things) I would consider a substantial change. We've gotten beef from app team and @tzarebczan for lesser changes :-D

lyoshenka: You're right. I'll let Tom know about this

PR #2633, opened 2019-11-22
added `settings_clear` command to reset a setting to its default value

jessopb:


jessopb:
waiting for comment on what settings_clear should return if anything
will test binary to see effects on daemon_settings.yml

lbry/lbry/extras/daemon/Daemon.py
```diff
@@ -1007,6 +1007,24 @@ def jsonrpc_settings_set(self, key, value):
...
+        Options:
+            None
+
+        Returns:
+            (dict) Updated dictionary of daemon settings
+        """
+        with self.conf.update_config() as c:
+            setattr(c, key, NOT_SET)
+            setting = self.conf.settings_dict[key]
+        return {key: setting}
```
eukreign: The get setting part doesn't need to be done inside the `update_config()` context manager, you can just inline it into the `return`:
```
return {key: self.conf.settings_dict[key]}
```

lbry/tests/integration/test_other_commands.py
```diff
@@ -10,3 +10,25 @@ class AddressManagement(CommandTestCase):
...
+        self.daemon.jsonrpc_settings_set('use_upnp', True)
+        settings2 = self.daemon.jsonrpc_settings_get()
+        self.assertEqual(settings2['lbryum_servers'][0], ('server', 50001))
+        self.assertEqual(settings2['use_upnp'], True)
+
+        clearing = self.daemon.jsonrpc_settings_clear('lbryum_servers')
+        self.assertEqual(clearing['lbryum_servers'][0], ('spv1.lbry.com', 50001))
+        settings3 = self.daemon.jsonrpc_settings_get()
+        self.assertNotEqual(settings3['lbryum_servers'][0], ('spv1.lbry.com', 50002))
+        self.assertEqual(settings3['lbryum_servers'][0], ('spv1.lbry.com', 50001))
```
eukreign: There is a bit too much redundancy here, I think you can significantly simplify this and get the same logical coverage with:
```
        default = ('127.0.0.1', 50001)
        self.assertEqual(self.daemon.jsonrpc_settings_get()['lbryum_servers'][0], default)

        setting = self.daemon.jsonrpc_settings_set('lbryum_servers', ['server:50001'])
        self.assertEqual(setting['lbryum_servers'][0], ('server', 50001))

        setting = self.daemon.jsonrpc_settings_clear('lbryum_servers')
        self.assertEqual(setting['lbryum_servers'][0], default)
```

jessopb: Alright.
This brings up a question about what settings_clear should mean.
The above example expects that settings_clear return the lbryum_servers to "default = ('127.0.0.1', 50001) but it actually returns it to the conf coded values [('spv1.lbry.com, 50001), ...].

I think it's fine that clear always return it to the default conf values. Anyone who knows what they're doing will set it to what they need.

PR #2635, opened 2019-11-24
adds wallet server information to `status` command

jessopb:
Adds 
known_servers: int
available_servers: int 
to wallet status.

Also changes
connected_servers: {...}
to 
servers: {
[
host, port, latency, available
],
}

### DIFF
--- a/lbry/lbry/extras/daemon/Components.py
+++ b/lbry/lbry/extras/daemon/Components.py
@@ -107,25 +107,34 @@ class WalletComponent(Component):
         return self.wallet_manager
 
     async def get_status(self):
-        if self.wallet_manager and self.wallet_manager.ledger.network.remote_height:
-            local_height = self.wallet_manager.ledger.local_height_including_downloaded_height
-            remote_height = self.wallet_manager.ledger.network.remote_height
-            best_hash = self.wallet_manager.get_best_blockhash()
-            progress = min(max(math.ceil(float(local_height) / float(remote_height) * 100), 0), 100)
-            return {
-                'connected_servers': [
-                    {
-                        'host': session.server[0],
-                        'port': session.server[1],
-                        'latency': round(session.connection_latency, 2),
-                    } for session in self.wallet_manager.ledger.network.session_pool.sessions
-                    if session and session.available
-                ],
-                'headers_synchronization_progress': progress,
-                'blocks': max(local_height, 0),
-                'blocks_behind': max(remote_height - local_height, 0),
-                'best_blockhash': best_hash,
-            }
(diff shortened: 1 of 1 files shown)


jessopb:
An alternative to this solution is listing all possible "sessions" - those that are available, and those that aren't. 

Then the app ui would be able to tell the user which custom servers are misbehaving.

lbry/lbry/extras/daemon/Components.py
```diff
@@ -107,25 +107,42 @@ def component(self):
...
+                        {
+                            'host': session.server[0],
+                            'port': session.server[1],
+                            'latency': round(session.connection_latency, 2) if session.connection_latency else None,
+                            'availability': session.available,
+                        } for session in sessions
+                    ],
+                    'known_servers': len(sessions),
+                    'available_servers': sum(s.available is True for s in sessions)
+                }
```
eukreign: Can you change the structure of the method to be:
```
result = # create the common response
if self.wallet_manager.ledger.network.remote_height:
   result[...] = # add the extra values available when we have remote_height
return result
```

lbry/lbry/extras/daemon/Components.py
```diff
@@ -107,25 +107,42 @@ def component(self):
...
+                local_height = self.wallet_manager.ledger.local_height_including_downloaded_height
+                remote_height = self.wallet_manager.ledger.network.remote_height
+                best_hash = self.wallet_manager.get_best_blockhash()
+                progress = min(max(math.ceil(float(local_height) / float(remote_height) * 100), 0), 100)
+                return {
+                    'servers': [
+                        {
+                            'host': session.server[0],
+                            'port': session.server[1],
+                            'latency': round(session.connection_latency, 2) if session.connection_latency else None,
```
eukreign: Can this just be `'latency': session.connection_latency,`?

lbry/lbry/extras/daemon/Components.py
```diff
@@ -107,25 +107,42 @@ def component(self):
...
+                            'availability': session.available,
+                        } for session in sessions
+                    ],
+                    'headers_synchronization_progress': progress,
+                    'blocks': max(local_height, 0),
+                    'blocks_behind': max(remote_height - local_height, 0),
+                    'best_blockhash': best_hash,
+                    'known_servers': len(sessions),
+                    'available_servers': sum(s.available is True for s in sessions)
```
eukreign: I don't know what this is doing... `sum()` is a function that adds numbers together... I would imagine that adding booleans together would be undefined behavior?

I think this should be: `len(s for s in sessions if s.available)`

PR #2652, opened 2019-12-04
First working version of alternative trending algorithm inside lbry-sdk

eggplantbren:
As discussed on Slack and elsewhere, this will compute trending scores using a different algorithm from the existing one. I would love for someone to get this running on a server I can connect to. It should give similar (though not identical) results to https://brendonbrewer.keybase.pub/trending.html.

I'm sure more optimisations are possible. My trending scores are going into `trending_mixed` in claims.db, and `trending_group` is zeroed out, so apps can already sort correctly when connected.

A resync is required, and it will start computing trending only on the last 288 blocks (~12 hours). After that, trending is calculated every block in RAM.

**Performance Note:**

In case it's relevant for `sqlite` considerations, every block involves a read that takes about 5-15 seconds, every block divisible by 10 involves a small fast write as well, and every block divisible by 1000 involves a big write taking a minute or two.

`claim.trending_mixed` is now also floats rather than ints.


kauffj:
@eggplantbren please give us just a few days on this, we are going to be shipping an app release that will allow easy configuration of the wallet server on the user side and we want to make sure this is up concurrent with that.

eggplantbren:
All good. I've got some things to iron out with this anyway

PR #2653, opened 2019-12-06
add wallet restart API

shyba:


eukreign:
@shyba what about `wallet_reconnect` instead of `restart`?

PR #2680, opened 2019-12-18
Resubmit of trending PR, sorry

eggplantbren:
As discussed on Slack and elsewhere, this will compute trending scores using a different algorithm from the existing one. I would be happy to explain it in detail if you need me to, but the basic ideas are:

* All claims have a real-valued trending score. Higher is better. Most are positive, but negative values are possible.
* When a new block occurs, all trending scores are multiplied by a constant just below 1. If this were all that happened, all trending scores would decay exponentially towards zero with a half life of 288 blocks (~ 12 hours), leaving rankings unchanged.
* If a claim's total amount changed, an additional "spike" value is added to the trending score, depending on the sign and (weakly) the magnitude of the change. Also, due to the "soften" function, it takes more LBC to bump a whale claim than a minnow one.

In effect, supports (or edits that increase the deposit) cause a spike in the trending score, withdrawal of supports (or edits that decrease the deposit) causes a drop, and all of these events smoothly become less important as time goes on.

To save on database writes, instead of _actually_ multiplying trending scores by the decay constant every block, I magnify the size of spikes, and then renormalise everything every 1000th block, which is the only time there is an expensive database write operation. This saves on disk activity while preserving the ratios of trending scores and hence their ordering, which is all that matters in the end.

Trending scores are calculated in RAM every block (involving a fast read of only claims that are in `recalculate_claim_hashes`), and written to `claims.db` every 10th block. Generally this is a fast write since most claims don't change. My trending scores are going into `claim.trending_mixed`, and `trending_group` is now redundant. I've set it to zero instead of removing it because apps expect it to be there. However, I have removed `trending_global` and `trending_local` columns and the entire `trend` table, since these were intermediate quantities in the old algorithm, and are not needed by apps.

The column `claim.trending_mixed` is now `REAL` rather than `INTEGER`.

If a snapshot would be helpful, you can find one at https://keybase.pub/brendonbrewer/lbry-sdk-stuff (however, this was made with version 0.51.2 + my changes, before I merged with more recent `master` branch changes).


eukreign:
@kauffj is the goal that this replaces the existing ZScore based trending?

@eggplantbren this needs to be rebased on master and some basic styling / code cleanup, for example, constants in Python should be all-caps.

is this code ready for review and merging?

eggplantbren:
If you guys like it, I would be super happy for it to become official. Please try it out first and make sure you think it's sensible.

I can catch it up to `master` whenever you need. Also, is there a Python style guide that you conform to? I've only ever programmed myself or with one other person, so have never learned many things.

eukreign:
There is indeed a Python style guide, first link here: https://www.google.com/search?q=python+style+guide

If you rebase on master there is a stricter pylint rule for entire code base as well. Just run `make lint` to run the pylinter (or `pylint --rcfile=setup.cfg lbry` without make).

eggplantbren:
I've done the rebase.

eggplantbren:
Just merged with upstream/master again and linted trending.py. Hope it's all good. I also removed some old indices that refer to trending_global etc which doesn't exist in this one.

eukreign:
@eggplantbren if possible can you undo your changes to result.py/reader.py and then move your new trending implementation into a new file (leaving trending.py unmodified). then you can change reader.py to import from your new file instead of from trending.py

i'm out sick right now but this is the next item on my plate when i'm back, my plan is to keep both trending algorithms and making it configurable which one to use. it will be easier to rebase your changes when i make trending configurable if your algorithm is in separate file.

let me know if you have any questions.

eukreign:
also, can you squash some of the commits, thanks!

eggplantbren:
I can do all that. Not sure what squashing commits means, though.

PR #2683, opened 2019-12-18
[WIP] pay wallet servers

lyoshenka:
- just pay the current master at the time when the payment is about to be sent. don't worry about if that's fair or not, what if you're connected to multiple servers, etc
- set the max payment to 1 LBC so ppl don't have all their credits taken by a high `daily_fee`
- make sure we're not paying if the wallet server `daily_fee` is 0


lyoshenka:
victor's taking over this work

shyba:
rebased and continued on #2707

PR #2696, opened 2019-12-30
remove bandwidth limit from server

shyba:
It was being ignored (still applying even when we set it disabled) and got removed from upstream.


lbry/wallet/rpc/session.py
```diff
@@ -62,7 +62,6 @@ def __init__(self, session_factory, host=None, port=None, proxy=None,
     async def __aenter__(self):
         transport, self.protocol = await self.create_connection()
         # By default, do not limit outgoing connections
-        self.protocol.bw_limit = 0
```
eukreign: code removed, but not the comment?

PR #2726, opened 2020-01-14
added video file validation and optimization when publishing streams (using `--validate_file` and `--optimize_file` arguments)

### LINKED ISSUE #1988: Optimize videos for quickstart (moov atom)
## The Issue
https://pypi.org/project/qtfaststart/

Add an additional parameter to publish  (default off?) which allows optimizations, such as the one above, to run automatically (this could be one of many). This ensures that the video (h264/quicktime) can start playing as soon as it writes the first bytes. Some video players have trouble otherwise, and will require the whole video to be downloaded. 

Need to consider if we overwrite the user's original file, or create a new one (create + delete?)? What if it's a large file and there's no enough space? 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

BrannonKing:
Please review and ask questions. Use the script to test files.

Fixes #1988

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /.tox
 /.coverage*
 /lbry-venv
+/venv
 
 lbry.egg-info
 __pycache__
@@ -14,4 +15,3 @@ _trial_temp/
 /tests/.coverage.*
 
 /lbry/wallet/bin
-/venv
--- a/lbry/conf.py
+++ b/lbry/conf.py
@@ -1,5 +1,6 @@
 import os
 import re
+import platform
 import sys
 import typing
 import logging
@@ -460,6 +461,13 @@ class BaseConfig:
         self.persisted = ConfigFileAccess(self, config_file_path)
         if self.persisted.upgrade():
             self.persisted.save()
+
+    @property
+    def needs_proactor(self):
+        major, minor, _ = platform.python_version_tuple()
+        if int(major) > 3 or (int(major) == 3 and int(minor) > 7):
+            return False
+        return platform.system() == "Windows"
 
 
 class TranscodeConfig(BaseConfig):
--- a/lbry/conf.py
+++ b/lbry/conf.py
@@ -1,6 +1,5 @@
 import os
 import re
-import platform
 import sys
 import typing
 import logging
@@ -461,13 +460,6 @@ class BaseConfig:
         self.persisted = ConfigFileAccess(self, config_file_path)
         if self.persisted.upgrade():
             self.persisted.save()
-
-    @property
-    def needs_proactor(self):
-        major, minor, _ = platform.python_version_tuple()
-        if int(major) > 3 or (int(major) == 3 and int(minor) > 7):
-            return False
-        return platform.system() == "Windows"
 
 
 class TranscodeConfig(BaseConfig):
--- a/lbry/conf.py
+++ b/lbry/conf.py
@@ -462,7 +462,20 @@ class BaseConfig:
             self.persisted.save()
 
 
-class CLIConfig(BaseConfig):
+class TranscodeConfig(BaseConfig):
+
+    ffmpeg_folder = String('The path to ffmpeg and ffprobe', '')
+    video_encoder = String('FFmpeg codec and parameters for the video encoding. '
+                           'Example: libaom-av1 -crf 25 -b:v 0 -strict experimental',
+                           'libx264 -crf 18 -vf "format=yuv420p"')
+    audio_encoder = String('FFmpeg codec and parameters for the audio encoding. '
+                           'Example: libopus -b:a 128k',
+                           'aac -b:a 192k')
+    volume_filter = String('FFmpeg filter for audio normalization.', '-af loudnorm')
+    volume_analysis_time = Integer('Maximum seconds into the file that we examine audio volume (0 to disable).', '240')
+
+
+class CLIConfig(TranscodeConfig):
 
     api = String('Host name and port for lbrynet daemon API.', 'localhost:5279', metavar='HOST:PORT')
 
--- a/lbry/extras/cli.py
+++ b/lbry/extras/cli.py
@@ -262,6 +262,8 @@ def setup_logging(logger: logging.Logger, args: argparse.Namespace, conf: Config
 
 
 def run_daemon(args: argparse.Namespace, conf: Config):
+    if conf.needs_proactor:
+        asyncio.set_event_loop(asyncio.ProactorEventLoop())
     loop = asyncio.get_event_loop()
     if args.verbose is not None:
         loop.set_debug(True)
--- a/lbry/extras/cli.py
+++ b/lbry/extras/cli.py
@@ -3,6 +3,7 @@ import sys
 import shutil
 import signal
 import pathlib
+import platform
 import json
 import asyncio
 import argparse
@@ -262,7 +263,8 @@ def setup_logging(logger: logging.Logger, args: argparse.Namespace, conf: Config
 
 
 def run_daemon(args: argparse.Namespace, conf: Config):
-    if conf.needs_proactor:
+    if sys.version_info < (3, 8) and platform.system() == "Windows":
+        # TODO: remove after we move to requiring Python 3.8
         asyncio.set_event_loop(asyncio.ProactorEventLoop())
     loop = asyncio.get_event_loop()
     if args.verbose is not None:
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -853,6 +853,7 @@ class Daemon(metaclass=JSONRPCServerType):
         """
 
         connection_code = await self.get_connection_status()
+        ffmpeg_status = await self._video_file_analyzer.status()
 
         response = {
             'installation_id': self.installation_id,
@@ -863,6 +864,7 @@ class Daemon(metaclass=JSONRPCServerType):
                 'code': connection_code,
                 'message': CONNECTION_MESSAGES[connection_code],
             },
+            'ffmpeg_status': ffmpeg_status
         }
         for component in self.component_manager.components:
             status = await component.get_status()
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -44,6 +44,7 @@ from lbry.extras.daemon.componentmanager import ComponentManager
 from lbry.extras.daemon.json_response_encoder import JSONResponseEncoder
 from lbry.extras.daemon import comment_client
 from lbry.extras.daemon.undecorated import undecorated
+from lbry.file_analysis import VideoFileAnalyzer
 from lbry.schema.claim import Claim
 from lbry.schema.url import URL
 
@@ -296,6 +297,7 @@ class Daemon(metaclass=JSONRPCServerType):
 
     def __init__(self, conf: Config, component_manager: typing.Optional[ComponentManager] = None):
         self.conf = conf
+        self._video_file_analyzer = VideoFileAnalyzer(conf)
         self._node_id = None
         self._installation_id = None
         self.session_id = base58.b58encode(utils.generate_id()).decode()
@@ -2808,6 +2810,7 @@ class Daemon(metaclass=JSONRPCServerType):
 
         Usage:
             publish (<name> | --name=<name>) [--bid=<bid>] [--file_path=<file_path>]
+                    [--validate_file] [--optimize_file]
                     [--fee_currency=<fee_currency>] [--fee_amount=<fee_amount>] [--fee_address=<fee_address>]
                     [--title=<title>] [--description=<description>] [--author=<author>]
                     [--tags=<tags>...] [--languages=<languages>...] [--locations=<locations>...]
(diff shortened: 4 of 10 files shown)


kauffj:
If the plan is for lbry-desktop to be the first place this is used by end-users, the app team will need a way to know if this feature is available on the system. Jamming it into `status` would be fine if it's not costly to report.

BrannonKing:
Instructions for using this:

There are two new parameters to the RPC methods `stream_create` and `publish`.
`--validate_file` will force the incoming video file to meet various parameters.
`--optimize_file` will attempt to fix failing file validation.
The parameters are not mutually exclusive. If you both validate and optimize we will error out if the optimization attempt fails to bring things within spec (which should be very rare).

Things we validate:
1. container is of type mp4, webm, or ogg.
2. video codec is h264 or vpx or av1 or theora.
3. audio codec is aac or opus or vorbis or mp3 or flac.
4. audio codec corresponds to video codec
5. bitrate isn't too horrible
6. audio volume isn't too soft

The audio volume check and repair is the most problematic in that it's hard to correct in a single pass.  It would be much better for the user to do that on their own time.

You can use a script to test things without running the full SDK. It's called video_checker.py and lives in the SDK's scripts folder.

There are some options that can be manually configured in the conf file. Here's a cut of the code for that:```
    ffmpeg_folder = String('The path to ffmpeg and ffprobe', '')
    video_encoder = String('FFmpeg codec and parameters for the video encoding. '
                           'Example: libaom-av1 -crf 25 -b:v 0 -strict experimental',
                           'libx264 -crf 18 -vf "format=yuv420p"')
    audio_encoder = String('FFmpeg codec and parameters for the audio encoding. '
                           'Example: libopus -b:a 128k',
                           'aac -b:a 192k')
    volume_filter = String('FFmpeg filter for audio normalization.', '-af loudnorm')
    volume_analysis_time = Integer('Maximum seconds into the file that we examine audio volume (0 to disable).', '240')
```

**Comment by kauffj** [comment]
Is the expectation that this PR will include everything sufficient to add a checkbox to lbry-desktop on publish?

**Comment by BrannonKing** [comment]
@kauffj , yes this will be sufficient for checkbox in the app. I think the question of what to do with the transcoded file is still lingering, but that may linger until there is an event system in the SDK that can let us know when a file is no longer needed.

**Comment by BrannonKing** [comment]
I also want some additional testing on the audio volume handling, to know if we want to ship with that feature on or off.

**Comment by BrannonKing** [comment]
@lyoshenka proposes that we name the parameters "check_browser_compatibility" and "optimize_for_browser_compatibility". Any others for or against?

**Comment by eukreign** [comment]
@kauffj depends, does the app need to know if transcoding capability is available before showing the checkbox option?

**Comment by kauffj** [comment]
Yes: https://github.com/lbryio/lbry-sdk/pull/2726#issuecomment-574737308

**Comment by tzarebczan** [comment]
Did some testing and debugging with Brannon on Windows yesterday and it can't find ffmpeg/ffprobe, even when setting the config path directly. Brannon will take another look.

**Comment by BrannonKing** [comment]
what would be the advantage of adding a separate field to status over simply checking for a minimum SDK version number?

**Comment by tzarebczan** [comment]
You'd still want to know if someone has the ffmpeg binaries in place before having them check the box. 

@kauffj how do you feel about having something in the UI that will guide them to a download, and extract the binaries to the sdk folder (which would be set in the config). Do we want the first version without that? I think it would be easier if we at least told people "put it in this directory" vs having them add it to path or having them configure the ffmpeg path in settings.

**Comment by kauffj** [comment]
I do not think the SDK version number guarantees ability to transcode (will lbry.tv be able to do it? will Android?).

I'm all for good instructions and lots of other things to make this more accessible, I'm just looking, as always, to _ship something as soon as possible_. It seems like simply only showing the box if it's already on the system and callable is faster than figuring out how to bundle, auto-install, handhold users, add configurable setting, etc.  (I want to reiterate, all of these are good things, but we'll start learning more and sooner by shipping.)

**Comment by tzarebczan** [comment]
Works from both path and ffmpeg directory now. I think we should default to checking the data_dir for ffmpeg or set the ffmpeg_folder to data_dir by default. This will allow us to let users know where to drop the ffmpeg executables vs having them config the directory/path. Also, when the app allows the download/extraction of it, that's where we'd drop it. 

Still missing/nice to have: 
- [ ] status to return if ffmpeg is installed/configured properly
- [ ] option to delete file after encoding. Users can redownload if they want, but the video will play from blobs. 

I think a first version is shippable regardless.

**Comment by tzarebczan** [comment]
Other nice to have in the future would be selecting a lower resolution option and maybe even lower bitrate.

File: .gitignore [hunk 1]
```diff
@@ -5,6 +5,7 @@
 /.tox
 /.coverage*
 /lbry-venv
+/venv
```
eukreign: Do we really need both `/lbry-venv` and `/venv`? I think in our docs we say to do `lbry-venv`: https://github.com/lbryio/lbry-sdk/blob/master/INSTALL.md#linuxmac

.gitignore
```diff
@@ -13,5 +14,4 @@ _trial_temp/
 /tests/integration/blockchain/files
 /tests/.coverage.*
 
-/lbry/wallet/bin
-/venv
+/lbry/wallet/bin
```
eukreign: Missing end of line?

lbry/conf.py
```diff
@@ -461,8 +462,28 @@ def set_persisted(self, config_file_path=None):
         if self.persisted.upgrade():
             self.persisted.save()
 
+    @property
+    def needs_proactor(self):
+        major, minor, _ = platform.python_version_tuple()
+        if int(major) > 3 or (int(major) == 3 and int(minor) > 7):
+            return False
+        return platform.system() == "Windows"
```
eukreign: I don't think we need this. See later comment in `cli.py`.

lbry/extras/cli.py
```diff
@@ -262,6 +262,8 @@ def setup_logging(logger: logging.Logger, args: argparse.Namespace, conf: Config
 
 
 def run_daemon(args: argparse.Namespace, conf: Config):
+    if conf.needs_proactor:
+        asyncio.set_event_loop(asyncio.ProactorEventLoop())
```
eukreign: This can just be (instead of adding a non-config related method to `Config`):
```
    if sys.version_info < (3, 8) and platform.system() == "Windows":
        # TODO: remove after we move to requiring Python 3.8
        asyncio.set_event_loop(asyncio.ProactorEventLoop())
```

lbry/extras/daemon/daemon.py
```diff
@@ -3107,6 +3121,8 @@ def jsonrpc_channel_list(self, account_id=None, wallet_id=None, page=None, page_
                     f"Use --allow-duplicate-name flag to override."
                 )
 
+        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
```
eukreign: Should this not even get called if `validate_file` and `optimize_file` are both `False`?

lbry/file_analysis.py
```diff
@@ -0,0 +1,338 @@
...
+        await self._verify_executable("ffprobe")
+        version = await self._verify_executable("ffmpeg")
+        log.debug("Using %s at %s", version.splitlines()[0].split(" Copyright")[0],
+                  shutil.which(f"{self._conf.ffmpeg_folder}ffmpeg"))
+        self._ffmpeg_installed = True
+
+    def __init__(self, conf: TranscodeConfig):
+        self._conf = conf
+        self._available_encoders = ""
+        self._ffmpeg_installed = False
```
eukreign: `__init__` should always go at the top of the class unless there is some very specific reason it shouldn't / can't.

lbry/file_analysis.py
```diff
@@ -0,0 +1,338 @@
...
+log = logging.getLogger(__name__)
+
+
+class VideoFileAnalyzer:
+    @staticmethod
+    def _matches(needles: list, haystack: list):
+        for needle in needles:
+            if needle in haystack:
+                return True
+        return False
```
eukreign: I don't think this method is necessary. You can then just do set operations:

instead of:
```
if not self._matches(container.split(","), ["webm", "mp4", "3gp", "ogg"])
```
do:
```
if not {"webm", "mp4", "3gp", "ogg"}.intersection(container.split(","))
```

setup.py
```diff
@@ -33,6 +33,7 @@
         ],
     },
     install_requires=[
+        'wheel',
```
eukreign: This doesn't make sense here, pretty much all Python packages depend on `wheel`, if this is missing there is something wrong with your setup.

tests/integration/blockchain/test_transcoding.py
```diff
@@ -0,0 +1,148 @@
...
+            with MeasureTime(f"Creating {file_webm.name}"):
+                output, code = await self.analyzer._execute("ffmpeg", command)
+                self.assertEqual(code, 0, output)
+
+        self.should_work = [self.video_file_name, str(file_ogg), str(file_webm)]
+
+    async def test_should_work(self):
+        for should_work_file_name in self.should_work:
+            new_file_name = await self.analyzer.verify_or_repair(True, False, should_work_file_name)
+            self.assertEqual(should_work_file_name, new_file_name)
```
eukreign: Can you unwrap this loop, there are only 3 items in the list. We have a practice of not using loops in tests if at all possible because it makes it impossible to set a breakpoint at a failing part of the test.

tests/integration/blockchain/test_transcoding.py
```diff
@@ -0,0 +1,148 @@
...
+
+    async def test_volume(self):
+        try:
+            self.conf.volume_analysis_time = 200
+            with self.assertRaisesRegex(Exception, "lower than prime"):
+                await self.analyzer.verify_or_repair(True, False, self.video_file_name)
+        finally:
+            self.conf.volume_analysis_time = 0
```
eukreign: Why is this in a `try`/`finally`? Does `volume_analysis_time` need to be set to 0 before the test ends?

tests/integration/blockchain/test_transcoding.py
```diff
@@ -0,0 +1,148 @@
...
+
+        fixed_file = await self.analyzer.verify_or_repair(True, True, file_name)
+        pathlib.Path(fixed_file).unlink()
+
+    async def test_extension_choice(self):
+
+        for file_name in self.should_work:
+            scan_data = await self.analyzer._get_scan_data(True, file_name)
+            extension = self.analyzer._get_best_container_extension(scan_data, "")
+            self.assertEqual(extension, pathlib.Path(file_name).suffix[1:])
```
eukreign: Can you unwind this loop.

.gitignore
BrannonKing: Pycharm uses venv out of the box. Besides, it's already added in master by somebody else.

eukreign: You moved it around then, which brought it to my attention? Probably unnecessary change for this PR then?

tests/integration/blockchain/test_transcoding.py
BrannonKing: Yes, the test video file from google fails the audio volume analysis (so it's disabled for the other tests).

eukreign: I don't understand, are you saying the tests leak information to each other?

.gitignore
eukreign: I guess it's fine, your cleanup is better.

setup.py
BrannonKing: I'm not the only one to complain about this issue.

BrannonKing: But I don't really know what to do to debug it.

lbry/extras/daemon/daemon.py
BrannonKing: That case is handled at the top of verify_or_repair.

eukreign: The `await` carries a non-trivial penalty though (the function will stop executing and go back onto the schedule and will not pickup again until much later) and especially right at the beginning of the function. In the cases where both are false it doesn't make sense to suspend the execution of the function just to run an if statement inside of `verify_or_repair`.

for example, lbry.tv and spee.ch call publish a lot so it's important not to introduce unnecessary performance penalties.

lbry/conf.py
BrannonKing: Fixed and pushed.

lbry/extras/cli.py
BrannonKing: Fixed and pushed.

lbry/file_analysis.py
BrannonKing: I like this intersection approach. Fixed and pushed.

tests/integration/blockchain/test_transcoding.py
BrannonKing: Fixed and pushed.

tests/integration/blockchain/test_transcoding.py
BrannonKing: Fixed and pushed.

BrannonKing:
I've added ffmpeg_status to the status() method on the daemon. It includes fields: availabe, which, and analyze_audio_volume.

PR #2727, opened 2020-01-14
Add prometheus metrics

lyoshenka:


lbry/conf.py
```diff
@@ -568,6 +569,8 @@ class Config(CLIConfig):
         ('spv7.lbry.com', 50001),
         ('spv8.lbry.com', 50001),
         ('spv9.lbry.com', 50001),
+        ('victor.lbry.tech', 50001),
+        ('spv9.lbry.com', 50002),
```
eukreign: Does this replace the other spv9? Seems confusing to have two spv9s?

lbry/wallet/server/block_processor.py
```diff
@@ -749,7 +749,7 @@ def advance_txs(self, height, txs, header):
         timer = self.timer.sub_timers['advance_blocks']
         undo = timer.run(super().advance_txs, height, txs, header, timer_name='super().advance_txs')
         timer.run(self.sql.advance_txs, height, txs, header, self.daemon.cached_height(), forward_timer=True)
-        if (height % 10000 == 0 or not self.db.first_sync) and self.logger.isEnabledFor(20):
+        if (height % 10000 == 0 or not self.db.first_sync) and self.logger.isEnabledFor(10):
```
eukreign: It's better to use the constants from logging module, so that it's clearer what 10 means.

Issue #2738, opened 2020-01-23
Add these prometheus metrics


lyoshenka:
- [x] number of db queries total
- [x] number of queries that are currently running (start counting right before you await it, stop right after)
- [x] how long do the queries take
- [x] what block we're on
- [x] server version, min/max versions it supports
- [x] Docker image hash
- [ ] how many address subscriptions we have
- [ ] how many unique addresses are subscribe
- [x] how many connections there are
- [ ] how much data we're sending
- [x] how many responses/notifications we're sending
- [x] how many lbrycrd calls
- [x] how long the lbrycrd calls take
- [x] how many clients are running which version
- [ ] aws region

lyoshenka:
didnt add everything but what we have is good enough

PR #2739, opened 2020-01-23
Include docker build info in wallet server image

lyoshenka:
also add some more metrics to prometheus


lyoshenka:
related to #2738

PR #2747, opened 2020-01-27
new trending algorithm contributed by Brendon Brewer, to enable on wallet server set `TRENDING_ALGORITHMS` environment variable on wallet server to `ar`

eggplantbren:
As requested by Lex, this version has many fewer commits, sections off my algorithm into its own file `trending_ar.py`, and doesn't affect much else. If you were to run this as-is, it would use the normal trending algorithm. If you want to use mine, you just need to import from `trending_ar` instead of `trending` at the top of `writer.py`.

### DIFF
--- a/lbry/wallet/server/db/trending/__init__.py
+++ b/lbry/wallet/server/db/trending/__init__.py
@@ -1,6 +1,8 @@
 from . import zscore
+from . import ar
 
 
 TRENDING_ALGORITHMS = {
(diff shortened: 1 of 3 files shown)


eggplantbren:
The write at every 1000th block is taking absurdly long and I have no idea why. I give up.

eggplantbren:
I've sped up the 1000-block writes and I hope they are tolerable now.

PR #2754, opened 2020-01-29
add duration to claim search params

jessopb:
Tests fail when they should, and work when they should.
The index is a best guess.

### DIFF
--- a/tests/integration/blockchain/test_claim_commands.py
+++ b/tests/integration/blockchain/test_claim_commands.py
@@ -73,6 +73,12 @@ class ClaimSearchCommand(ClaimTestCase):
                 (result['txid'], result['claim_id'])
             )
 
+    async def assertFindsNoClaims(self, **kwargs):
+        kwargs.setdefault('order_by', ['height', '^name'])
+        results = await self.claim_search(**kwargs)
+        print(results)
+        self.assertEqual(0, len(results))
+
     async def test_basic_claim_search(self):
         await self.create_channel()
         channel_txo = self.channel['outputs'][0]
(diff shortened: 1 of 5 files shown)


tests/integration/blockchain/test_claim_commands.py
```diff
@@ -328,6 +334,13 @@ class ClaimSearchCommand(ClaimTestCase):
         await self.assertFindsClaims([image], media_types=['image/png'])
         await self.assertFindsClaims([image, video], media_types=['video/mp4', 'image/png'])
 
+        # duration
+        await self.assertFindsClaim(video, duration=f'>{14}')
+        await self.assertFindsClaim(video, duration=f'<{16}')
+        await self.assertFindsClaim(video, duration=15)
+        await self.assertFindsNoClaims(duration=f'>{100}')
+        await self.assertFindsNoClaims(duration=f'<{14}')
+
```
eukreign: What's the `{` and `}` for here?

tests/integration/blockchain/test_claim_commands.py
```diff
@@ -73,6 +73,12 @@ class ClaimSearchCommand(ClaimTestCase):
                 (result['txid'], result['claim_id'])
             )
 
+    async def assertFindsNoClaims(self, **kwargs):
+        kwargs.setdefault('order_by', ['height', '^name'])
+        results = await self.claim_search(**kwargs)
+        print(results)
+        self.assertEqual(0, len(results))
+
```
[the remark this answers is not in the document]
jessopb: Sorry - missed my print. 
I thought it was important that the logic for duration not return results if duration asked for was greater than that which existed.  I can remove it if you like.  I just added it because I wanted to make sure I wasn't getting false positives.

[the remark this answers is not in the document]
eukreign: You just do: `await self.assertFindsClaims([], ...)`

tests/integration/blockchain/test_claim_commands.py
```diff
@@ -328,6 +333,13 @@ class ClaimSearchCommand(ClaimTestCase):
         await self.assertFindsClaims([image], media_types=['image/png'])
         await self.assertFindsClaims([image, video], media_types=['video/mp4', 'image/png'])
 
+        # duration
+        await self.assertFindsClaim(video, duration=f'>14')
+        await self.assertFindsClaim(video, duration=f'<16')
+        await self.assertFindsClaim(video, duration=15)
+        await self.assertFindsNoClaims(duration=f'>100')
+        await self.assertFindsNoClaims(duration=f'<14')
+
```
eukreign: Don't need the `f`

tests/integration/blockchain/test_claim_commands.py
eukreign: Don't need the `f` either.

PR #2765, opened 2020-02-05
fixed some issues with AR trending

eggplantbren:
* The `trending_ar.log` file is only created if `install` is called
* Skip trending calculations except for most recent 5*HALF_LIFE = 670 blocks to speed up initial sync (this was always the intent but was broken).
* Use claim_hash instead of claim_id to identify claims
* No claim_id index required.

The changes are quite simple but I am still testing them.


eggplantbren:
All seems okay with my test. When resyncing from scratch it did crash once due to `std::bad_alloc`, but that had nothing to do with trending, was probably just my computer running out of RAM

PR #2772, opened 2020-02-07
Add `is_fully_reflected` to `file_list` response

jackrobison:


lyoshenka:
@jackrobison does this need to be reviewed? plz request a review (or request from @lbryio/protocools if you're not sure whom to ask), or just merge it if no review is necessary

tzarebczan:
Can we get progress for this also? Would be nice to show the user a percentage. Also need to improve the failure case and make sure users can retry reflection - how would they know to retry it if it's not 'stuck'?

PR #2773, opened 2020-02-07
fix video analysis error

BrannonKing:
found by @tzarebczan


lbry/wallet/orchstr8/node.py
```diff
@@ -228,6 +228,7 @@ def pipe_data_received(self, fd, data):
 
     def process_exited(self):
         self.stopped.set()
+        self.ready.set()
```
eukreign: Nice!

PR #2777, opened 2020-02-09
`claim_list` argument `--claim_type` can now be repeated to pass a list of claim_types

eukreign:


seanyesmunt:
@eukreign can this be added to other calls too? https://github.com/lbryio/lbry-desktop/pull/3712 is blocked because it isn't supported.

seanyesmunt:
Nevermind. Wasn't on the latest release.

PR #2792, opened 2020-02-12
torrent support

shyba:


lbry/extras/daemon/json_response_encoder.py
```diff
@@ -126,6 +127,8 @@ def default(self, obj):  # pylint: disable=method-hidden,arguments-differ,too-ma
             return self.encode_wallet(obj)
         if isinstance(obj, ManagedStream):
             return self.encode_file(obj)
+        if isinstance(obj, TorrentSource):
+            return self.encode_file(obj)
```
eukreign: `isinstance()` takes a tuple as second argument, so it could just be: `isinstance(obj, (ManagedStream, TorrentSource))`

lbry/extras/daemon/json_response_encoder.py
```diff
@@ -301,7 +310,7 @@ def encode_file(self, managed_stream):
             'height': tx_height,
             'confirmations': (best_height + 1) - tx_height if tx_height > 0 else tx_height,
             'timestamp': self.ledger.headers[tx_height]['timestamp'] if 0 < tx_height <= best_height else None,
-            'is_fully_reflected': managed_stream.is_fully_reflected
+            'is_fully_reflected': managed_stream.is_fully_reflected if is_stream else False
         }
```
eukreign: Instead of all of those inline ifs can you split this out into:
```
result = { all of the stuff that's the same }
if is_stream:
  result.update({
    all of the stream stuff
  })
if output_exists:
  result.update({
    all of the output exists stuff
  })
...
return result
```

lbry/file/file_manager.py
```diff
@@ -0,0 +1,286 @@
...
+                        None if not stream else stream.sd_hash,
+                        None if not stream else stream.downloader.time_to_descriptor,
+                        None if not (stream and stream.descriptor) else stream.descriptor.blobs[0].blob_hash,
+                        None if not (stream and stream.descriptor) else stream.descriptor.blobs[0].length,
+                        None if not stream else stream.downloader.time_to_first_bytes,
+                        None if not error else error.__class__.__name__,
+                        None if not error else str(error),
+                        None if not server else f"{server[0]}:{server[1]}"
+                    )
+                )
```
eukreign: there's gotta be a better way to do above

shyba:
2930 supersedes this one

PR #2793, opened 2020-02-12
don't validate and repair files that aren't videos; add maximum video bit rate setting

BrannonKing:
Also fix ogg vs ogv mismatch.

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -3142,11 +3142,8 @@ class Daemon(metaclass=JSONRPCServerType):
                     f"Use --allow-duplicate-name flag to override."
                 )
 
-        try:
-            file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
-        except ValueError:
-            pass  # it's not a video file
-
+        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file,
+                                                                     file_path, ignore_non_video=True)
         claim = Claim()
         claim.stream.update(file_path=file_path, sd_hash='0' * 96, **kwargs)
         tx = await Transaction.claim_create(
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -3142,8 +3142,9 @@ class Daemon(metaclass=JSONRPCServerType):
                     f"Use --allow-duplicate-name flag to override."
                 )
 
-        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file,
-                                                                     file_path, ignore_non_video=True)
+        file_path = await self._video_file_analyzer.verify_or_repair(
+            validate_file, optimize_file, file_path, ignore_non_video=True
+        )
         claim = Claim()
         claim.stream.update(file_path=file_path, sd_hash='0' * 96, **kwargs)
         tx = await Transaction.claim_create(
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -3142,7 +3142,10 @@ class Daemon(metaclass=JSONRPCServerType):
                     f"Use --allow-duplicate-name flag to override."
                 )
 
-        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
+        try:
+            file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
+        except ValueError:
+            pass  # it's not a video file
 
         claim = Claim()
         claim.stream.update(file_path=file_path, sd_hash='0' * 96, **kwargs)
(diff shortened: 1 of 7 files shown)


lyoshenka:
@BrannonKing is this ready for review? if yes, please assign to @lbryio/protocools

lbry/extras/daemon/daemon.py
```diff
@@ -3123,7 +3123,10 @@ def jsonrpc_channel_list(self, account_id=None, wallet_id=None, page=None, page_
                     f"Use --allow-duplicate-name flag to override."
                 )
 
-        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
+        try:
+            file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
+        except ValueError:
+            pass  # it's not a video file
```
eukreign: This isn't a very ergonomic API if it throws an exception under arguably normal circumstances (a file is not a video). Can this be handled internally in `verify_or_repair`? You can just return a  boolean from `verify_or_repair` if anything was actully done (or False if nothing was done / it was not a video, etc).

lbry/schema/mime_types.py
```diff
@@ -162,6 +163,7 @@
 
 
 def guess_media_type(path):
+    # should we be using "file --mime-type -b $filename" on linux?
```
eukreign: Can you ask this in the PR discussion or in a new issue where it can be answered / discussed?

lbry/extras/daemon/daemon.py
BrannonKing: If the mime type was determined before this point that would help a lot. As it is, the method is used both here and in the check_video script. However, the check_video script handles this exception in a much different fashion.

eukreign: mime type should be available after these lines:

https://github.com/lbryio/lbry-sdk/blob/4b87cb45eef1c5b661c96462febb34a50ddd3523/lbry/extras/daemon/daemon.py#L3127-L3128

eukreign: https://github.com/lbryio/lbry-sdk/blob/4b87cb45eef1c5b661c96462febb34a50ddd3523/lbry/schema/claim.py#L239-L245

eukreign: but... `verify_or_repair` shouldn't rely on that to the point of blowing up. I think in almost all circumstances calling `verify_or_repair` shouldn't throw an error unless `validate_file` is true and the file is not valid.

BrannonKing: So if I add a boolean parameter for "ignore non-video files" that would be better than the try/except?

lbry/schema/mime_types.py
BrannonKing: I've since realized that there is a built-in mimetypes module that we should probably use instead.

eukreign: yeah, that's how it worked originally but it was missing types and so we made a our own custom lookup

lbry/extras/daemon/daemon.py
eukreign: yes, sounds good

lbry/testcase.py
```diff
@@ -220,6 +220,9 @@ def __init__(self, *args, **kwargs):
         self.account: Optional[Account] = None
 
     async def asyncSetUp(self):
+        if sys.version_info < (3, 8):
+            # hide warning about TaskWakeupMethWrapper, see bugs.python.org/issue38608
+            asyncio.get_running_loop().set_debug(False)
```
eukreign: This is outside the scope of the PR but also we don't want to turn off debug made. In fact, we explicitly turn it on in the same file so it doesn't make sense to then explicitly turn it off: https://github.com/lbryio/lbry-sdk/blob/75a1cc0d3305743b2a8a233d1169d31f5954cace/lbry/testcase.py#L125

lbry/extras/daemon/daemon.py
```diff
@@ -3142,8 +3142,8 @@ def jsonrpc_channel_list(self, account_id=None, wallet_id=None, page=None, page_
                     f"Use --allow-duplicate-name flag to override."
                 )
 
-        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file, file_path)
-
+        file_path = await self._video_file_analyzer.verify_or_repair(validate_file, optimize_file,
+                                                                     file_path, ignore_non_video=True)
```
eukreign: We try to avoid the "pan handle" style. Can you move the arguments to be on a new line:

```
        file_path = await self._video_file_analyzer.verify_or_repair(
            validate_file, optimize_file, file_path, ignore_non_video=True
        )
```

lbry/extras/daemon/daemon.py
BrannonKing: Done

lbry/testcase.py
BrannonKing: Fixed

lbry/extras/daemon/daemon.py
BrannonKing: Fixed

PR #2805, opened 2020-02-13
fix claim_search rpc and cli behaving differently

jackrobison:
don't autopopulate order_by in the server, instead populate an empty list in the client (which is what cli already did)

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -2179,7 +2179,12 @@ class Daemon(metaclass=JSONRPCServerType):
         return paginate_rows(claims, claim_count, page, page_size, claim_type=claim_type, resolve=resolve)
 
     @requires(WALLET_COMPONENT)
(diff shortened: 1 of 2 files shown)


lyoshenka:
@jackrobison does this need to be reviewed? plz request a review (or request from @lbryio/protocools if you're not sure whom to ask)

eukreign:
i reviewed this in slack, will copy/paste it here as well

eukreign:
@jackrobison after reflecting on this more, i think this PR should really just be to remove the default order_by, so just removing these two lines and fixing any tests that may fail expecting the default to be used (not sure if there are any tests that expect the default...):
https://github.com/lbryio/lbry-sdk/blob/0d08858dfb5178056fc1ad4c5afc3d4c1bfff511/lbry/wallet/server/db/reader.py#L421-L422

eukreign:
done in https://github.com/lbryio/lbry-sdk/pull/2810

PR #2808, opened 2020-02-14
Enable resolve proof checks

lyoshenka:
This PR implements the current method of checking resolve proofs. It doesn't actually turn this on. I propose:

- turn on resolve proof checking for every resolve we do. raise an exception if any proof fails
- add `--skip-resolve-proofs` to `resolve` and any other RPC call that does resolves to skip proof checking and improve performance if the wallet server is trusted
- add `SKIP_RESOLVE_PROOFS` env var to disable resolve proofs completely (same use case as above but you don't have to pass the flag every time)

### DIFF
--- a/lbry/wallet/claim_proofs.py
+++ b/lbry/wallet/claim_proofs.py
@@ -15,8 +15,18 @@ def get_hash_for_outpoint(txhash, nout, height_of_last_takeover):
     )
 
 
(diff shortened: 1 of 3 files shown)


eukreign:
closing this for now as it will likely need significant work to integrate after the big SDK refactor, will re-open / review it then and re-integrate then

PR #2811, opened 2020-02-17
wallet server indexes

eukreign:


tzarebczan:
LGTM, thanks so much @eukreign ! @jackrobison , do you think we can have these added to existing wallet servers ? I'll test locally first.

PR #2815, opened 2020-02-18
apply share_usage_data as its set, without restarting

shyba:


lbry/extras/daemon/analytics.py
```diff
@@ -110,28 +110,33 @@ def __init__(self, conf: Config, installation_id: str, session_id: str):
...
 
     async def run(self):
         while True:
-            await self._send_heartbeat()
+            if self.enabled:
+                await self._send_heartbeat()
             await asyncio.sleep(1800)
-            self.external_ip = await utils.get_external_ip()
+            if self.enabled:
+                self.external_ip = await utils.get_external_ip()
```
eukreign: if the process was started in not enabled state and thus external ip address was not set but then later it was enabled, can `self._send_heartbeat()` work without an external_ip set? it appears above that first heart beat is sent and only after it gets the ip again...

lbry/extras/daemon/loggly_handler.py
```diff
@@ -47,6 +47,11 @@ def __init__(self, loggly_token: str, fqdn=False, localname=None, facility=None,
         )
         self._loop = asyncio.get_event_loop()
         self._session = aiohttp.ClientSession()
+        self._toggle = feature_toggle
+
+    @property
+    def enabled(self):
+        return self._toggle is None or (self._toggle and self._toggle())
```
eukreign: could this just be: `return self._toggle and self._toggle()`?

Issue #2799, opened 2020-02-13
Force signed comments 


tzarebczan:
See https://github.com/lbryio/comment-server/issues/34

tzarebczan:
This was completed.

PR #2816, opened 2020-02-18
removes ability to comment anonymously

osilkin98:
Changes the `channel_name`/`channel_id` parameters on the `comment_create` command from being optional to required.

Addresses #2799

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -4584,7 +4584,7 @@ class Daemon(metaclass=JSONRPCServerType):
         Usage:
             comment_create  (<comment> | --comment=<comment>)
                             (<claim_id> | --claim_id=<claim_id> | --parent_id=<parent_id>)
-                            [--channel_id=<channel_id>] [--channel_name=<channel_name>]
+                            (--channel_id=<channel_id> | --channel_name=<channel_name>)
                             [--channel_account_id=<channel_account_id>...] [--wallet_id=<wallet_id>]
 
         Options:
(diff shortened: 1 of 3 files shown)


lbry/extras/daemon/daemon.py
```diff
@@ -4592,23 +4592,25 @@ def jsonrpc_routing_table_get(self):
             }
         """
         wallet = self.wallet_manager.get_wallet_or_default(wallet_id)
+        channel = await self.get_channel_or_none(
```
eukreign: why not `get_channel_or_error()`?

tests/integration/other/test_comment_commands.py
```diff
@@ -355,7 +328,8 @@ class CommentCommands(CommandTestCase):
         self.assertIs(comment_list['page'], 1)
         self.assertIs(comment_list['total_items'], 28)
```
eukreign: if you want to check that comment_list has a set of keys, use a set instead of a loop:
```
self.assertEqual(
    {'items', 'page', 'page_size', 'has_hidden_comments', 'total_items', 'total_pages'}, set(claim_list)
)
```

also, `self.assertIs()` is for checking instance equality, here you want to use regular equality check with `assertEqual` instead

osilkin98:
@eukreign This PR is about removing anonymous comments, not changing the existing tests.

tests/integration/other/test_comment_commands.py
osilkin98: That's super useful, I had no idea that set(dict) returned the set of its keys.

eukreign:
@osilkin98 okay, it's fine to fix `assertIs`->`assertEqual` in another PR but I had two other comments that I don't think you've addressed

osilkin98:
I'm only seeing one other comment, where's the other one?

osilkin98:
Never mind.

osilkin98:
They've been addressed

osilkin98:
I've looked over your comments & made the relevant changes.

eukreign:
@osilkin98 can you rebase this branch it without the merge commit from master and then i think it's ready to be merged into master

osilkin98:
@eukreign Got it

PR #2817, opened 2020-02-18
AR Trending with time delay

eggplantbren:
As discussed in the Slack group and at lbry://yata, here's the autoregressive trending algorithm with a time delay applied to large positive LBC changes (insofar as they effect trending scores). I'm finding it quite pleasant (it's running on 199.192.31.122:50001).

I won't be offended if you don't want this, as it was primarily an experiment, but I thought I'd open the PR anyway because it's pretty much an extra harmless .py file.


eukreign:
@eggplantbren is it possible to fix the original ar algorithm to do this instead? i'm very wary of adding code to the repo that won't be used since it just leads to bitrot

eggplantbren:
Best to just leave it then. This was an experiment.

eukreign:
okay, please continue experimenting and if you come up with good improvements for ar algo just make them to the existing ar algorithm

PR #2842, opened 2020-03-05
revert LD_LIBRARY_PATH before calling ffmpeg

### LINKED ISSUE #2838: ffmpeg not found when using linux binary / folder issue
Seems to work fine in venv/running from source. Works on Windows too, have not tried Mac. 

If using ffmpeg_folder, the / is not concatenated at the end of the directory so you see something like this: ```[Errno 2] No such file or directory: '/usr/binffprobe': '/usr/binffprobe'```

BrannonKing:
PyInstaller modifies the LD_LIBRARY_PATH variable. This affects the execution of files external to the package. It was affecting the execution of ffmpeg but is handled here as suggested in the documentation at https://pyinstaller.readthedocs.io/en/stable/runtime-information.html#ld-library-path-libpath-considerations

Fixes #2838

### DIFF
--- a/lbry/file_analysis.py
+++ b/lbry/file_analysis.py
@@ -7,6 +7,8 @@ import re
 import shlex
 import shutil
 import platform
+
+import lbry.utils
 from lbry.conf import TranscodeConfig
 
 log = logging.getLogger(__name__)
@@ -28,7 +30,9 @@ class VideoFileAnalyzer:
         self._which = None
         self._checked_ffmpeg = False
         self._env_copy = dict(os.environ)
-        self._replace_or_pop_env('LD_LIBRARY_PATH')
+        if lbry.utils.is_running_from_bundle():
+            # handle the situation where PyInstaller overrides our runtime environment:
+            self._replace_or_pop_env('LD_LIBRARY_PATH')
 
     async def _execute(self, command, arguments):
         if DISABLED:
--- a/lbry/file_analysis.py
+++ b/lbry/file_analysis.py
@@ -37,10 +37,10 @@ class VideoFileAnalyzer:
         try:
             version, code = await self._execute(name, "-version")
         except Exception as e:
-            log.warning("Unable to run %s, but it was requested. Message: %s", name, str(e))
             code = -1
-            version = ""
+            version = str(e)
         if code != 0 or not version.startswith(name):
+            log.warning("Unable to run %s, but it was requested. Code: %d; Message: %s", name, code, version)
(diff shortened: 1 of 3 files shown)


lbry/file_analysis.py
```diff
@@ -14,18 +14,26 @@
 
 class VideoFileAnalyzer:
 
+    def _replace_or_pop_env(self, variable):
+        if variable + '_ORIG' in self._env_copy:
+            self._env_copy[variable] = self._env_copy[variable + '_ORIG']
+        else:
+            self._env_copy.pop(variable, None)
```
eukreign: What's the reason for poping the variable if there is no `_ORIG`? This seems like it would give you a different behavior during development vs. running a built binary since in development no LD_LIBRARY_PATH is forwarded at all but when run in binary you actually get the same LD_LIBRARY_PATH that would have been forwarded during development...

BrannonKing: That's what their instructions said to do. If there is no `_ORIG` value then there was not a LD_LIBRARY_PATH value existing before we started running. So if there still isn't one then the pop is a no-op, if there is one then it must have been added by the pyinstaller bootstrapper.

eukreign: There will be no `_ORIG` if lbrynet was run outside of pyinstaller boostrapper but yet there will still be a valid `LD_LIBRARY_PATH` that should be passed to the executor.

lbry/conf.py
```diff
@@ -461,6 +461,11 @@ def set_persisted(self, config_file_path=None):
         if self.persisted.upgrade():
             self.persisted.save()
 
+    @property
+    def is_running_from_bundle(self):
+        # see https://pyinstaller.readthedocs.io/en/stable/runtime-information.html
+        return getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS')
+
```
eukreign: This isn't a configurable setting, please move this to `lbry.utils`.

lbry/file_analysis.py
BrannonKing: Fixed.

lbry/conf.py
BrannonKing: Done. However, some further thoughts: in my mind conf could easily include both user configuration and platform configuration. I'm fine with them separate also. It seems that platform configuration is generally missing or unorganized the lbry-SDK. It should be there in OOP form, a platform service of sorts, easily testible, stubible, mockable, etc.

PR #2844, opened 2020-03-06
fix UDP being broken by ProactorEventLoop on windows

jackrobison:
fixes https://github.com/lbryio/lbry-sdk/issues/2843

### DIFF
--- a/lbry/extras/cli.py
+++ b/lbry/extras/cli.py
@@ -3,7 +3,6 @@ import sys
 import shutil
 import signal
 import pathlib
-import platform
 import json
 import asyncio
 import argparse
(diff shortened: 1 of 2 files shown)


BrannonKing:
This is not a viable solution, but Python 3.8 is.

jackrobison:
I think the options are:

a) drop ProactorEventLoop, break ffmpeg, fix p2p (or a variant of this, perhaps switch which is used with a setting)
b) use a ProactorEventLoop for ffmpeg and SelectorEventLoop as the default
c) use 3.8

I don't know how/if we can do the 3.8 route, afaik we can't do it with pyinstaller.

eukreign:
@jackrobison @BrannonKing can we explore one of these two options:

1) switch to 3.8, this seems like a good enough reason (are there any hard blockers? do all the blockers have reasonable work arounds?)

2) don't use `asyncio.create_subprocess_exec` and instead use the blocking subprocess exec API and then run that using the main loops asyncio.run_in_executor. (the current implementation, especially with this PR, is totally insane: main event loop -> starts a thread pool -> spawning a thread -> which creates an new event loop -> which then creates a new thread pool -> spawning a thread -> which creates an new event loop (or reuses the previously created event loop from the outer thread pool? wtf?) -> which then... i give up)

BrannonKing:
Superseded by https://github.com/lbryio/lbry-sdk/pull/2861

PR #2846, opened 2020-03-07
added `is_spent` attribute to transaction outputs

### LINKED ISSUE #2841: transaction list - add property if claim or tip is active
To support https://github.com/lbryio/lbry-desktop/issues/3770, we need a way to identify which tips/supports can be unlocked, or what claims can be abandoned. Today we load all supports and all claims, and match up against transaction list.

eukreign:
fixes #2841


tzarebczan:
LGTM
ship

PR #2854, opened 2020-03-11
Adds tracemalloc api for memory troubleshooting

shyba:


shyba:
added separated commands instead of `set true/false`

tests/integration/other/test_other_commands.py
```diff
@@ -38,3 +38,19 @@ class SettingsManagement(CommandTestCase):
         self.assertTrue(self.daemon.analytics_manager.enabled)
         self.assertTrue(loggly.enabled)
         self.daemon.jsonrpc_settings_set('share_usage_data', False)
+
+
+class TroubleshootingCommands(CommandTestCase):
+    async def test_tracemalloc_commands(self):
+        self.assertFalse(self.daemon.jsonrpc_tracemalloc_disable())
+        self.assertTrue(self.daemon.jsonrpc_tracemalloc_enable())
```
eukreign: where do you disable tracemalloc after the test runs?

shyba: nowhere, pushed a fix. nice catch!

PR #2857, opened 2020-03-11
Trending algorithm with time delay and variable decay rate

eggplantbren:
This is the one I've been working on in #tech-trending on Slack. It's similar to the AR algorithm, but with the following modifications, inspired by Tom's idea that in the future large supports should require time locking, but able to be implemented before that's a thing.

* Large, positive LBC changes take time before they affect the trending score. Specificially, the delay is equal to LBC^0.4 blocks, with a maximum delay of 200 blocks.

* Very high trending scores decay at a faster rate than lower scores.

The time delays look like this - people might wonder about it:

```
10 LBC:   2 blocks
100 LBC:  6 blocks
1000 LBC: 15 blocks
10000 LBC: 39 blocks
100000 LBC: 100 blocks
>= 565000 LBC: 200 blocks (edited) 
```

Jeremy and Tom are keen to A/B test this on lbry.tv.

Ultimately, the degree to which trending is full of "whales" depends on the shape of the tails of the LBC wealth distribution. However, this reduces the effect somewhat. I've done some analytic work on it and under certain approximations, time delays or locking has an equivalent effect to changing the Pareto slope of the tails.

Anecdotally, when I use this, the top trending items are whales, but you don't have to scroll very far down before you start seeing more modest supported claims, and ones that are merely being watched.

### DIFF
--- a/lbry/wallet/server/db/trending/__init__.py
+++ b/lbry/wallet/server/db/trending/__init__.py
@@ -1,8 +1,9 @@
 from . import zscore
 from . import ar
-
+from . import variable_decay
 
 TRENDING_ALGORITHMS = {
     'zscore': zscore,
-    'ar': ar
+    'ar': ar,
+    'variable_decay': variable_decay
 }
--- a/lbry/wallet/server/db/trending/variable_decay.py
+++ b/lbry/wallet/server/db/trending/variable_decay.py
@@ -1,0 +1,431 @@
+"""
+Delayed AR with variable decay rate.
+
+The spike height function is also simpler.
+"""
+
+import copy
+import time
+import apsw
+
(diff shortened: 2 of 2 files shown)


kauffj:
@eukreign @jackrobison or @andybeletsky, once lbry.tv is running SDK 0.63.0+, @seanyesmunt will deploy a lbry.tv update that stores `trending_algorithm` against analytics (whoever does that, please ping him once this is done).

Once Sean does that, we can run a mix of trending algorithms simultaneously against the same data, which will make for more useful comparisons.

For whoever does the SDK updating, feel free to just put the SDKs in a (balanced) mix of modes even prior to Sean's update going live.

PR #2859, opened 2020-03-16
add config option to skip transaction verification

lyoshenka:


lbry/wallet/ledger.py
```diff
@@ -589,6 +589,8 @@ def process_status_update(self, update):
 
     async def maybe_verify_transaction(self, tx, remote_height):
         tx.height = remote_height
+        if self.config.get("skip_transaction_verification"):
+            return
```
eukreign: Without setting `txo.position` this may lead to difficult to debug problems down the road where it's expected that `txo.position` has a valid value.

PR #2880, opened 2020-03-23
Handle MKV files

### LINKED ISSUE #2868: force mp4/use ffmpeg metadata if file does not require transcoding, but missing metadata
For MKV files (and possibly others), you get the below. Which means we don't populate metadata for it, i.e. height and width. If we forced mp4, it should, or if we read the data from ffmpeg instead. 

```
2020-03-19 17:45:51,686 ERROR    lbry.schema.attrs:67: Could not extract height from file metadata.
  File "c:\users\thoma\documents\lbry-sdk\lbry\schema\attrs.py", line 65, in _extract
    setattr(self, field, file_metadata.getValues(field)[0])
IndexError: list index out of range
2020-03-19 17:45:51,694 ERROR    lbry.schema.attrs:67: Could not extract width from file metadata.
  File "c:\users\thoma\documents\lbry-sdk\lbry\schema\attrs.py", line 65, in _extract
    setattr(self, field, file_metadata.getValues(field)[0])
IndexError: list index out of range```

### PR DESCRIPTION
**Pull request by BrannonKing**
Fixes #2868

1. Verifies that MKV files are WebM compatible
2. Reuses the height/width/duration data from ffmpeg if possible.

### DIFF
--- a/lbry/schema/claim.py
+++ b/lbry/schema/claim.py
@@ -253,7 +253,7 @@ class Stream(BaseClaim):
         if stream_type in ('image', 'video', 'audio'):
             media = getattr(self, stream_type)
             media_args = {'file_metadata': None}
-            if file_path is not None:
+            if file_path is not None and not all((duration, width, height)):
                 try:
                     media_args['file_metadata'] = binary_file_metadata(binary_file_parser(file_path))
                 except:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lbry/schema/claim.py [hunk 1]
```diff
@@ -253,7 +253,7 @@ def update(self, file_path=None, height=None, width=None, duration=None, **kwarg
         if stream_type in ('image', 'video', 'audio'):
             media = getattr(self, stream_type)
             media_args = {'file_metadata': None}
-            if file_path is not None:
+            if file_path is not None and (not duration or not width or not height):
```
eukreign: I think this can be `if file_path is not None and not all((duration, width, height))` but please confirm.

BrannonKing: it can be that "all" approach. unfortunately t only saves a couple of chars

PR #2881, opened 2020-03-24
add `wallet_syncing` to `status`, which is set to true when processing transaction updates

shyba:


tzarebczan:
This is good for a first start for desktop users (they'll need to re-sync, so I'd really like to get this in) but it should be returned in wallet_status and not status. 

For lbry.tv, it may not be worth it if we are moving away from sync all together.

kauffj:
https://github.com/lbryio/lbry-desktop/issues/3903

PR #2883, opened 2020-03-25
WIP: Try apsw build

shyba:
should build, but tests wont pass with apsw yet


shyba:
looks like we're going to use peewee, so closing this as there wont be apsw directly on client for now

PR #2887, opened 2020-03-27
added `txo_spend` command to support liquidating large number of txos (eg. tips) by batching them across several transactions

### LINKED ISSUE #941: Add support to abandon multiple claims/txid+nouts at once
## The Issue
The [claim abandon](https://lbryio.github.io/lbry/cli/#claim_abandon) command only takes a single claim or txid/nout currently.  Going forward, it may be beneficial in some cases to abandon multiple claims in a single transaction.  One such use case would be where a user is deleting their channel and all associated claims. This ticket is to add batch support to abandoning claims or change the daemon to support https://github.com/lbryio/lbry-app/issues/625#issuecomment-336473481. 

The daemon should still be responsible for making sure the claims are allowed to be abandoned - i.e. don't allow abandoning of a channel if there are still published claims under it and provide a response to any supports (tips + self supports) that are part of a claim being abandoned so that the user can abandon all in 1 shot.

### LINKED ISSUE #2528: Mass abandoning supports / support create times out
When trying to abandon a support (for example this one with 1634 transactions), the SDK fails:

```
1d/1z [lbry@ytsync4:~] 1m18s 148 $ /opt/lbry/lbrynet support abandon --blocking --claim_id=173078c44b5b7ec794f8bf4a341c7b5a346b6a8e
  File "lbry/extras/cli.py", line 317, in <module>
  File "lbry/extras/cli.py", line 307, in main
  File "asyncio/base_events.py", line 573, in run_until_complete
  File "lbry/extras/cli.py", line 32, in execute_command
  File "site-packages/aiohttp/client.py", line 1005, in __aenter__
  File "site-packages/aiohttp/client.py", line 497, in _request
  File "site-packages/aiohttp/client_reqrep.py", line 857, in start
  File "site-packages/aiohttp/helpers.py", line 585, in __exit__
concurrent.futures._base.TimeoutError
[2660] Failed to execute script cli
```

this also never returns to the JSON RPC, which means that if the call is invoked over HTTP, the client will remain hanging until the timeout is reached.

### LINKED ISSUE #2847: add --tips parameter to support abandon
When we allow users to claim multiple tips at once, they should have an option of keeping their supports active and only claim their tips.

eukreign:
fixes #2847
fixes #2528
fixes #941

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -4260,7 +4260,7 @@ class Daemon(metaclass=JSONRPCServerType):
                       [--claim_id=<claim_id>...] [--channel_id=<channel_id>...] [--name=<name>...]
                       [--is_my_input | --is_not_my_input]
                       [--exclude_internal_transfers] [--wallet_id=<wallet_id>]
-                      [--batch_size=<batch_size>]
+                      [--preview] [--blocking] [--batch_size=<batch_size>]
 
         Options:
             --type=<type>              : (str or list) claim type: stream, channel, support,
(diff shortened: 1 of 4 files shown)


tzarebczan:
Can wallet send also use the batching please?

tzarebczan:
Actually, now I remembered that would be a breaking change since the return signature changes. 

You'd expect other commands to support this as well though, even something like support create that's trying to create too large of a transaction. But that wouldn't work for something like stream/channel creation.  We'd probably tell the user to considate their wallet in that case, and wallet fund would need to support batching.

PR #2892, opened 2020-03-29
fixed is_valid_public_ipv4 not working correctly when ipv6 is given as address

endes123321:
When UPnP gives an ipv6 address to the function **is_valid_public_ipv4**, instead of return False, it throws an error: 
```
  File "lbry/extras/daemon/component.py", line 63, in _setup
  File "lbry/extras/daemon/component.py", line 55, in _setup
  File "lbry/extras/daemon/components.py", line 493, in start
  File "lbry/extras/daemon/components.py", line 430, in _maintain_redirects
  File "lbry/dht/peer.py", line 39, in is_valid_public_ipv4
  File "ipaddress.py", line 79, in ip_network
  File "ipaddress.py", line 2178, in __init__
ValueError: <redacted ipv6>/32 has host bits set
```
This is because the checks doesn't stop when `parsed_ip.version != 4`, but `CARRIER_GRADE_NAT_SUBNET.supernet_of(ipaddress.ip_network(f"{address}/32"))` and `IPV4_TO_6_RELAY_SUBNET.supernet_of(ipaddress.ip_network(f"{address}/32")))` needs a valid IPv4 address.


tzarebczan:
Thanks so much for finding this and opening a PR! Can we show you some [appreciation](https://lbry.com/faq/appreciaiton)?

endes123321:
Just a small fix, I don't want appreciation. Also, sorry for responding late.

PR #2897, opened 2020-03-31
handle claim reorgs in the wallet server

jackrobison:


tests/integration/blockchain/test_blockchain_reorganization.py
```diff
@@ -13,21 +17,65 @@ class BlockchainReorganizationTests(IntegrationTestCase):
...
+        self.assertEqual(2, REORG_COUNT._samples()[0][2])
+
+    async def test_reorg_dropping_claim(self):
+        # sanity check
+        result_txs, _, _, _ = await self.ledger.claim_search([], name='hovercraft')
+        self.assertListEqual(result_txs, [])
+
+        # create a claim and verify it is returned by claim_search
+        self.assertEqual(self.ledger.headers.height, 206)
+        broadcast_tx = Transaction(unhexlify((await self.stream_create(name='hovercraft'))['hex'].encode()))
```
eukreign, line 49: fyi, if you go straight to `self.daemon.jsonrpc_stream_create()` it returns the original Transaction object without having to jump through hoops to recreate it

tests/integration/blockchain/test_blockchain_reorganization.py
```diff
@@ -13,21 +17,65 @@ class BlockchainReorganizationTests(IntegrationTestCase):
...
+        self.assertEqual(self.ledger.headers.height, 207)
+        result_txs, _, _, _ = await self.ledger.claim_search([], name='hovercraft')
+        self.assertEqual(1, len(result_txs))
+        tx = result_txs[0]
+        self.assertEqual(tx.tx_ref.id, broadcast_tx.id)
```
eukreign, line 54: variable name would be clearer as `txo` (that is what claim_search returns) then `txo.tx_ref.id` makes more sense

PR #2898, opened 2020-03-31
`ffmpeg` file analysis returns duration as integer now

eukreign:


lbry/file_analysis.py
```diff
@@ -354,7 +354,7 @@ def _get_best_container_extension(scan_data, video_encoder):
     def _build_spec(scan_data):
         assert scan_data
 
-        duration = float(scan_data["format"]["duration"])  # existence verified when scan_data made
+        duration = int(scan_data["format"]["duration"])  # existence verified when scan_data made
```
BrannonKing: This plan isn't going to work:
```
int("10.23")
ValueError: invalid literal for int() with base 10: '10.23'
```

PR #2908, opened 2020-04-05
Fix connection to wallet server breaking upon giant resolve requests

jackrobison:
fixes https://github.com/lbryio/lbry-sdk/issues/2907

### DIFF
--- a/lbry/wallet/ledger.py
+++ b/lbry/wallet/ledger.py
@@ -752,7 +752,11 @@ class Ledger(metaclass=LedgerRegistry):
 
     async def resolve(self, accounts, urls, **kwargs):
         resolve = partial(self.network.retriable_call, self.network.resolve)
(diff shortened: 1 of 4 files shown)


lbry/wallet/ledger.py
```diff
@@ -752,7 +752,11 @@ def broadcast(self, tx):
 
     async def resolve(self, accounts, urls, **kwargs):
         resolve = partial(self.network.retriable_call, self.network.resolve)
-        txos = (await self._inflate_outputs(resolve(urls), accounts, **kwargs))[0]
+        urls_copy = list(urls)
+        txos = []
+        while urls_copy:
+            batch, urls_copy = urls_copy[:500], urls_copy[500:]
```
eukreign, line 758: fyi, this will re-allocate memory for two new lists on each loop, if the lists are big this will probably use lots of RAM

PR #2910, opened 2020-04-07
Comment_hide update

osilkin98:
`comment_hide` now returns lists of `comment_ids` that are both `hidden` and `visible`.

The combined lists contain all the `comment_ids` passed into the parameters


osilkin98:
This is needed for comment hiding on the desktop app, please check ASAP

PR #2946, opened 2020-05-07
duplicate on lines 2047 to 2052 / 2059 to 2064

lpessin:
deleted lines 2047 to 2052

### DIFF
--- a/docs/api.json
+++ b/docs/api.json
@@ -2043,13 +2043,7 @@
                         "type": "str",
                         "description": "description of the collection",
                         "is_required": false
-                    },
-                    {
-                        "name": "clear_languages",
-                        "type": "bool",
-                        "description": "clear existing languages (prior to adding new ones)",
-                        "is_required": false
-                    },
+                    },                    
                     {
                         "name": "tags",
                         "type": "list",


eukreign:
the modified file is generated and should not be manual edited, if there is duplication it needs to be fixed at the source of the duplication

lpessin:
Got it. Mb. I was following the link 'edit this page on GithHub', on the API Doc. webpage.

eukreign:
this is probably the source of the problem you found: https://github.com/lbryio/lbry-sdk/blob/master/lbry/extras/daemon/daemon.py#L3616-L3618

lpessin:
Yes, on collection_create. I'll take a look, thanks )

lpessin:
> 
> 
> this is probably the source of the problem you found:
> 
> https://github.com/lbryio/lbry-sdk/blob/8be1c8310d5ed2069ed4f97620e9297fd59eb449/lbry/extras/daemon/daemon.py#L3616-L3618

well, I guess if generates from there it's only needed to delete one of these lines. am I right? Should I open an issue or another pull request?

eukreign:
I would open a new PR with the fix but also it would be great if you can setup your environment to be able to generate the fixed doc; to generate the docs you run this script: https://github.com/lbryio/lbry-sdk/blob/master/scripts/generate_json_api.py (this is also where all of the examples in the documentation come from, they are actually "live examples", so that they don't get stale and are actually verified to work as we make progress on LBRY.)

PR #2947, opened 2020-05-07
properly handle integer environment values, fixes issue with specifying ports via environment vars

thebubbleindex:
make sure tcp and udp port for dht are int type

Reason for fix (Ubuntu 20.04 and desktop app):
1. Without fix users specify env vars, LBRY_TCP_PORT=34562 and LBRY_UDP_PORT=34562, but then when you run the daemon it complains about string comparisons.
2. With this fix people can download the desktop application, set the above environment variables and have their seeding working with VPN! PIA VPN gives you a randomly assigned port forward port, with this fix a user can now use a VPN (with ENV vars).

### DIFF
--- a/lbry/conf.py
+++ b/lbry/conf.py
@@ -277,8 +277,17 @@ class Strings(ListSetting):
 class EnvironmentAccess:
     PREFIX = 'LBRY_'
 
-    def __init__(self, environ: dict):
-        self.environ = environ
+    def __init__(self, config: 'BaseConfig', environ: dict):
+        self.configuration = config
+        self.environ = {}
+        if environ:
+            self.load(environ)
+
+    def load(self, environ):
+        for setting in self.configuration.get_settings():
+            value = environ.get(f'{self.PREFIX}{setting.name.upper()}', NOT_SET)
+            if value != NOT_SET and not (isinstance(setting, ListSetting) and value is None):
+                self.environ[f'{self.PREFIX}{setting.name.upper()}'] = setting.deserialize(value)
 
     def __contains__(self, item: str):
         return f'{self.PREFIX}{item.upper()}' in self.environ
@@ -443,7 +452,7 @@ class BaseConfig:
         self.arguments = ArgumentAccess(self, args)
 
     def set_environment(self, environ=None):
-        self.environment = EnvironmentAccess(environ or os.environ)
+        self.environment = EnvironmentAccess(self, dict(environ or os.environ))
 
     def set_persisted(self, config_file_path=None):
         if config_file_path is None:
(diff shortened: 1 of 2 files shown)


eukreign:
Thanks for the PR!

The proper place to fix this is in the `conf.py` module, it is already doing the conversion to `int` there, so if the conversion is not working this is a bug and we would need a test showing it failing and then the accompanying fix:

Where conversion is already being done:
https://github.com/lbryio/lbry-sdk/blob/8be1c8310d5ed2069ed4f97620e9297fd59eb449/lbry/conf.py#L101

Tests that need to be updated to reproduce the problem you are experiencing:
https://github.com/lbryio/lbry-sdk/blob/8be1c8310d5ed2069ed4f97620e9297fd59eb449/tests/unit/test_conf.py#L91-L95

thebubbleindex:
Might need some help, I cannot find where the environment variables are being converted into a type.

https://github.com/lbryio/lbry-sdk/blob/1d1f0527ee4c4c643c19cb64e749d3ffb2262a27/lbry/conf.py#L277

eukreign:
I think you've found the problem, if you look at the `ArgumentAccess` and `ConfigFileAccess` they all have `load()` methods which are called in `__init__` to `deserialize()` the arguments but the `EnvironmentAccess` does not have one.

thebubbleindex:
Thanks! I updated the conf.py to a working version, adding some type parsing and loading to the EnivronmentAccess

https://github.com/thebubbleindex/lbry-sdk/blob/13612eefcc7cec2d4052b10edcf20b8ae3745d91/lbry/conf.py#L277

eukreign:
It should be very close to how `ArgumentAccess` works and loop over all of the pre-defined settings and look for them inside the environment: `for setting in self.configuration.get_settings()`.

Also, this needs tests, they would go here: https://github.com/lbryio/lbry-sdk/blob/master/tests/unit/test_conf.py#L91 (look at all of the other tests in that file to get inspiration).

You are making good progress, don't hesitate to ask about how something work if you are confused, this `conf` code in LBRY heavily uses Python descriptors which are explained here:

https://docs.python.org/3/howto/descriptor.html
https://docs.python.org/3.7/reference/datamodel.html#implementing-descriptors
https://docs.python.org/3.7/reference/datamodel.html#invoking-descriptors

thebubbleindex:
Cool, think I have it now. I'll try to add the tests (I'm new to Python, but the examples you mentioned should be enough). Cheers!

eukreign:
Looking good! Instead of adding two new config options to the test config could you just re-use `test_int` that's already there (it's type `Integer`).

tzarebczan:
@thebubbleindex thanks for the PR, congrats on your first one! 

Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the contribution?

PR #2957, opened 2020-05-17
added `download_path` as a filter for `file_list` command

akinwale:


jackrobison:
@akinwale please update the docstring for `file_list` so the new arg will work from cli

PR #2959, opened 2020-05-18
Add all sqlite coin chooser

jackrobison:


tzarebczan:
This also does not take into account unconfirmed utxos (even though the commit says so!)

eukreign:
If there are concerns about this selection algorithm not being as complete as the others then perhaps it should not be the default and it can then be explicitly configured in cases where performance is slow?

PR #2966, opened 2020-05-25
add `allowed_origin` to config, by default no longer allow any requests which pass Origin in header

jackrobison:
backwards-incompatible: This release requires that no Origin value is passed in HTTP API requests. Electron based apps accessing the SDK will need to make sure that for `POST` requests that they remove the `Origin` from the header, see: https://github.com/electron/electron/issues/7931 . When you want to use the API from web browser the SDK needs to have `allowed_origin` configuration value set to the host from which you want to access the API.


kauffj:
- `'null'` will be replaced by `localhost` as default
- origin will only be enforced _if_ origin is passed

PR #2968, opened 2020-05-26
more aggressive video transcoding to reduce file sizes

tzarebczan:
Have noticed the defaults aren't aggressive enough to stream smoothly (yet). Downgrade max rate to 5500K, higher crf = smaller file size for now.


tzarebczan:
Please include in next release :)

PR #2972, opened 2020-06-04
Improve wallet server address history cache and the rate of sent notifications

jackrobison:
Minimizes but does not entirely solve https://github.com/lbryio/lbry-sdk/issues/2945 as the cache is invalidated each block and the first slow request can back up further notifications during the same block.

### DIFF
--- a/lbry/wallet/rpc/session.py
+++ b/lbry/wallet/rpc/session.py
@@ -206,7 +206,7 @@ class SessionBase(asyncio.Protocol):
         """
         return self._address
 
-    def peer_address_str(self):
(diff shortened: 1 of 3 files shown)


lbry/wallet/rpc/session.py
```diff
@@ -206,7 +206,7 @@ def peer_address(self):
         """
         return self._address
 
-    def peer_address_str(self):
+    def peer_address_str(self, for_log=True):
```
eukreign, line 209: where is `for_log` used?

eukreign, line 209: where is `for_log` used?

PR #2986, opened 2020-06-25
check sample rate and lower if too high

Death916:
New pull request with same commit as this one https://github.com/lbryio/lbry-sdk/pull/2933

Just had to clean up some git stuff.
 
Issue https://github.com/lbryio/lbry-sdk/issues/2906

### DIFF
--- a/lbry/file_analysis.py
+++ b/lbry/file_analysis.py
@@ -195,10 +195,13 @@ class VideoFileAnalyzer:
             if stream["codec_type"] != "audio":
                 continue
             codec = stream["codec_name"]
+            sample_rate = stream['sample_rate']
             log.debug("   Detected audio codec is %s", codec)
             if not {"aac", "mp3", "flac", "vorbis", "opus"}.intersection(codec.split(",")):
                 return "Audio codec is not in the approved list of AAC, FLAC, MP3, Vorbis, and Opus. " \
                        f"Actual: {codec} [{stream['codec_long_name']}]"
+            if sample_rate > "48000":
+                return "Sample rate out of range"
 
         return ""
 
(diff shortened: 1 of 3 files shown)


lbry/file_analysis.py
```diff
@@ -195,10 +195,13 @@ def _verify_audio_encoding(scan_data: json):
             if stream["codec_type"] != "audio":
                 continue
             codec = stream["codec_name"]
+            sample_rate = stream['sample_rate']
             log.debug("   Detected audio codec is %s", codec)
             if not {"aac", "mp3", "flac", "vorbis", "opus"}.intersection(codec.split(",")):
                 return "Audio codec is not in the approved list of AAC, FLAC, MP3, Vorbis, and Opus. " \
                        f"Actual: {codec} [{stream['codec_long_name']}]"
+            if sample_rate > "48000":
```
eukreign, line 203: What are the possible other `sample_rate` values? This comparison will produce an incorrect result if the length of the string is different even if one rate is higher than the other rate... since it's doing string comparisons, not integer comparisons.

tzarebczan: @Death916 can you address the above? Maybe we can make the check more robust in case it does not exist.

Death916: @tzarebczan @eukreign 
Sorry was busy for a bit. I changed the checks to int on my local now. i originally just figured it wouldn't matter since the values were already strings from the json.
What do you mean by it does not exist though? 
I was mostly just trying to address the issue mentioned specifically for this PR, but if you could elaborate on what you mean it shouldn't be a problem to do. Kind of late here so I could b missing something obvious. Do you mean if it cant find a sample rate to just give it one that's acceptable?

eukreign: @Death916 doing string comparisons leads to unexpected behavior:
```
>>> "50" > "48000"
True
>>> "3000000000" > "48000"
False
>>> "a" > "48000"
True
```

what are the possible values that `sample_rate` can be? can it be an empty string? can it be something other than an integer (perhaps a float? what else?)?
once we have a better idea of possible range of values then at least for the integer/float case the two sides of the `>` equality need to be converted to integers or floats.

PR #2988, opened 2020-06-29
Added Better Permission Checks for Configuration Directories

Geo25rey:
Inspired by Bug #877, this amends the initial checks for directories required by the daemon. Additionally, this self fixes permissions for the directories. 

I'd like to add the option to not exit the daemon when optional directories are inaccessible, but I'm not sure which are important. In my testing, "~/Downloads" seems to be 1 of those optional directories.

Also, it maybe be useful to set the directory permissions to 755 instead of 777. Your thoughts?

P. S. I can't replicate the issue described in #877. I think PR #2454 fixed the issue, but the changes may have gotten lost in a merge (I can't find them in a current version).

### DIFF
--- a/lbry/extras/cli.py
+++ b/lbry/extras/cli.py
@@ -224,8 +224,15 @@ def get_argument_parser():
 
 
 def ensure_directory_exists(path: str):
-    if not os.path.isdir(path):
-        pathlib.Path(path).mkdir(parents=True, exist_ok=True)
+    pathObj = pathlib.Path(path)
+    try:
+        if pathObj.is_dir():
+            pathObj.chmod(0o777)
+        else:
+            pathObj.mkdir(parents=True, exist_ok=True)
+    except:
+        print("Could not access", path)
+        sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


kauffj:
Thanks @Geo25rey! @eukreign will review

Geo25rey:
@eukreign Which directories are required for startup, though? I know the logger doesn't start unless its specific directory is writable. Should I try to remove the directory permission checking/fixing until each directory is required? Would a warning that permissions are getting changed be sufficient? Plus, the changing of permissions only happens to directories (not files) which the user specifically owns or is created within the daemon. 

I looked in the documentation for pathlib and I didn't see any way to read permissions of a directory. If there's something available to check permissions, I'd much prefer that over writing new permissions every time.

Also, regarding the #877 comment, I don't have a Mac to test on; I think that the permissions differ slightly between hfs+ and ext4 filesystems. I can test under Linux.

eukreign:
@Geo25rey it needs the data directory available at a minimum so that it can sync with the wallet server and a directory to download blobs

PR #3011, opened 2020-08-01
resolve comments faster by batching maybe

jessopb:


kauffj:
@jessopb is this an old draft?

kauffj:
re-open if I got it wrong

PR #3033, opened 2020-09-02
comment rpc call includes per jsonrpc standard

jessopb:


eukreign:
done in https://github.com/lbryio/lbry-sdk/pull/3051

PR #3037, opened 2020-09-04
ignore "raise from" linter errors

shyba:


shyba:
cherry-picked into #3034

PR #3039, opened 2020-09-07
Include enough information in resolve protobuf so its possible to verify signatures

shyba:


eukreign (changes requested):
Provided feedback in slack, but roughly:
1. need to make the code closer to the old pattern of a `search_claims` which usable by both resolve and search, currently resolve calls a version of search_claims which applies the search censor
2. integration test is too low level, maybe it should be broken out into a few unit tests and a higher level integration test (one which makes less assumptions about internals)

lbry/blockchain/database.py
```diff
@@ -16,7 +16,9 @@
 
 def make_short_url(r):
     try:
-        return f'{normalize_name(r["name"].decode())}#{r["shortestID"] or r["claimID"][::-1].hex()[0]}'
+        # fixme: we describe it as normalized but the old SDK didnt do that
```
eukreign, line 19: can you change this to `TODO`, that's what i've been using elsewhere for reminders of stuff that needs to be fixed, that way we can just grep for `TODO`

eukreign (changes requested):

lbry/blockchain/database.py
shyba: changed, good point

eukreign (approved):

PR #3040, opened 2020-09-07
Synchronizer: wait for lbrycrd to start if its not ready, instead of raising

shyba:


tests/integration/blockchain/test_blockchain.py
```diff
@@ -613,6 +613,14 @@ async def test_multi_block_file_sync(self):
 
 
 class TestGeneralBlockchainSync(SyncingBlockchainTestCase):
+    async def test_sync_waits_for_lbrycrd_to_start(self):
+        await self.sync.stop()
+        await self.chain.stop()
+        sync_start = asyncio.ensure_future(self.sync.start())
```
eukreign, line 619: Should this be `create_task()` instead?

according to python docs create_task is preferred: https://docs.python.org/3.7/library/asyncio-future.html#asyncio.ensure_future

lbry/blockchain/sync/synchronizer.py
```diff
@@ -43,9 +43,19 @@ def __init__(self, chain: Lbrycrd, db: Database):
         self.advance_loop_task: Optional[asyncio.Task] = None
         self.advance_loop_event = asyncio.Event()
 
+    async def wait_for_chain_ready(self):
```
eukreign, line 46: if `lbrycrd` was started without zmq enable would this just hang forever?

eukreign (changes requested):

lbry/blockchain/sync/synchronizer.py
shyba: yes. from the original report: `ZMQ readiness detection (currently just quits if lbrycrd isn't ready for ZMQ requests)`
so it wont give up until its ready, unless i misunderstood it

currently it waits in silence but thats because logging is broken. there is code on the exception handler to log what happened while it waits

tests/integration/blockchain/test_blockchain.py
shyba: nice catch, pushed a change

lbry/blockchain/sync/synchronizer.py
eukreign: The method name is misleading then because it makes it seem like this waits for lbrycrd to be ready (in general) but in fact it's only waiting for zmq to be available... which may not be enabled, so lbrycrd would be ready (but not ZMQ).

lbry/blockchain/lbrycrd.py
```diff
@@ -243,6 +243,8 @@ def unsubscribe(self):
             self.subscription = None
 
     async def rpc(self, method, params=None):
+        if self.session.closed:
+            raise Exception("session is closed! We are shutting down.")
```
kauffj, line 247: who's we

lbry/blockchain/sync/synchronizer.py
shyba: changed to instruct which conf key fixes it and exit in this case: `Lbrycrd is misconfigured. Please double check if zmqpubhashblock is properly set on lbrycrd.conf`

lbry/blockchain/lbrycrd.py
shyba: changed to explain the real issue: `session is closed! RPC attempted during shutting down.`

eukreign (approved):

PR #3045, opened 2020-09-14
add forward compatibility for byte datagram keys

jackrobison:


lbry/dht/serialization/datagram.py
```diff
@@ -144,27 +144,37 @@ def __init__(self, packet_type: int, rpc_id: bytes, node_id: bytes, exception_ty
...
     primitive: typing.Dict = bdecode(datagram)
-    if primitive[0] in [REQUEST_TYPE, ERROR_TYPE, RESPONSE_TYPE]:  # pylint: disable=unsubscriptable-object
-        datagram_type = primitive[0]  # pylint: disable=unsubscriptable-object
+
+    converted = {
+        str(k).encode() if not isinstance(k, bytes) else k: v for k, v in primitive.items()
+    }
+
+    if converted[b'0'] in [REQUEST_TYPE, ERROR_TYPE, RESPONSE_TYPE]:  # pylint: disable=unsubscriptable-object
+        datagram_type = converted[b'0']  # pylint: disable=unsubscriptable-object
```
eukreign, line 161: Does more recent `pylint` still have this issue? What if you add type information to `converted`, maybe `pylint` will detect it as `dict` then?

lbry/dht/serialization/datagram.py
```diff
@@ -144,27 +144,37 @@ def __init__(self, packet_type: int, rpc_id: bytes, node_id: bytes, exception_ty
         self.response = response.decode()
 
 
-def decode_datagram(datagram: bytes) -> typing.Union[RequestDatagram, ResponseDatagram, ErrorDatagram]:
+def _decode_datagram(datagram: bytes):
```
lyoshenka, line 147: you lost return type here

lyoshenka (approved):

lbry/dht/serialization/datagram.py
eukreign: i assume that since @jackrobison made this an internal helper function (with underscore) it didn't benefit from having the type info? i think either way is fine: having it doesn't hurt anything, not having it is probably okay in this special case

personally, my preference would be to avoid having such large type `Union` in the first place, if all of the types have the same parent class, perhaps just say it returns the parent class type? unions should be used as last resort (or when the values returned are truely different, like a function returning an int and a str, in this case it seems it returns `Datagram`s)

PR #3052, opened 2020-09-28
Fixed docker-compose - no snapshot error

kodxana:
Curently new containers do not get provide snapshots from LBRYinc (Wallet and CRD) that causes containers to get into restart loop.


jackrobison (approved):
lgtm

PR #3057, opened 2020-10-08
support comment pinning

jessopb:


eukreign (approved):
@jessopb there is an extraneous print statement in the tests, otherwise looks good

PR #3073, opened 2020-10-29
recover from state where wallet is set to encrypt but not encrypted and no password given

### LINKED ISSUE #3026: sync_hash fails if encrypted on disk = true, but wallet not encrypted (is_encrypted shows true in wallet status also)
There are some scenarios where the encrypted on disk value is incorrect(one is: if sync apply with a blank password was used on an encrypted wallet) and in this case, sync hash fails when run. But after the app runs sync apply to get a new hash, then sync hash works again. 

Error: 'NoneType' object has no attribute 'encode'

It should not fail and should try to run the hash knowing that the wallet is not encrypted and the encrypted on disk value is 
incorrect. 

This prevents startup on Android and causes issues on desktop/lbry.tv as well.

is_encrypted in wallet status should also be false.  encrypted on disk setting should be reset in wallet.

shyba:
closes #3026

### DIFF
--- a/lbry/wallet/wallet.py
+++ b/lbry/wallet/wallet.py
@@ -145,14 +145,15 @@ class Wallet:
             elif not self.is_locked:
                 log.warning(
                     "Disk encryption requested but no password available for encryption. "
-                    "Saving wallet in an unencrypted state."
+                    "Resetting encryption preferences and saving wallet in an unencrypted state."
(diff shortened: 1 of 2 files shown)


tzarebczan:
This works, thank you!

tzarebczan (approved):

PR #3077, opened 2020-11-06
Claim search fixes with a simple trending example and content filters/blockers

shyba:


lbry/blockchain/sync/synchronizer.py
```diff
@@ -43,6 +44,12 @@ def __init__(self, chain: Lbrycrd, db: Database):
         self.on_block_subscription: Optional[BroadcastSubscription] = None
         self.advance_loop_task: Optional[asyncio.Task] = None
         self.advance_loop_event = asyncio.Event()
+        self.filtering_channel_hashes = {
+            unhexlify(channel_id)[::-1] for channel_id in
+            os.getenv('FILTERING_CHANNEL_IDS', '').split(' ') if channel_id}
+        self.blocking_channel_hashes = {
+            unhexlify(channel_id)[::-1] for channel_id in
+            os.getenv('BLOCKING_CHANNEL_IDS', '').split(' ') if channel_id}
```
eukreign, line 52: New SDK uses the same config for client and server so these values should come from that config (which does support environment variables).

lbry/blockchain/sync/claims.py
```diff
@@ -281,3 +281,19 @@ def update_channel_stats(blocks: Tuple[int, int], initial_sync: int, p: Progress
...
         p.ctx.execute_notx(text("VACUUM claim;"))
     p.step(result.rowcount)
+
+
+@event_emitter("blockchain.sync.claims.filters", "claim_filters")
+def update_claim_filters(blocking_channel_hashes, filtering_channel_hashes, p: ProgressContext):
+    def select_reposts(channel_hashes, filter_type=0):
+        return select(
+            Claim.c.reposted_claim_hash, filter_type, Claim.c.channel_hash).where(
+            (Claim.c.channel_hash.in_(channel_hashes)) & (Claim.c.reposted_claim_hash.isnot(None)))
```
eukreign, line 291: If this query can be re-used in other places it should be moved out into the outer namespace. Also, the formatting should be something like:
```
        return (
            select(Claim.c.reposted_claim_hash, filter_type, Claim.c.channel_hash).where(
                (Claim.c.channel_hash.in_(channel_hashes)) &
                (Claim.c.reposted_claim_hash.isnot(None))
            )
        )
```

lbry/blockchain/sync/claims.py
```diff
@@ -281,3 +281,19 @@ def update_channel_stats(blocks: Tuple[int, int], initial_sync: int, p: Progress
...
+
+    p.ctx.execute(ClaimFilter.delete())
+    # order matters: first we insert the blocked ones. Then the filtered ones.
+    # If there is already a block in place, that takes priority because a block is just a harder filter
+    p.ctx.execute(ClaimFilter.insert().from_select(
+        ['claim_hash', 'filter_type', 'owner_channel_hash'], select_reposts(blocking_channel_hashes, 2)))
+    p.ctx.execute(p.ctx.insert_or_ignore(ClaimFilter).from_select(
+        ['claim_hash', 'filter_type', 'owner_channel_hash'], select_reposts(filtering_channel_hashes, 1)))
```
eukreign, line 299: the last parenthesis should be on line by itself

lbry/db/database.py
```diff
@@ -277,6 +277,17 @@ async def get_purchases(self, **constraints) -> Result[Output]:
...
+            channel_url = constraints.pop('channel')
+            match = await self.resolve([channel_url])
+            if isinstance(match, dict):
+                for value in match.values():
+                    if isinstance(value, Output):
+                        constraints['channel_hash'] = value.claim_hash
+                    else:
+                        return Result([], 0)
+            else:
+                return Result([], 0)
```
eukreign, line 290: It would be a lot more performant to do this in `q.search_claims` because that cuts in half the main process <-> sub tasks communication. I ended up implementing this in my branch earlier: https://github.com/lbryio/lbry-sdk/blob/cae800565dc276c4b05cd93738aa82e757596b07/lbry/db/queries/search.py#L284-L291

eukreign (changes requested):
i have more feedback i will provide later but there is some actionable stuff here already:

shyba:
all comments addressed on last push

tests/integration/blockchain/test_blockchain.py
```diff
@@ -1338,6 +1381,39 @@ async def test_create_and_abandon_in_same_block(self):
...
         )
 
+    async def test_content_filtering(self):
+        user_chan = await self.get_claim(
+            await self.create_claim(claim_id_startswith='ab', is_channel=True, name="@some_channel"))
+        await self.create_claim(claim_id_startswith='cd', sign=user_chan, name="good_content")
+        bad_content = await self.get_claim(
+            await self.create_claim(claim_id_startswith='ef', sign=user_chan, name="bad_content"))
+        moderator_chan = await self.get_claim(
+            await self.create_claim(claim_id_startswith='ab', is_channel=True, name="@filters"))
```
eukreign, line 1391: `claim_id_startswith` is pretty expensive operation (it generates lots of claims until the hash matches requested prefix), unless you actually need it, it's best not to use it.

eukreign (changes requested):

tests/integration/blockchain/test_blockchain.py
shyba: pushed, good find

PR #3087, opened 2020-11-17
One spv session per loaded wallet

jackrobison:
Create a per-wallet connection to the spv server for account address subscriptions and sync


lbry/extras/daemon/daemon.py
```diff
@@ -1303,8 +1303,9 @@ async def jsonrpc_wallet_create(
                     'name': SingleKey.name if single_key else HierarchicalDeterministic.name
```
ho-ngoc-thoai, line 1303: Hồ Ngọc thoại

ho-ngoc-thoai (approved):

lyoshenka:
no longer needed

PR #3114, opened 2020-12-14
Fix getting all needed headers before beginning to sync txs

jackrobison:


tzarebczan:
Not working with test wallet... gave Jack results.

PR #3115, opened 2020-12-14
update staked_amount to include supports for claims in the channel's stake

lyoshenka:


lbry/db/queries/txio.py
```diff
@@ -143,28 +143,35 @@ def distribute_unspent_txos(
...
+        claims_with_changed_supports(blocks)
     )
 
 
 def count_claims_with_changed_supports(blocks: Optional[Tuple[int, int]]) -> int:
-    sql = (
-        select(func.count(distinct(TXO.c.claim_hash)).label('total'))
-        .where(where_changed_support_txos(blocks))
-    )
+    subq = claims_with_changed_supports(blocks).subquery()
```
lyoshenka, line 173: @eukreign i think this pattern would work better for us going forward. instead of passing the aggregate to the function to select the right thing (e.g. `staked_support_count_calc`), have the function return a query and use that as a subquery to `count(distinct())`

eukreign (commented):
There are some substantial query changes in this PR that might affect performance. Did you check that these changes are still performant on a full database?

lbry/blockchain/sync/claims.py
```diff
@@ -38,26 +38,46 @@ def channel_content_count_calc(signable):
 support = TXO.alias('support')
 
 
-def staked_support_aggregation(aggregate):
+def staked_support_subquery(claim_hash, aggregate):
```
eukreign, line 41: The first parameter should probably be `claim_hash_column` so as not to be confused with a `claim_hash` value.

lbry/blockchain/sync/claims.py
```diff
@@ -247,8 +269,12 @@ def update_stakes(blocks: Tuple[int, int], claims: int, p: ProgressContext):
         Claim.update()
         .where(where_claims_with_changed_supports(blocks))
         .values(
-            staked_support_amount=staked_support_amount_calc(Claim),
-            staked_support_count=staked_support_count_calc(Claim),
+            staked_amount=(
+                    Claim.c.amount + claims_in_channel_amount_calc(Claim.c.claim_hash) +
+                    staked_support_amount_calc(Claim.c.claim_hash)
```
eukreign, line 274: may be better formatted like so:
```
                    Claim.c.amount +
                    claims_in_channel_amount_calc(Claim.c.claim_hash) +
                    staked_support_amount_calc(Claim.c.claim_hash)
```

lbry/db/queries/txio.py
eukreign: if possible, could you rename or spell out the variable `subq` in full to stay with the general Python naming guidance of using whole words

tests/integration/blockchain/test_blockchain.py
```diff
@@ -1032,7 +1032,7 @@ async def test_claim_search_sum(self):
         self.assertEqual(total, 1600000000)
 
         # create a claim in channel A and have channel B support that claim
-        claim_a = await self.get_claim(await self.create_claim(name="bob", amount='2.0', sign=channel_a))
+        claim_a = await self.get_claim(await self.create_claim(name="alfred", amount='2.0', sign=channel_a))
```
eukreign, line 1035: what do you have against bob?! :-D

eukreign (changes requested):

tests/integration/blockchain/test_blockchain.py
lyoshenka: claim_a is alfred, claim_b is bob ;-)

PR #3119, opened 2020-12-15
fix windows tests on rewrite branch

shyba:


lbry/blockchain/sync/synchronizer.py
```diff
@@ -259,7 +259,7 @@ async def count_takeovers(self, blocks) -> int:
 
     async def sync_claims(self, blocks) -> bool:
         delete_claims = takeovers = claims_with_changed_supports = claims_with_changed_reposts = 0
-        initial_sync = not await self.db.has_claims()
+        initial_sync = not await self.db.has_filters()
```
eukreign, line 262: why this change? i would intuitively expect a function `sync_claims` to check the `claims` table.

tests/integration/blockchain/test_blockchain.py
```diff
@@ -1149,7 +1149,7 @@ async def test_simple_support_trending(self):
         claim1 = await self.get_claim(await self.create_claim(name="one"))
         claim2 = await self.get_claim(await self.create_claim(name="two"))
         await self.generate(1)
-        results = await self.db.search_claims(order_by=["trend_group", "trend_mixed"])
+        results = await self.db.search_claims(order_by=["trending_group", "trending_mixed"])  # test compat layer
```
eukreign, line 1152: maybe outside the scope of this PR, but i would think the compat layer should be somewhere at the API level, not deep down in the db. don't have to fix it for this PR, just an observation for another PR.

eukreign (changes requested):

lbry/blockchain/sync/synchronizer.py
shyba: some tests doesnt create claims, so it will think its the initial sync all the time and try to create indexes that already exist

tests/integration/blockchain/test_blockchain.py
shyba: yes, agree. there are more compat changes I will organize better in another pr

PR #3146, opened 2021-01-14
Find unused ports for media server

### LINKED ISSUE #3032: Port 5280 taken on Linux, app fails to start
We bind mediaserver on 5280 and there's some xmpp client that runs on it that users may have installed. 

If 5280 is not available, try next available port.

michaeltintiuc:
Fixes #3032 and other duplicate or similar bugs. If the default 5280 port is taken try next 10 ports and store the open port back in the config class as I've seen that some other classes are reading from it.
Looking forward to your feedback, thank you.
Testing:
- Run `nc -klv 5280` to start listening on port 5280
- Run `lbrynet start` to start lbrynet
- Notice the logs:
```
2021-01-14 15:06:30,775 INFO     lbry.extras.daemon.daemon:517: Address localhost:5280 is already in use, trying next port
2021-01-14 15:06:30,776 INFO     lbry.extras.daemon.daemon:526: media server listening on TCP ::1:5281
```


tzarebczan:
Thanks for the PR! We'll get a review on it soon. 

 Will reply on DM in slack for appreciation.

eukreign:
@michaeltintiuc thank you for the PR, my initial feedback would be that the code added to `daemon.py` should probably go as a function in one of the modules that deals with blobs/files? `daemon.py` is already huge.

michaeltintiuc:
@eukreign sounds good, I can easily extract it. I assume the rest of the implementation sounds good?

lyoshenka:
thanks for updating the PR. after thinking about it some more, there's a problem with the sdk picking it's own rpc port - which is that other things (the cli, the app, etc) expect the sdk to be bound to the known port. if the sdk binds to a different port on startup, these apps won't know how to connect to it.

perhaps the better solution here is to provide a more helpful error message. most apps i'm aware of will tell you that their port is in use, and simply fail to start. 

can you update your PR to do this instead. or maybe you have a better idea for how to solve this?

lyoshenka:
another way this could be fixed is if the `status` returned the mediaserver port that the sdk ends up picking. in that case it's fine to bind whatever port is available, and other apps would have to know to check `status` for the port.

michaeltintiuc:
Thanks @lyoshenka will check the `status` way first, do we have a list of what has that port number hard-coded in or should I just search through the org repos?

lyoshenka:
Here are the ports (@jackrobison correct me if i missed any)

3333: peer port. used for transferring data between peers
4444: dht port. for all dht-related stuff
5279: rpc port. for local clients (apps, the cli) to issue commands and control sdk behavior
5280: mediaserver. for streaming content to apps using http range requests

lyoshenka:
you can also search for "port" in this file to double-check https://github.com/lbryio/lbry-sdk/blob/master/lbry/conf.py

lyoshenka:
@michaeltintiuc are you still working on this?

michaeltintiuc:
@lyoshenka no not actively, sorry

lyoshenka:
no worries. i'm going to close this PR for now. if you come back to this in the future, we'd love to have you

PR #3153, opened 2021-01-20
hub: use Elasticsearch for `claim_search` and `resolve` calls

shyba:


eukreign (changes requested):
- `torba` was merged into `lbry-sdk` and i believe the name was deprecated, you should use either `wallet`, `wallet-server` or `hub`
- `tests/unit/wallet/server/test_sqldb.py` was deleted but what replaces all of those tests?

docker/docker-compose-wallet-server.yml
```diff
@@ -1,36 +1,40 @@
...
-      # Curently not snapshot provided
-      #- SNAPSHOT_URL=${LBRYCRD_SNAPSHOT_URL-https://lbry.com/snapshot/blockchain}
-      - RPC_ALLOW_IP=0.0.0.0/0
   wallet_server:
-    image: lbry/wallet-server:${WALLET_SERVER_TAG:-latest-release}
-    depends_on:
-      - lbrycrd
+    depends_on: 
+      - es01
+    image: lbry/wallet-server:${WALLET_SERVER_TAG:-development}
```
jackrobison, line 11: should be `latest-release`, once this is merged and released it'll work, we can manually change to `development` for testing

lbry/wallet/rpc/util.py
```diff
@@ -32,10 +32,13 @@
 
 # other_params: None means cannot be called with keyword arguments only
 # any means any name is good
+from functools import lru_cache
```
jackrobison, line 35: should `lbry.utils.LRUCache` be used here instead? `functools.lru_cache` has a `RLock`

jackrobison (changes requested):

PR #3162, opened 2021-01-24
Fix mimetypes for HLS streams, add mimetypes for segment files

DispatchCommit:
Fix mimetype for HLS playlist .m3u8 file. Currently it returns an outdated vendor specific mimetype that has poor support and does not correctly identify in most modern browser.

Instead it should be replaced with the the generic x-mpegURL mimetype as this has wider support.

Also, add mimetypes for segment files, in case those are needed at some point in the future.


DispatchCommit:
This is just a minor improvement upon a much older contribution of mine: https://github.com/lbryio/lbry-sdk/pull/2996

This modification would just make it slightly easier to implement some additional features related to livestreaming by removing some of the logic required to be client side.

eukreign:
Looks good to me; can you rebase on master please?

PR #3176, opened 2021-01-27
fix comment tests and json generat

jessopb:
comment tests now pass
json generation gets 99% of the way, then doesn't write.
I stepped through the recording one r() at a time and it got all the way through.
recording works up until the very last step, where it's apparently unable to write.
apparently wallet method "channel_sign" is returing ' {}' with that extra space.


jessopb:
@shyba said he'd fix the channel sign return type

tzarebczan:
Can we please get this reviewed/merged.

Issue #3158, opened 2021-01-20
Remove external SDK calls


lyoshenka:
- [x] ~~check if online (#3196)~~
- [x] ~~get external ip (#2142)~~
- [x] ~~exchange rate manager (related bug: #2158)~~
- [x] comments (#3408)

PR #3201, opened 2021-02-17
removed `connection_status` field from the `status` command, use `wallet.connected` instead to determine if SDK is connected to a wallet server

### LINKED ISSUE #3196: better connectivity checks
the idea is to stop checking lbry.com to see if we're online

- make a list of places where `is_online` (or whatever its called) is used

- if its the dht announcement, don't announce unless you have peers in routing table
- if its for wallet server, only do things that need a connection if you have a connection
- if its something else, lets discuss on next standup

eukreign:
backwards-incompatible: removes `connection_status` field from `status` API call. see PR link for details.

how to differentiate the states of lbrynet `status` wallet connection

- if `wallet.connected` has a value, then it's connected.
- otherwise:
  - if `startup_status.wallet` is false, then sdk is still connecting to wallet servers. it will never stop trying till it connects.
  - if `startup_status.wallet` is true, then the sdk used to have a wallet server connection but that dropped and it's still trying to reconnect
 
fixes #3196 
related to #3158

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -119,13 +119,6 @@ STREAM_STAGES = [
     (DOWNLOAD_TIMEOUT_CODE, 'Stream timed out')
 ]
 
-CONNECTION_STATUS_CONNECTED = 'connected'
-CONNECTION_STATUS_NETWORK = 'network_connection'
-CONNECTION_MESSAGES = {
-    CONNECTION_STATUS_CONNECTED: 'No connection problems detected',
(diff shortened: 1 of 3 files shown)


lyoshenka:
@seanyesmunt here's how you can differentiate the states of `lbrynet status` wallet connection

- if `wallet.connected` has a value, then it's connected.
- otherwise:
	- if `startup_status.wallet` is false, then sdk is still connecting to wallet servers. it will never stop trying till it connects. 
	- if `startup_status.wallet` is true, then the sdk used to have a wallet server connection but that dropped and it's still trying to reconnect

does this work for you?

Issue #3143, opened 2021-01-11
handling support and claim reorgs in leveldb


kauffj:
leveldb claimtrie
does not touch claim_search, removes usage of claims.db

- [x] fix blockchain.claimtrie.getclaimbyid, swap single claim searches (in tests) to use this.
- [x] handle abandon
- [x] store undo info
- [x] what to do about invalid signatures? no result vs return result and say signature is invalid. resolution: clients should validate signatures
- [x] handle same-block claims/supports
- [x] sync chain
- [x] PR the code
- [x] finish rebase
- [x] TAKEOVER/ACTIVATION HEIGHTS
- [x] wait till victor is done with claim_search because jack's branch drops claim search completely
- [x] confirm that app no longer uses trending fields from `resolve` and `claim_search` responses, since they have been dropped
- [x] handle takeovers from spent supports
- [x] feed ES claims from the block processor 
- [x] claim search tests pass
- [x] test takeover due to a dropped support in a reorg
- [x] test invalidated channel resolution for stream claims where the channel is dropped in a reorg
- [x] track repost counts in ES(45min)
- [x] make sure these make it back https://github.com/lbryio/lbry-sdk/blob/master/tests/unit/wallet/server/test_sqldb.py
- [x] test channel state changes
	- [x] invalidate prev claims for a channel when channel key changes
	- [x] check existing invalid claims in chan to see if any have become valid
	- [x] claim signature changes (adds sig, drops sig, signed by diff channel)
- [x] blocking censored content in `resolve`
  - [x] updates to plugins should let them know that update happened because censoring state changed
- [ ] write a summary of code changes in the PR to aid review
- [x] run on jack's local server so tom can do performance and correctness testing
  - check that as reorgs happen in the wild, this server stays correct

after PR is merged:
- [ ] ship a new snapshot (no more trending data)
- [ ] update all wallet servers
- [ ] consider doing #3131
- [x] have tests use claim_search and getclaimbyid, not just the latter

(later)

maybe redo this stuff inside hub instead of fetching from lbrycrd

- [x] activation delay (calc in #1031) - DONT LET THIS BOG YOU DOWN IF ITS NOT SIMPLE
- [x] add other height fields: active at, created
- [x] compare takeover heights against lbrycrd
- [x] compare effective amounts
- [x] compare controlling status
- [ ] replace block prefetcher with zmq client
- [ ] replace mempool polling with zmq client
- [ ] make zmq subscriber in Go (#3222)
[...]

PR #3205, opened 2021-02-19
use leveldb to resolve urls

jackrobison:
#3143


Ely68 (approved):

belikor:
Hi, I'm looking into the code of the SDK in a general way, and noticed some compilation issues with `plyvel` due to missing `libleveldb-dev`. See #3349.

I noticed that currently `plyvel` is compiled from source, as it's in `setup.py`. Currently it only seems to be used in `lbry.wallet.server.storage`.

I'm not sure how `plyvel` is used but wouldn't it make sense to simply use a prepackaged version of `plyvel`? In Ubuntu the package is in `python3-plyvel`, and in Arch it is in `python-plyvel`. I was thinking about mentioning the package as a dependency in `INSTALL.md`, and removing the variable `PLYVEL` from `setup.py` but I'm not sure if this affects anything in the code.

I tested by not installing `plyvel` at all, and then installing and running the unit tests; they seemed to work okay without referencing `plyvel` at all. So I'm wondering if the current code doesn't use `plyvel`, and it will only use it once this pull request is merged.

belikor:
@jackrobison about my previous comment; I'm wondering if there is a particular reason to use plyvel 1.0.5.

The errors that I mentioned don't occur if we just set `plyvel=1.2.0` or `plyvel=1.3.0` in `setup.py`. This seems to be an easy solution.

jackrobison:
@belikor updated leveldb to 1.3.0

belikor:
Nice.

Is this branch meant to be merged soon? If that's the case, I will close #3357. Otherwise I expect #3357 to be merged before, and you'll have to rebase this branch before merging.

lbry/testcase.py
```diff
@@ -484,13 +484,15 @@ def sout(self, value):
...
+    async def confirm_and_render(self, awaitable, confirm, return_tx=False) -> Transaction:
         tx = await awaitable
         if confirm:
             await self.ledger.wait(tx)
             await self.generate(1)
             await self.ledger.wait(tx, self.blockchain.block_expected)
-        return self.sout(tx)
+        if not return_tx:
+            return self.sout(tx)
+        return tx
```
eukreign, line 495: This should be split into two methods. One that confirms and renders and another one that just confirms (the confirm_and_render would call the confirm only one). Otherwise you have a single method with different return types.

lbry/error/__init__.py
```diff
@@ -215,9 +215,10 @@ def __init__(self, url):
 
 class ResolveCensoredError(WalletError):
 
-    def __init__(self, url, censor_id):
+    def __init__(self, url, censor_id, censor_row):
         self.url = url
         self.censor_id = censor_id
+        self.censor_row = censor_row
         super().__init__(f"Resolve of '{url}' was censored by channel with claim id '{censor_id}'.")
```
eukreign, line 222: This change will get overwritten when the error file is re-generated from the `lbry/error/README.md`.

eukreign (changes requested):

PR #3206, opened 2021-02-19
provide --resolve tag for collection claim, separate from resolving its contents

jessopb:
…ts contents
collection_list was missing canonical url, etc. there was no way to get it.


lbry/extras/daemon/daemon.py
```diff
@@ -4020,11 +4022,19 @@ def jsonrpc_collection_list(self, resolve_claims=0, account_id=None, wallet_id=N
         wallet = self.wallet_manager.get_wallet_or_default(wallet_id)
         if account_id:
             account = wallet.get_account_or_error(account_id)
-            collections = account.get_collections
+            collections = account.get_collections(resolve=resolve, resolve_claims=resolve_claims)
```
eukreign, line 4025: I don't think this is correct. Have you tried running this code path? I think it will error because `collections` is expected to be a callable, not the result of a callable.

eukreign (changes requested):

lbry/extras/daemon/daemon.py
jessopb: Seems to work making it a partial() - trying with and without account_id, and with various --resolve/--resolve_claims flags.

jessopb: Working on the docs generation - just remembered I had to generate a block between create and list.

jessopb: take a look?

lbry/extras/daemon/daemon.py
```diff
@@ -3712,7 +3712,7 @@ async def jsonrpc_collection_create(
 
         Usage:
             collection_create (<name> | --name=<name>) (<bid> | --bid=<bid>)
-                    (--claims=<claims>...)
+                    (--claims=<claims>... | --claims=<claims>)
```
eukreign, line 3715: what does this change do?

lbry/extras/daemon/daemon.py
```diff
@@ -4020,11 +4022,24 @@ def jsonrpc_collection_list(self, resolve_claims=0, account_id=None, wallet_id=N
...
+            collections = partial(
+                self.ledger.get_collections,
+                wallet=wallet,
+                accounts=wallet.accounts,
+                resolve=resolve,
+                resolve_claims=resolve_claims)
+            collection_count = partial(
+                self.ledger.get_collection_count,
+                wallet=wallet,
+                accounts=wallet.accounts)
```
eukreign, line 4042: there seems to be a bit of redundancy here. adding `resolve_claims=resolve_claims` to `partial()` is not necessary because you also pass it to `paginate_rows()` which also passes it to the `get_collections` call.

scripts/generate_json_api.py
```diff
@@ -365,6 +365,26 @@ async def play(self):
 
         await self.daemon.jsonrpc_comment_abandon(reply['comment_id'])
 
+        # collections
+        collectiontom = await r(
```
eukreign, line 369: in Python the standard practice is to separate words in variables with underscores. this should be `collection_tom`, but what is a `tom`?

eukreign (changes requested):
see inline comments but also this new feature needs an integration test, please update or add to one of the tests in this class: https://github.com/lbryio/lbry-sdk/blob/a8712422bccb75d452000c02ac893f1736895d1b/tests/integration/blockchain/test_claim_commands.py#L1986

lbry/extras/daemon/daemon.py
jessopb: thought that was necessary for cli / docs generation. guess not.

eukreign (approved):

PR #3215, opened 2021-03-02
New API for "preference delete".  

### LINKED ISSUE #3214: Preference API does not support key removal
The `lbrynet preference` API only offers `get` and `set` operations.  There is no facility to `delete` a key, once it has been `set` to some value.  Suggested API signature:

        (lbry-venv) patrick@illuminati:~$ lbrynet preference delete --help

        Delete preference key/value

        Usage:
            preference_delete (<key>) [--wallet_id=<wallet_id>]

        Options:
            --key=<key> : (str) key to be deleted
            --wallet_id=<wallet_id>   : (str) restrict operation to specific wallet

        Returns:
            (bool) true, if key exists and was removed

patrickkeane:
Fixes #3214 and updates an integration test to exercise new code paths.

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -1236,6 +1236,29 @@ class Daemon(metaclass=JSONRPCServerType):
         wallet.save()
         return {key: value}
 
+    def jsonrpc_preference_delete(self, key=None, wallet_id=None):
+        """
+        Delete preference
+
+        Usage:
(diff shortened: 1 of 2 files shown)


eukreign:
thanks for the PR but as we discussed in slack, it is not necessary to add a preference delete API since we do not actually want to allow deleting preferences (makes it harder to do proper wallet data sync'ing between devices).

PR #3218, opened 2021-03-08
support for streams without data: `publish` without `--file_path` and then `claim search` using `--has_no_source`

### LINKED ISSUE #3180: Cleanly represent in-progress livestreams
## todo
- [ ] add `--has-source` and `--has-no-source` flags to `claim_search` to select only claims that do/dont have a source
- [x] make sure that when you `publish`, the `source` is not required (maybe add flag to confirm you really intended not to have a source)
- [x] make sure `claim_update` can add a source
- [ ] notify @DispatchCommit when this is live on the cluster

## what
- need a clean way to represent a livestream that's in progress
- could just be hiding streams that don't have a source. but ideally the protobuf has a clean way of saying "this is a livestream"
- if hiding live streams, odysee needs a flag to include live streams in results

## why?
- want to tie comments to a claim
- want a url that can be shared (at least after stream has been uploaded, but maybe also during the stream)

## plan
- make a claim with no Source and optionally a future release date (if you know)
- add flag to claim_search to exclude streams without a source. app will start using this anytime it needs to exclude livestreams.
- after creator goes live, they (or we, for for them) will update this claim with the recorded livestream. this adds a source and takes it from livestream to normal claim.
- (nice to have) if you want a fixed url, repost the sourceless claim to another url (maybe `/live`). the repost will redirect to your next livestream claim, and comments will be attached correctly. once you've gone live and uploaded that stream, you make a new sourcecless claim and point your fixed repost to this new claim.

jackrobison:
fixes #3180


lbry/wallet/server/db/writer.py
```diff
@@ -55,6 +55,7 @@ class SQLDB:
             description text,
 
             claim_type integer,
+            no_source bool,
```
eukreign, line 58: instead of adding a `no_source` field, would it be more flexible to add a `source` field with the actual source, that would have the added benefit of being able to search for claims with particular source?

eukreign (changes requested):

lbry/wallet/server/db/writer.py
shyba: `source` is a dictionary with [more than one field](https://github.com/lbryio/types/blob/master/v2/proto/claim.proto#L56) right? unless you have one of them in mind.

I like the idea but it would be nice to do this in a way where we don't need one column for each subitem. (something like a [flattened field](https://www.elastic.co/guide/en/elasticsearch/reference/master/flattened.html) maybe)

eukreign (approved):

PR #3232, opened 2021-03-13
Add `concurrent_hub_requests` and `hub_timeout` config options

shyba:
next step: this on speech (cc @nikooo777 )


lyoshenka:
@shyba is going to add it to the dev cluster by may 4th

lyoshenka:
gitlab license expired. will do now that gitlab is back up

PR #3236, opened 2021-03-17
limit each test time to 2 minutes, then log tasks and fail

shyba:


shyba:
done in https://github.com/lbryio/lbry-sdk/pull/3238 instead

Issue #3142, opened 2021-01-11
reorg mempool processing crash


kauffj:

tzarebczan:
This was the cause of the issues on Sunday morning

lyoshenka:
waiting for:
- app users to update to latest sdk
- turn down max_sessions
- add logging for this error
- make server to fully crash on error so we can see the log

tzarebczan:
```wallet_server_1  | ERROR:lbry.wallet.server.mempool.MemPool:MEMPOOL DIED
wallet_server_1  | Traceback (most recent call last):
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/mempool.py", line 215, in _mempool_loop
wallet_server_1  |     return await self._refresh_hashes(synchronized_event)
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/mempool.py", line 233, in _refresh_hashes
wallet_server_1  |     touched = await self._process_mempool(hashes)
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/mempool.py", line 278, in _process_mempool
wallet_server_1  |     deferred, unspent = await fetch
wallet_server_1  |   File "/usr/lib/python3.7/asyncio/tasks.py", line 533, in _wait_for_one
wallet_server_1  |     return f.result()  # May raise f.exception().
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/mempool.py", line 326, in _fetch_and_accept
wallet_server_1  |     utxos = await self.api.lookup_utxos(prevouts)
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/leveldb.py", line 790, in lookup_utxos
wallet_server_1  |     hashX_pairs = await asyncio.get_event_loop().run_in_executor(self.executor, lookup_hashXs)
wallet_server_1  |   File "/usr/lib/python3.7/concurrent/futures/thread.py", line 57, in run
wallet_server_1  |     result = self.fn(*self.args, **self.kwargs)
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/leveldb.py", line 769, in lookup_hashXs
wallet_server_1  |     return [lookup_hashX(*prevout) for prevout in prevouts]
wallet_server_1  |   File "/home/lbry/lbry/wallet/server/leveldb.py", line 769, in <listcomp>
wallet_server_1  |     return [lookup_hashX(*prevout) for prevout in prevouts]
[...]

PR #3246, opened 2021-03-26
temporary fix for mempool sync failing during reorg

eukreign:
partially addresses #3142 
see the test pasted into #3142 which was attempting to reproduce the bug without success
final fix will come in #3205


lyoshenka:
this is a good solution

PR #3247, opened 2021-03-28
adding `Access-Control-*` headers in SDK HTTP responses and support for OPTIONS HTTP request

### LINKED ISSUE #3241: JavaScript CORS problems
Using either lbry-js or just a built in `fetch` results in:

```
Access to fetch at 'http://localhost:5279/' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

--allowed-origin is set and the logs show that it is getting the request:

```
2021-03-20 01:00:30,884 DEBUG    lbry.wallet.network:59: send blockchain.claimtrie.search('text', 'offset', 'limit') to 147.135.15.197:50001 (30 timeout)
```

here is the javascript i'm using:

```
    const response = await fetch(
      "http://localhost:5279/",
      {
        method: 'POST',
        body: '{"method": "claim_search", "params": {"text": "aliens"}}',
        // mode: 'no-cors',
      },
    );
    console.log('response', await response.text());
```

with mode 'no-cors' on, the response is blank

curl on the command line does work

freakypie:
fixes #3241

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -338,6 +338,7 @@ class Daemon(metaclass=JSONRPCServerType):
         rpc_app.router.add_get('/lbryapi', self.handle_old_jsonrpc)
         rpc_app.router.add_post('/lbryapi', self.handle_old_jsonrpc)
         rpc_app.router.add_post('/', self.handle_old_jsonrpc)
+        rpc_app.router.add_options('/', self.add_cors_headers)
         self.rpc_runner = web.AppRunner(rpc_app)
 
         streaming_app = web.Application()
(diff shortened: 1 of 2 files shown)


eukreign:
@freakypie thanks for the PR. can you add a unit test which verifies that this change actually fixes the problem or at least that the OPTIONs stuff is working. we already have some existing tests related to this and you can add yours there: https://github.com/lbryio/lbry-sdk/blob/master/tests/unit/lbrynet_daemon/test_allowed_origin.py

eukreign (changes requested):
@freakypie thanks for the PR. can you add a unit test which verifies that this change actually fixes the problem or at least that the OPTIONs stuff is working. we already have some existing tests related to this and you can add yours there: https://github.com/lbryio/lbry-sdk/blob/master/tests/unit/lbrynet_daemon/test_allowed_origin.py

freakypie:
I did look at the tests you suggested @eukreign , however, I think that they aren't related enough for me to be able to easily copy and paste one (as they seem to test before a request is made). I tried to find another test that showed getting a response from the lbrynet server, but I'm not familiar enough with this so it's going to take me a while.

eukreign (approved):

PR #3255, opened 2021-04-09
Count data network events and save them + add daemon command to retrieve history

eggplantbren:
People running LBRY Desktop would like to know whether they are being a useful peer or not. If they could see numbers about their participation in the data network, they would be reassured that their setup is correctly configured, and also will feel good about leaving LBRY (or at least the daemon) running.

To that end, I have written this. The main new file is `data_network_stats.py` which manages counts in RAM which are synced to an SQLite database every two minutes (and at other times it makes sense to, such as shutdown). That database lives in the usual lbrynet directory. The result is that you can see counts of the following events, by hour:

* Blobs downloaded
* Blobs uploaded/seeded
* Blobs announced to the network by you
* Blob announcements received and stored by you
* FindNode calls handled by you
* FindValue calls handled by you

To retrieve the data, I added a daemon command `data_network_stats` which returns the hourly numbers in reverse time order. If you want to look at only recent history, you can limit the response size via an optional argument. Hours are indexed by the unix time at which they begin:

image

The extra overhead of all this is quite minimal, so I did not use asyncio at all. To do the syncing to disk, I run a loop in a separate daemon thread, because I understand basic threads and I do not understand asyncio (though I am assured that it is _just a monad_).

I hope this is helpful.


eggplantbren:
Gah, I broke your tests.

eukreign (changes requested):
Would it be possible to just extend the existing code which tracks similar stats here:

https://github.com/lbryio/lbry-sdk/blob/master/lbry/connection_manager.py#L56

eggplantbren:
I will discuss it with Tom. I'm not sure they actually have much in common.

eggplantbren:
I'm not going to work on this anymore. Best to just reject it.

eggplantbren:
Reopening this because I wouldn't mind getting it through. What specifically do you want to see? It's not obvious to me how I would put these two things together, or why it would be better.

jackrobison:
@eggplantbren 
We've got a bunch of metrics (and live monitoring) for the daemon and hub using prometheus - I think these would fit in well there. If these metrics were added where lex linked then a user can see them live with the `status` api in addition to being able to scrape them and graph with prometheus.

eggplantbren:
Thanks @jackrobison - I will work on this after trending+leveldb is done, probably concurrently with the web of trust

kauffj:
@eggplantbren in progress or close this?

PR #3260, opened 2021-04-14
ability to set `sd_hash`, `file_name` and `file_hash` when updating a stream claim

eukreign:


eggplantbren:
this seems like a bad idea because people could pretend to have files they don't have.

PR #3271, opened 2021-04-21
finished switch from using hash `#` in URLs to a colon `:` when delineating claim name from claim id

### LINKED ISSUE #3256: update resolve to return : instead of # in URLs

eukreign:
fixes #3256


lyoshenka:
to test:

run the app against new hub and see if resolve/claim_search/claim_list all work with it

PR #3276, opened 2021-04-23
Update ensure_directory_exists to check or writeability and add tests.

jeffreypicard:


lbry/extras/cli.py
```diff
@@ -294,6 +297,10 @@ def main(argv=None):
     for directory in (conf.data_dir, conf.download_dir, conf.wallet_dir):
         ensure_directory_exists(directory)
 
+    total, used, free = shutil.disk_usage('/')
```
lyoshenka, line 300: free disk space should not be required to start the sdk. and even if we did want to require this, your check misses some edge cases: maybe the download directory is on a different disk than `/`, or maybe the path is symlinked to elsewhere.

if you want to cover these cases, then it might be a worthwhile add. otherwise i'd suggest removing this check altogether.

lyoshenka (commented):
thanks for this PR @jeffreypicard!

let me know what you want to do with the disk space check i pointed out

lyoshenka (changes requested):
.

lbry/extras/cli.py
jeffreypicard: Yeah, after I made the PR I thought about the download directory being on a different disk, etc. I only included this because of the comment on the original issue "bonus: also check if disk is full when sdk starts. this check should be outside ensure_directory_exists(). it probably goes somewhere near here". I'll update the PR and remove this.

jeffreypicard: Changed my mind, I decided to try to do the checks on the directories properly. I used an IOError for the failure, which I'm not 100% sure if that's the wrote one to throw, but it seemed reasonable. Let me know if this needs any other changes!

lyoshenka:
btw [shutil.disk_usage](https://docs.python.org/3/library/shutil.html#shutil.disk_usage) may be a good way to check if the target filesystem for a download is full

eukreign:
is this still in Draft? is @lyoshenka the reviewer on this?

jeffreypicard:
> btw [shutil.disk_usage](https://docs.python.org/3/library/shutil.html#shutil.disk_usage) may be a good way to check if the target filesystem for a download is full

Yup, that's what I'm using https://github.com/lbryio/lbry-sdk/pull/3276/files#diff-abec92543a13125966ba02bafcda742a380efad323cba2649bca1d4e3eca94a1R228 I think the only questions left are what's a reasonable threshold to say the disk is too full - I set it at 1MB in the PR - and what's the right error to type to throw in that case? I used an IOError. If those things are okay this PR is ready.

lyoshenka:
there are a bunch of corner cases here so we're not gonna try to deal with the disk being full and let the OS handle that

PR #3314, opened 2021-05-31
Use new GRPC hub API if it's available

jeffreypicard:
To run the go hub tests.
Clone and run the hub from here https://github.com/lbryio/hub using branch "claim_search_port"
* `git clone git@github.com:lbryio/hub.git`
* `./dev.sh`
Alternatively you can build the hub and copy the `lbry/wallet/bin` in this repo.
* `git clone git@github.com:lbryio/hub.git`
* `go build .`
* `cp hub <lbry-sdk-location>/lbry/wallet/bin/`
Run the tests
* `export GO_HUB=true`
* `python -m unittest discover tests.integration.hub.test_hub_commands`
The claim search tests now also all work with GO_HUB=true set.
* `python -m unittest discover tests.integration.blockchain.test_claim_commands`

You can also set HUB_HOST and HUB_PORT, it defaults to localhost:50051


eukreign:
should this be marked as Draft?

jeffreypicard:
Actually, I think the hub needs to be merged and create a release first, so the HubNode can download the latest release from there if needed.

eukreign:
all of the automated tests should be green before this is moved out of Draft

lbry/wallet/network.py
```diff
@@ -432,7 +436,17 @@ async def new_resolve(self, server, urls):
             return result['result']
 
     async def new_claim_search(self, server, **kwargs):
+        if os.environ.get("GO_HUB") and os.environ.get("GO_HUB") == "true":
```
eukreign, line 439: This can probably be simplified, there are no other use cases of `new_claim_search` other than the new go hub. So if this function is being called is for the go hub.

lbry/wallet/orchstr8/node.py
```diff
@@ -429,3 +445,300 @@ def decode_raw_transaction(self, tx):
...
+                self.log.info(data.decode())
+        if b'Error:' in data:
+            self.ready.set()
+            raise SystemError(data.decode())
+        if b'listening on' in data:
+            self.ready.set()
+
+    def process_exited(self):
+        self.stopped.set()
+        self.ready.set()
```
eukreign, line 477: A lot of the above is copied from `BlockchainProcess` class specifically for handling `lbrycrd` output, does it apply to the go hub? I think this can be simplified.

lbry/wallet/orchstr8/node.py
```diff
@@ -429,3 +445,300 @@ def decode_raw_transaction(self, tx):
...
+
+        if 'fee_amount' == key:
+            value = kwargs['fee_amount']
+            value.update({"value": [str(Decimal(value['value'][0]) * 1000)]})
+            kwargs['fee_amount'] = value
+        if 'stream_types' == key:
+            kwargs['stream_type'] = kwargs.pop('stream_types')
+        if 'media_types' == key:
+            kwargs['media_type'] = kwargs.pop('media_types')
+    return kwargs
```
eukreign, line 600: What is the need for this? Shouldn't the go hub server do this conversion from JSON like request to elastic search?

lbry/wallet/orchstr8/node.py
```diff
@@ -429,3 +445,300 @@ def decode_raw_transaction(self, tx):
...
+        self.block_expected = 0
+        self.hostname = 'localhost'
+        # self.peerport = 9246 + 13  # avoid conflict with default peer port
+        self.rpcport = 50051 # avoid conflict with default rpc port
+        self.rpcuser = 'rpcuser'
+        self.rpcpassword = 'rpcpassword'
+        self.stopped = False
+        self.restart_ready = asyncio.Event()
+        self.restart_ready.set()
+        self.running = asyncio.Event()
```
eukreign, line 626: This class needs some cleanup, still lots of remnants from when it was copied from `BlockchainNode`.

lbry/wallet/orchstr8/node.py
```diff
@@ -429,3 +445,300 @@ def decode_raw_transaction(self, tx):
...
+
+    async def clear_mempool(self):
+        self.restart_ready.clear()
+        self.transport.terminate()
+        await self.protocol.stopped.wait()
+        self.transport.close()
+        self.running.clear()
+        # os.remove(os.path.join(self.data_path, 'regtest', 'mempool.dat'))
+        self.restart_ready.set()
+        await self.running.wait()
```
eukreign, line 720: does this method make sense for go hub?

eukreign (changes requested):
Is `test_hub_commands.py` testing something that isn't tested by `test_claim_commands.py`? Do we need both?

See inline comments for more feedback.

lbry/wallet/orchstr8/node.py
jeffreypicard: To simplify the protobuf for the request and minimize the amount of code on the hub side I decided to support all the varied ways of querying the hub by converting the parameters before sending them. I figured it made more sense to have one `repeated string` field for media_type, rather than a `string` field for media_type and a `repeated string` field for media_types.

eukreign: will it be easy for the app to update the code to the changed API?

@lyoshenka what are the backwards compatibility requirements for the go hub?

i'm assuming that the app will be querying the go hub to do the claim search, which could be wrong assumption...

jeffreypicard: Everything gets routed through the daemon jsonrpc calls it looks like, which is where I hooked in the hub with the environment variable check. The entire current API is supported, there's just a client side fixup of the json before doing the grpc call.

jeffreypicard:
I fixed the issues mentioned and the build works on ubuntu, both the windows and macos build fail so I commented them out for now so that I could make sure the ubuntu build would actually run and pass.

lyoshenka:
does this connect the new hub code to our CI/CD process?

jeffreypicard:
> does this connect the new hub code to our CI/CD process?

I set the environment variable GO_HUB=true in the github pipeline yaml, so as long as that works the way I think it does, it should be using my go hub for the tests.

lyoshenka:
to add:
- drop env var and always have it enabled
- drop HUB_HOST - it should use the same host as the Python hub
- drop HUB_PORT and hardcode it to 50051
- when connecting to the Python hub, also try to connect to the GRPC port 50051. if it connection works, then use that for claim_search. if not, fall back to using the existing python claim_search
  - hook into existing reconnect event to do this

lyoshenka:
todo: review the kwargs changes
- some of the changes remove features (like allowing claim_ids and not_claim_ids at the same time)
- need to have consistency between the SDK api fields and the hub api fields

lbry/extras/daemon/daemon.py
```diff
@@ -2478,30 +2598,54 @@ async def jsonrpc_claim_search(self, **kwargs):
...
+            }
+            if not kwargs.pop('no_totals', False):
+                result['total_pages'] = int((total + (page_size - 1)) / page_size)
+                result['total_items'] = total
+            return result
+        except Exception as e:
+            if self.ledger.config['use_go_hub']:
+                log.warning("failed, trying again without hub")
+                self.ledger.config['use_go_hub'] = False
+                return await self.jsonrpc_claim_search(**kwargs_old)
```
lyoshenka, line 2647: @jeffreypicard two things

1. we shouldn't modify the config like this. if another config is saved elsewhere, this setting will also be saved
2. if this fallback is disabled, a bunch of tests fail. therefore we think this is hiding test failures by falling back to old working code.

more generally, if tests pass for all the new code, then errors here should be treated like production errors -- the sdk should report the error to the user.

lyoshenka:
closing because #3394 supersedes it

PR #3346, opened 2021-06-28
change analytics, preparing it to be used by foundation endpoint

shyba:
removed: installation_id, session_id
added: analytics_id (random. once per run. like a session)
removed: external ip

missing: monitoring other calls to see what else to change.
should analytics_id be session based or installation based? if installation based, save a file like installation id?


lyoshenka:
we need to rethink how to do anon analytics

Issue #3281, opened 2021-04-26
make install doesn't install leveldb.


Kakiking:
make install does not install leveldb and it isn't mentioned in the INSTALL.md either.
Without leveldb building plyvel fails.

belikor:
With #3349, the required package `libleveldb-dev` is added to the list of requisites to install in Debian/Ubuntu.

belikor:
The previous #3349 was closed because we don't need `libleveldb-dev` to compile `plyvel`.

With #3205 or #3357 this is no longer a problem; if we use `plyvel==1.2.0` pip will download a pre-compiled version that should work fine.

Issue #2769, opened 2020-02-06
Connection reset from API when using Python 3.8


mkroman:
After some testing and some discussion with @er013 it's apparent that there's some kind of  silent incompatibility with Python `3.8.x` that's not present in Python `3.7.6`.

When `lbrynet` (`0.58.1`) is started using Python `3.8.1`, requests to the API will be met with a connection reset:

```
% python -V
Python 3.8.1
% lbrynet start    
2020-02-06 03:49:21,641 INFO     lbry.extras.daemon.daemon:450: Starting LBRYNet Daemon
2020-02-06 03:49:21,642 INFO     lbry.extras.daemon.daemon:452: Platform: {
  "processor": "",
  "python_version": "3.8.1",
  "platform": "Linux-5.5.2-arch1-1-x86_64-with-glibc2.29",
  "os_release": "5.5.2-arch1-1",
  "os_system": "Linux",
  "lbrynet_version": "0.58.1",
  "version": "0.58.1",
  "build": "dev",
  "distro": {
    "id": "arch",
    "version": "rolling",
    "version_parts": {
      "major": "",
      "minor": "",
      "build_number": ""
    },
    "like": "",
    "codename": "n/a"
  },
  "desktop": "Unknown"
}
2020-02-06 03:49:21,654 INFO     lbry.extras.daemon.daemon:461: RPC server listening on TCP ::1:5279
2020-02-06 03:49:21,654 INFO     lbry.extras.daemon.daemon:471: media server listening on TCP 127.0.0.1:5280
2020-02-06 03:49:22,200 INFO     lbry.extras.daemon.exchange_rate_manager:184: Starting exchange rate manager
2020-02-06 03:49:22,201 INFO     lbry.extras.daemon.components:457: detecting external ip
2020-02-06 03:49:22,203 INFO     lbry.extras.daemon.components:65: Loading databases
2020-02-06 03:49:22,203 INFO     lbry.wallet.database:238: connecting to database: /home/mk/.local/share/lbry/lbrynet/lbrynet.sqlite
2020-02-06 03:49:22,207 WARNING  lbry.extras.daemon.components:387: upnp discovery failed: 
2020-02-06 03:49:22,208 ERROR    lbry.extras.daemon.components:472: failed to setup upnp
2020-02-06 03:49:22,657 INFO     lbry.extras.daemon.components:476: detected external ip using lbry.com fallback
[...]

PR #3349, opened 2021-07-05
INSTALL.md: add `libleveldb-dev` to required packages

belikor:
With the provided instructions in `INSTALL.md`, the `make` command may result in errors:
```
    plyvel/_plyvel.cpp:589:10: fatal error: leveldb/db.h: No such file or directory
      589 | #include "leveldb/db.h"
          |          ^~~~~~~~~~~~~~
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

...
Failed to build plyvel prometheus-client pylru pyyaml.
```

The required header files are contained in the `libleveldb-dev` package in Ubuntu 18.04 and 20.04 at least.

After installing this package the compilation proceeds successfully.
In some cases `make` may have to be run twice.

This addresses issue #3281.

----

Also add more information on entering and exiting the virtual environment.
```
deactivate  # leave
source lbry-venv/bin/activate  # enter
```

To test the daemon interactively, we can run the `cli.py` program.
```
python lbry/extras/cli.py start
```

----

Because of issue #2769 at the moment the `lbrynet` daemon will only work correctly with Python 3.7.

If Python 3.8+ is used, the daemon will start but the RPC server may not accept messages,

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -72,6 +72,13 @@ To install on Linux/Mac:
 
 To verify your installation, `which lbrynet` should return a path inside of the `lbry-venv` folder created by the `python3.7 -m venv lbry-venv` command.
 
+To exit the virtual environment simply use the command `deactivate`.
+
+When developing, remember to enter the environment again.
+```
+source lbry-venv/bin/activate
+```
+
 ### Windows
 
 To install on Windows:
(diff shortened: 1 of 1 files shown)


shyba (approved):

belikor:
This pull request was divided into two, #3357 and #3358.

We think using `plyvel=1.2.0` is a better solution than recommending the installation of `libleveldb-dev` but keeping `plyvel=1.0.5`.

PR #3353, opened 2021-07-07
Fix `ClientSession` import in two scripts

belikor:
Nothing special, just fixing the import in two scripts.

scripts/find_max_server_load.py
scripts/test_claim_search.py
```py
from lbry.wallet.network import ClientSession
```


eukreign:
@belikor thank you for fixing the scripts!

Issue #2070, opened 2019-04-30
blob get fails and locks up SDK


tzarebczan:
## The Issue
While this is running, the sdk fails to shutdown on first attempt. 
```
C:\Users\thoma\Desktop>lbrynet blob get 17aa0c77c404bed99c84c35b1d3621244a611ef9bb522b8db7f3bacaf64c54098b5fe4c1ccabaf656cf843f916fc5e8a --timeout=5
Traceback (most recent call last):
  File "lbrynet\extras\cli.py", line 322, in <module>
  File "lbrynet\extras\cli.py", line 310, in main
  File "asyncio\base_events.py", line 573, in run_until_complete
  File "lbrynet\extras\cli.py", line 32, in execute_command
  File "site-packages\aiohttp\client.py", line 1005, in __aenter__
  File "site-packages\aiohttp\client.py", line 497, in _request
  File "site-packages\aiohttp\client_reqrep.py", line 857, in start
  File "site-packages\aiohttp\helpers.py", line 585, in __exit__
concurrent.futures._base.TimeoutError
[57460] Failed to execute script cli
```

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

belikor:
This bug is still not solved. I only added a note in #3354, so I can investigate further what is causing this error.

cristi-zz:
Hi! I just (forked) cloned lbry-sdk commit be544d6d899d2fdcfdf5e91bd091f97656de5809 and started locally in debug mode.
I can confirm, the bug is still present.

When I try to download something (eg using `get`) everything works as expected. When I am trying to download a blob using its _blob_hash_ and `blob_get` method, the daemon outputs `lbry.dht.protocol.iterative_find:379: blob peer search bottomed out`

So, @lyoshenka  I would kindly suggest opening the bug again.

@belikor are you actively working on this bug now? I will start a bit of investigation on my own, but please stop me if you found a solution/are close/etc.

[...]

PR #3354, opened 2021-07-07
scripts: note to further investigate in `download_blob_from_peer`

belikor:
Currently `lbrynet blob get <hash>` does not work to download single blobs which are not already present in the system.
The function locks up and never returns.
It only works for blobs that are in the `blobfiles` directory already.

This bug is reported in lbryio/lbry-sdk, issue #2070.

Since this script does in fact allow downloading an arbitrary blob, we may decide to copy parts of it and add them to `lbry.extras.daemon.daemon.jsonrpc_blob_get` in order to solve the previous issue.

### DIFF
--- a/scripts/download_blob_from_peer.py
+++ b/scripts/download_blob_from_peer.py
@@ -1,3 +1,19 @@
+"""A simple script that attempts to directly download a single blob.
+
+To Do:
+------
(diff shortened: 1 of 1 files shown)


lyoshenka:
thanks @belikor :+1:

PR #3358, opened 2021-07-11
Improve the instructions in `INSTALL.md`

belikor:
Because of issue #2769 at the moment the `lbrynet` daemon will only work correctly with Python 3.7.

The `deadsnakes` personal package archive (PPA) provides Python 3.7 for Ubuntu distributions that no longer have it in their official repositories like 18.04 and 20.04.

If Python 3.8+ is used, the daemon will start but the RPC server may not accept messages, returning the following:
```
Could not connect to daemon. Are you sure it's running?
```

----

Add more information regarding virtual environments.

Enter the environment with
```
source lbry-venv/bin/activate
```

And leave with `deactivate`.

When developing, we can start the server interactively.
```
python lbry/extras/cli.py start
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -70,7 +70,10 @@ To install on Linux/Mac:
  $ make idea
  ```
 
-To verify your installation, `which lbrynet` should return a path inside of the `lbry-venv` folder created by the `python3.7 -m venv lbry-venv` command.
+To verify your installation, `which lbrynet` should return a path inside
+of the `lbry-venv` folder.
+
+To exit the virtual environment simply use the command `deactivate`.
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


eukreign:
We don't currently support Python 3.8 so I'm not sure it makes sense to add to docs instructions on how to setup with it. Let's figure out and fix the issues with SDK on Python 3.8 before offering it in the docs as an option.

eukreign (changes requested):
We don't currently support Python 3.8 so I'm not sure it makes sense to add to docs instructions on how to setup with it. Let's figure out and fix the issues with SDK on Python 3.8 before offering it in the docs as an option.

belikor:
> We don't currently support Python 3.8 so I'm not sure it makes sense to add to docs instructions on how to setup with it. 

For reference, this pull request does not explain how to set up Python 3.8, it just explicitly mentions that it does not work with Python 3.8. That was my intention.

If the other problems with `aiohttp` are solved then we can reword this to mention Python 3.7, 3.8, and 3.9.

PR #3359, opened 2021-07-12
cli.py: print `ClientConnectionError` exception

belikor:
As reported in issue #2769 and #3358, the `lbrynet` daemon doesn't respond correctly when it is compiled against Python 3.8+; it only work correctly with Python 3.7.

In this case it will raise the exception `aiohttp.ClientConnectionError`.

We want to print this exception for the future to know what went wrong and be able to troubleshoot.


eukreign:
We don't want to show a python exception to the user whenever the cli fails to connect to a daemon. Let's just upgrade `aiohttp` so that this isn't an issue.

PR #3362, opened 2021-07-14
setup.py: use `aiohttp==3.6.2` instead of `aiohttp==3.5.4`

belikor:
As reported in issue #2769, the `lbrynet` daemon's RCP server doesn't respond correctly when it is compiled against Python 3.8+, only with Python 3.7.

Instead of using `aiohttp==3.5.4` which was released on January 12, 2019, we use `aiohttp=3.6.2` which was released October 9, 2019, and is available in Ubuntu 20.04.

By using this newer version, the RCP server seems to work correctly both with Python 3.8 and 3.9.

We could also try a newer version, like `3.7.4` as it was launched on February 25, 2021, and is available in Arch.


eukreign (changes requested):
This change appears to cause the `tests / integration (datanetwork)` integration tests to consistently fail. There must be something in `3.6.2` that is not backwards compatible with the existing SDK code. Can you investigate?

belikor:
> This change appears to cause the `tests / integration (datanetwork)` integration tests to consistently fail. There must be something in `3.6.2` that is not backwards compatible with the existing SDK code. Can you investigate?

I'm not entirely sure how to run the integration tests in my machine, so this may take a while. I don't see any obvious errors in the logs, other than Timeouts. We could try other versions of `aiohttp` perhaps, 3.6.3, 3.7.4, etc.

lyoshenka:
to run integration tests, run `make test-integration`

lyoshenka:
@belikor are you still planning to work on this?

belikor:
> @belikor are you still planning to work on this?

The integration tests pass okay in my machine (`tox`), so at the moment I'm not sure what to test further.

lyoshenka:
@eukreign whats the next step?

eukreign:
@lyoshenka next step would be to rebase and see if the tests pass, if they don't pass @belikor needs to figure out why they are failing

alternately, there is another PR to simply upgrade the SDK to newer Python altogether (#3404 ) but that PR is also failing (due to DHT tests not working with newer Python version)

eukreign (changes requested):
tests need to pass before this can be merged

belikor:
This is a very small change, which just changes aiohttp from 3.5.4 to 3.6.2. The only test that seems to fail is `test_download_torrent`.

```
python -m unittest tests.integration.datanetwork.test_file_commands.FileCommands.test_download_torrent
```

This test is actually pretty short, maybe some of you can figure out why it fails because the error message doesn't give much information of where it fails. The failure seems to occur in the async calls, so maybe not in the LBRY code itself.
```
  File "/home/runner/work/lbry-sdk/lbry-sdk/.tox/datanetwork/lib/python3.7/site-packages/lbry/testcase.py", line 143, in run
    self.loop.run_until_complete(maybe_coroutine)
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
concurrent.futures._base.CancelledError
```

What I notice is that this test is optional; it only runs if `libtorrent` is installed. As far as I can tell, LBRY provides their own copy of `libtorrent` ([lbry-libtorrent](https://github.com/lbryio/libtorrent)). Maybe we need to update this code as well because it seems it was last updated early 2020. Maybe a newer `libtorrent` will play better with a newer `aiohttp`.

Issue #2342, opened 2019-07-24
Claim search does not handle invalid arguments


jackrobison:
When any_tags=[]` or `fee_amount=None` claim search will fail with an internal server error. It's likely there are many other odd values for keyword arguments that will fail the same way.

nikooo777:
```
./lbrynet claim search --claim_id=331ae805223be359a3ada29921b980088f85370c --txid="" --nout="" --name=""
```
this will trigger it as well

belikor:
> ```
> ./lbrynet claim search --claim_id=331ae805223be359a3ada29921b980088f85370c --txid="" --nout="" --name=""
> ```
> 
> this will trigger it as well

The error I see with this is caused by `--claim_id`, which is probably different from other errors seen in the original post.
```
...
    "name": "ValueError",
    "traceback": [
      "Traceback (most recent call last):",
      "  File \"lbry/extras/daemon/daemon.py\", line 693, in _process_rpc_call",
      "  File \"lbry/extras/daemon/daemon.py\", line 2483, in jsonrpc_claim_search",
      "ValueError: Only 'claim_id' or 'claim_ids' is allowed, not both.",
      ""
    ]
  },
  "message": "Only 'claim_id' or 'claim_ids' is allowed, not both."
}
```

PR #3367, opened 2021-07-18
fix error when using `--claim_id` with `lbrynet claim search`

belikor:
For some reason, when using `claim_search` with `--claim_id`, the arguments dictionary will also contain `claim_ids` with an empty list, even if we didn't specify it.
```
lbrynet claim search --claim_id=8945573bcfcb7f8276187dfbb93545eac4ebf71a
```
```
{
...
    "name": "ValueError",
    "traceback": [
      "Traceback (most recent call last):",
      "  File \"lbry/extras/daemon/daemon.py\", line 693, in _process_rpc_call",
      "  File \"lbry/extras/daemon/daemon.py\", line 2483, in jsonrpc_claim_search",
      "ValueError: Only 'claim_id' or 'claim_ids' is allowed, not both.",
      ""
    ]
  },
  "message": "Only 'claim_id' or 'claim_ids' is allowed, not both."
}
```

Using both `claim_id` and `claim_ids` will raise a `ValueError` exception so the daemon won't return a valid result even if the claim ID is in fact valid.

So if `claim_id` exists, we need to discard `claim_ids` if it is empty, before proceeding with the rest of the code.

On the other hand, if `claim_ids` is used, and `claim_id` is absent, there will be no problem as `claim_id` won't be added to the dictionary.
```
lbrynet claim search --claim_ids=8945573bcfcb7f8276187dfbb93545eac4ebf71a
```

Related #2342

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -2479,6 +2479,8 @@ class Daemon(metaclass=JSONRPCServerType):
         Returns: {Paginated[Output]}
         """
         wallet = self.wallet_manager.get_wallet_or_default(kwargs.pop('wallet_id', None))
+        if "claim_ids" in kwargs and not kwargs["claim_id"]:
+            kwargs.pop("claim_ids")
         if {'claim_id', 'claim_ids'}.issubset(kwargs):
             raise ValueError("Only 'claim_id' or 'claim_ids' is allowed, not both.")
         if kwargs.pop('valid_channel_signature', False):


lbry/extras/daemon/daemon.py
```diff
@@ -2479,6 +2479,10 @@ async def jsonrpc_claim_search(self, **kwargs):
         Returns: {Paginated[Output]}
         """
         wallet = self.wallet_manager.get_wallet_or_default(kwargs.pop('wallet_id', None))
+        if ("claim_ids" in kwargs and not kwargs["claim_ids"]
+                and "claim_id" in kwargs and kwargs["claim_id"]):
+            # Discard `claim_ids` when it is empty, otherwise `ValueError`.
+            kwargs.pop("claim_ids")
```
eukreign, line 2485: This can be simplified, don't need the second check and don't need a redundant comment (it's obvious below what's happening: an empty value is popped):
```
        if "claim_ids" in kwargs and not kwargs["claim_ids"]:
            kwargs.pop("claim_ids")
```

eukreign (changes requested):

lbry/extras/daemon/daemon.py
belikor: Well, I only want to discard `'claim_ids'` if `'claim_id'` is present, as if the latter is not present, there is no need to discard it, even if empty, hence the second test.

PR #3368, opened 2021-07-20
Fix blob_get() would timeout. Issue #2070

cristi-zz:
Tentative fixing. The node is primed with some fixed peers before triggering the blob download.

Please check comments on issue #2070

### DIFF
--- a/lbry/dht/node.py
+++ b/lbry/dht/node.py
@@ -257,3 +257,10 @@ class Node:
                          ) -> typing.Tuple[asyncio.Queue, asyncio.Task]:
         queue = peer_queue or asyncio.Queue(loop=self.loop)
         return queue, self.loop.create_task(self._accumulate_peers_for_value(search_queue, queue))
+
+
+async def get_kademlia_peers_from_hosts(peer_list: typing.List[typing.Tuple[str, int]]) -> typing.List['KademliaPeer']:
+    peer_address_list = [(await resolve_host(url, port, proto='tcp'), port) for url, port in peer_list]
+    kademlia_peer_list = [make_kademlia_peer(None, address, None, tcp_port=port, allow_localhost=True)
+                          for address, port in peer_address_list]
+    return kademlia_peer_list
--- a/lbry/stream/downloader.py
+++ b/lbry/stream/downloader.py
@@ -3,9 +3,9 @@ import typing
 import logging
 import binascii
 
-from lbry.dht.peer import make_kademlia_peer
+from lbry.dht.node import get_kademlia_peers_from_hosts
 from lbry.error import DownloadSDTimeoutError
-from lbry.utils import resolve_host, lru_cache_concurrent
+from lbry.utils import lru_cache_concurrent
 from lbry.stream.descriptor import StreamDescriptor
 from lbry.blob_exchange.downloader import BlobDownloader
 if typing.TYPE_CHECKING:
@@ -48,26 +48,19 @@ class StreamDownloader:
         self.cached_read_blob = cached_read_blob
 
     async def add_fixed_peers(self):
-        def _delayed_add_fixed_peers():
+        def _add_fixed_peers(fixed_peers):
+            self.peer_queue.put_nowait(fixed_peers)
             self.added_fixed_peers = True
-            self.peer_queue.put_nowait([
-                make_kademlia_peer(None, address, None, tcp_port=port, allow_localhost=True)
-                for address, port in addresses
-            ])
 
         if not self.config.fixed_peers:
             return
-        addresses = [
-            (await resolve_host(url, port, proto='tcp'), port)
-            for url, port in self.config.fixed_peers
-        ]
         if 'dht' in self.config.components_to_skip or not self.node or not \
                 len(self.node.protocol.routing_table.get_peers()) > 0:
             self.fixed_peers_delay = 0.0
         else:
             self.fixed_peers_delay = self.config.fixed_peer_delay
-
-        self.fixed_peers_handle = self.loop.call_later(self.fixed_peers_delay, _delayed_add_fixed_peers)
+        fixed_peers = await get_kademlia_peers_from_hosts(self.config.fixed_peers)
+        self.fixed_peers_handle = self.loop.call_later(self.fixed_peers_delay, _add_fixed_peers, fixed_peers)
 
     async def load_descriptor(self, connection_id: int = 0):
         # download or get the sd blob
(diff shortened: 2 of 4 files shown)


tzarebczan:
@cristi-zz nice to see your face here - we've come full circle since, huh! 

Thanks for the PR and congrats on the first one, we'll get this reviewed soon. 

Can we show you some [appreciation](https://lbry.com/faq/appreciation)?

cristi-zz:
@tzarebczan thank you for the appreciation!

This bug was on my mind for so long! However the fix is ugly (duplicate code, etc), some guidance is needed to proper fix this. My guess is that peers must be added to the Node object before interacting with Kademlia, and this "priming" should happen at the startup of the library and not be rebooted each time an API call finishes. Now, the peers are added in some API handling routines (download a stream with `get` ) but not in others (like `blob_get`). Don't know what is the design intention so I can't dig further.

cristi-zz:
Ready for review. All tests (int+unit) are green on local machine.

lyoshenka:
@jackrobison please review this issue. it's been open for over a month

lbry/dht/node.py
```diff
@@ -258,3 +258,13 @@ def accumulate_peers(self, search_queue: asyncio.Queue,
...
+
+
+async def prime_node_with_peer_addresses(peer_list: typing.List[typing.Tuple[str, int]], peer_queue: asyncio.Queue):
+    try:
+        peer_address_list = [(await resolve_host(url, port, proto='tcp'), port) for url, port in peer_list]
+        kademlia_peer_list = [make_kademlia_peer(None, address, None, tcp_port=port, allow_localhost=True)
+                              for address, port in peer_address_list]
+        peer_queue.put_nowait(kademlia_peer_list)
+    except asyncio.CancelledError:
+        pass
```
jackrobison, line 270: why is the cancelled error trapped? usually trapping this error causes unexpected behavior.

jackrobison (changes requested):
I'm not understanding what the desired change in behavior here is. It looks like the `call_later` is refactored to be a `asyncio.Task` with a `asyncio.sleep` instead of the call_later delay - which should behave essentially the same.

Can you explain a bit what the problem is / what the desired behavior looks like?

lbry/dht/node.py
cristi-zz: answer in the comments of this PR

cristi-zz:
Hi! Thank you for looking into this! This bug, is bugging me for a while, now.

Overall intention:

Make RPC call ``blob_get`` connect to some peers and get the actual blobs. Imho there were some refactoring on the way the DHT operations are handled and this particular method was left out. Fixing it directly would result in duplicate code (true, 2-3 lines only but still, a new source of grief). So I tried to extract the common behavior so future changes in the semantics of DHT will not break these calls. 

Mechanics:
In the original code the ``_delayed_add_fixed_peers`` was a regular method that was called after ``fixed_peers_delay``. Also, the handle returned by ``call_later`` is used to cancel this task [why the delay? why the need to cancel? Not important right now, maybe]. I tried to reproduce the exact behavior , constrained by the facts:
(1) the _delayed_add_fixed_peers must call an async code (so it must be also made async. The ``call_later`` does not take a coroutine) and (2) not being an expert on asyncio. 

Reading [loop.call_later](https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.loop.call_later) docs I assume that if the task is cancelled before the delay, no exception is thrown. I THINK I tried to mimic the same behavior in ``prime_node_with_peer_addresses`` by catching only the Cancelled exception. Looking back at the code, maybe is the wrong place to catch it? Where this exception would propagate? To ``add_fixed_peers``? Is there a more pythonic way to add peers after a delay while being able to cancel the whole operation? 

Suggestions are more than welcome wrt to this issue!

lbry/stream/downloader.py
```diff
@@ -48,26 +48,18 @@ async def cached_read_blob(blob_info: 'BlobInfo') -> bytes:
...
-            for url, port in self.config.fixed_peers
-        ]
         if 'dht' in self.config.components_to_skip or not self.node or not \
                 len(self.node.protocol.routing_table.get_peers()) > 0:
             self.fixed_peers_delay = 0.0
         else:
             self.fixed_peers_delay = self.config.fixed_peer_delay
-
-        self.fixed_peers_handle = self.loop.call_later(self.fixed_peers_delay, _delayed_add_fixed_peers)
+        self.fixed_peers_handle = self.loop.call_later(self.fixed_peers_delay, asyncio.create_task, _add_fixed_peers())
```
cristi-zz, line 62: imho it would be cleaner to have ``sleep`` followed by ``_add_peers`` and no cancel handle. But maybe I am missing some important info and the cancelling handle should stay in place. Anyway, it is more asyncio-ish than before.

cristi-zz:
Hi! @jackrobison !

Refactored the offending line [see comments on commit] and removed the try/catching the Cancel exception. Also don't know if you get notified at comments or only when "resolve conversation" is pressed. So sorry if I spammed you with tags.

Cheers!

tests/unit/blob_exchange/test_transfer_blob.py
```diff
@@ -326,3 +329,33 @@ async def sendfile(writer):
...
+        with mock.patch('lbry.extras.daemon.componentmanager.ComponentManager.all_components_running',
+                        return_value=True):
+            with mock.patch('lbry.extras.daemon.daemon.Daemon.dht_node', new_callable=mock.PropertyMock) \
+                    as daemon_mock_dht:
+                with mock.patch('lbry.extras.daemon.daemon.Daemon.blob_manager', new_callable=mock.PropertyMock) \
+                        as daemon_mock_blob_manager:
+                    daemon_mock_dht.return_value = mock_node
+                    daemon_mock_blob_manager.return_value = self.client_blob_manager
+                    result = await daemon.jsonrpc_blob_get(blob_hash)
+                    self.assertIsNotNone(result)
```
jackrobison, line 361: this could return an error and still pass - should check that the result is expected

cristi-zz: Checked the downloaded blob against the "original" one, injected in the server. HOWEVER, the original bug is timeout. Didn't look into the codebase to see how it is done here, and saw that unittest does not handle the timeout natively [sth like timeout_decorator]. CI will handle timeouts anyway so this test will fail on CI (if ofc the peers will not be added correctly). Not local maybe. Want to add a depedency? Hack/simulate the timeout check using some async? (make a code that sleep for x seconds and assertFalse. Kill the future after blob_get is done). Or good enough for now?

Also, saw some coverage reductions, but I guess they are from incoming code.

cristi-zz:
Hm, it worked on my github instance, Maybe retrigger the CI pipeline?

https://github.com/cristi-zz/lbry-sdk/actions/runs/1420543979

PR #3370, opened 2021-07-20
Split `Daemon` class into individual classes

belikor:
Separating the big `daemon.Daemon` class into smaller classes `daemon.daemon_*.Daemon_*` makes it easier to modify each method without affecting the rest of the code.

In this case, the individual classes are "parent" classes, while the `Daemon` class is a "child" class that inherits all the `jsonrpc_*` methods from the other classes.

Instead of
```py
class Daemon(metaclass=JSONRPCServerType):
```

We have
```py
class Daemon(Daemon_base, Daemon_get, Daemon_settings, Daemon_wallet,
             Daemon_account, Daemon_sync, Daemon_address, Daemon_file,
             Daemon_purchase, Daemon_claim, Daemon_channel,
             Daemon_stream, Daemon_collection, Daemon_support,
             Daemon_transaction, Daemon_txo, Daemon_utxo, Daemon_blob,
             Daemon_peer, Daemon_tracemalloc, Daemon_comment):
```

Each class itself derives from the metaclass `JSONRPCServerType` in `daemon.daemon_meta`.

This is a mock implementation; the code hasn't been fully tested to work in all situations, although it seems to pass the unit tests.


eukreign:
You've put a lot of effort into this PR @belikor, thank you for contributing so much to the SDK. Unfortunately, as we discussed in slack, this restructuring doesn't really fit with the goals of this API class since it's intended to be a flat public API and therefore it makes sense to keep it simple and in one class/file. Also, there are inter-dependencies between some of these API methods and therefore by splitting them up we are also introducing new dependency complexity which wasn't there before. Thank you and please continue to contribute; in the future I would suggest discussing any major initiatives like this one with us before putting in the work to make sure it's something we will want to merge into master.

belikor:
Sure, I mean, this was just a quick effort to showcase some ideas, it wasn't intended to be merged.

PR #3371, opened 2021-07-20
rename ES fields with _hash to _id and remove redundants

shyba:


lyoshenka (approved):

lbry/wallet/server/db/elasticsearch/search.py
```diff
@@ -484,9 +486,8 @@ def expand_query(**kwargs):
             if not many and key in ('_id', 'claim_id') and len(value) < 20:
                 partial_id = True
             if key == 'public_key_id':
-                key = 'public_key_hash'
                 value = Base58.decode(value)[1:21].hex()
-            if key == 'signature_valid':
+            if key in ('signature_valid', 'has_source'):
```
lyoshenka, line 490: shoudln't this be `is_signature_valid`

lyoshenka: or is this after the transformation?

shyba: `signature_valid` is part of the public API, so it comes in the request. It gets translated to `is_signature_valid` some lines later, when generating the final ES query from the API claim_search parameters

PR #3378, opened 2021-07-30
ability to limit disk spaced used for blobs via `blob_storage_limit` setting (oldest blobs are deleted when disk space limit is reached)

eukreign:
disable the component in server environment: `disk_space`


tzarebczan:
A must have would be to allow moving of the data directory completely or blob files (along with data / any db changes) : https://github.com/lbryio/lbry-sdk/issues/2723

Issue #3381, opened 2021-08-02
`claim_search` has a limit of 50 for `page_size`


belikor:
This is similar to #3365 in that there seems to be a claim limit when searching.

```
lbrynet claim search helplbrysavecrypto
```
```
{...
  [...
  ],
  "page": 1,
  "page_size": 20,
  "total_items": 269,
  "total_pages": 14
}
```

If we specify a large `page_size` so that all items are included in a single page, the `page_size` is set to 50 maximum.
```
lbrynet claim search helplbrysavecrypto --page_size=269
```
```
{...
  [...
  ],
  "page": 1,
  "page_size": 50,
  "total_items": 269,
  "total_pages": 6
}
```

Using a page size as large as `total_items` is necessary to be able to get all claims that are competing for the same `name`. At the moment the limit is 50, so 50 claims maximum can be compared. This is not such a big deal when most of those claims are reposts, but it may be a problem if those are meant to be unique streams.

Comparing claims that have the same name was inspired by the following script: https://odysee.com/@BrendonBrewer:3/vanity:8

shyba:
This is the intended behavior.

Above 50 it would be too much to transfer for each request, so the maximum is 50, but that should not impact your use case. You should be able to get all results by specifying `--page=<n>` across all pages.

belikor:
> You should be able to get all results by specifying `--page=<n>` across all pages.

So, if I understand correctly, there is no way to get all items in a single call. The only way is to get batches of 50, and iterate over all pages to get all items.

If this is the case, then the `page_size` option is not very useful because you can only select between 1 and 50 items at a time. Why not make it fixed so that you don't have the illusion that you can make the page size arbitrarily bigger?

shyba:
Yes, there is no way to get everything on a single call.

The downside is waste. Sometimes you want to fetch the single lowest/highest for some sorting criteria or a small section of 5 claims for some category. If you have a page doing that a lot, it is far more efficient to transfer 1-5 for each query than all 50 to use 1.

For the upper limit, it impacts the server: Imagine 100 users querying together. If one of them is able to ask 1,000 results in a single trip it take away resources from the others. So you need to go back to the end of the queue multiple times and give a turn for the others to do their queries too as you dig into all results. On the other hand, if the server is idle, going back to the queue 100 times can be wasteful, but it is never the case for the current usage pattern.

Issue #3365, opened 2021-07-15
Remove 1K claim search overall limit


tzarebczan:
You can't go past page 50 on claim search, even with 1 channel. This should at least work in the case of a single channel search.

belikor:
The `@mises` channel has around 1179 claims as of this moment, but it only shows 1000 claims distributed as 50 pages with 20 items per page.
```
lbrynet claim search --channel=@mises
```
```
{...
  [...
  ],
  "page": 1,
  "page_size": 20,
  "total_items": 1000,
  "total_pages": 50
}
```

We can specify the page size larger than the total number of items. In this case we would expect to have only 1 page with `page_size` being 2000. However, the page size is limited to 50, and the number of pages now is 20, accommodating a total of 1000 items, same as before.
```
lbrynet claim search --page_size=2000 --channel=@mises
```
```
{...
  [...
  ],
  "page": 1,
  "page_size": 50,
  "total_items": 1000,
  "total_pages": 20
}
```

----

The page size seems to be controlled by the minimum value of `DEFAULT_PAGE_SIZE`, which is 20, and 50.

Or if `page_size` is given, it will be the minimum of `page_size` or 50.

https://github.com/lbryio/lbry-sdk/blob/1ec74a89e22e1b7b0161943d58ee59c56cd06b12/lbry/extras/daemon/daemon.py#L2492

If, instead of using `min`, we use `max`, then it will set the `page_size` to the specified large value. However, this will still not result in more than 1000 items.
```
        page_num, page_size = abs(kwargs.pop('page', 1)), max(abs(kwargs.pop('page_size', DEFAULT_PAGE_SIZE)), 50)
```
```
{...
  [...
  ],
  "page": 1,
  "page_size": 2000,
  "total_items": 1000,
  "total_pages": 1
}
```

----

The actual search seems to be performed by the `Ledger` class.

https://github.com/lbryio/lbry-sdk/blob/1ec74a89e22e1b7b0161943d58ee59c56cd06b12/lbry/extras/daemon/daemon.py#L2494

This `Ledger.claim_search` in turn seems to be executed by the `Network` class.

https://github.com/lbryio/lbry-sdk/blob/1ec74a89e22e1b7b0161943d58ee59c56cd06b12/lbry/wallet/ledger.py#L879-L890

This `Network.claim_search` in turn seems to be a `Network.rpc` call to `'blockchain.claimtrie.search'`. Is this the blockchain?

[...]

PR #3384, opened 2021-08-03
Daemon.jsonrpc_claim_search: `page_size` is no longer a minimum value

belikor:
This addresses issue #3381, and partially #3365.

Currently with `lbrynet claim search` when `--page_size` is specified, it will use the minimum of this value or 50. So, if it's `--page_size=1000` it will use 50.

If `--page_size` is omitted, it will use the minimum of `DEFAULT_PAGE_SIZE` (normally 20) or 50.

So, instead of using `min` we use `max`, meaning that a larger `page_size` value will be used. This will allow us to fit as many elements as possible in the claim search.

However, there is still a separate bug, issue #3365, which limits the total results to 1000, no matter how big `page_size` is.

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -2507,7 +2507,7 @@ class Daemon(metaclass=JSONRPCServerType):
             kwargs['signature_valid'] = 0
         if 'has_no_source' in kwargs:
             kwargs['has_source'] = not kwargs.pop('has_no_source')
-        page_num, page_size = abs(kwargs.pop('page', 1)), min(abs(kwargs.pop('page_size', DEFAULT_PAGE_SIZE)), 50)
+        page_num, page_size = abs(kwargs.pop('page', 1)), max(abs(kwargs.pop('page_size', DEFAULT_PAGE_SIZE)), 50)
         kwargs.update({'offset': page_size * (page_num - 1), 'limit': page_size})
         txos, blocked, _, total = await self.ledger.claim_search(wallet.accounts, **kwargs)
         result = {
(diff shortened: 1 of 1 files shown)


belikor:
Integration tests that don't pass
```
python -m unittest tests.integration.blockchain.test_claim_commands.ClaimSearchCommand.test_pagination

python -m unittest tests.integration.blockchain.test_claim_commands.ClaimSearchCommand.test_limit_claims_per_channel_across_sorted_pages
```

These tests compare essentially two lists, a list of claims that are created for this test, and the return value of `claim_search`.

Since we changed the `page_size`, the number of items returned per page is different, and thus the tests don't pass cleanly. So we need to update the tests, or reject the change to `page_size`.

belikor:
Instead of using the minimum of `page_size` or `DEFAULT_PAGE_SIZE` or 50, we simply use `page_size` if it exists, and `DEFAULT_PAGE_SIZE` if `page_size` doesn't exist.

This respects `page_size` always, and passes all integration tests.

However, there is still a separate bug, issue #3365, which limits the total results to 1000, no matter the `page_size`.
```
lbrynet claim search --channel=@mises --page_size=1500
```
```
{
  [...
  ],
  "page": 1,
  "page_size": 1500,
  "total_items": 1000,
  "total_pages": 1
}
```

lyoshenka:
given the discussion in #3381, it sounds like we don't want to increase the max `page_size`.  but we do want to be able to paginate to more than 1000 claims, right (#3365)?

belikor:
Yes, we definitely want to be able to get more than 1000 results, regardless of the page size.

The `SearchIndex.search_ahead` method seems to limit the number of results to 1000 by specifying `kwargs['limit'] = 1000`.

https://github.com/lbryio/lbry-sdk/blob/d09bfdc4ffc7adf27f77bef1477904f779084167/lbry/wallet/server/db/elasticsearch/search.py#L260-L266

Issue #3389, opened 2021-08-05
claim list + name param - case sensitive ? 


tzarebczan:
If you pass the name param on claim list, it's case sensitive. i.e. if you have a claim called Test, and you pass test, it won't find it. I don't think our other apis are case sensitive this way.

belikor:
If I understand the description correctly, these two return different results.

This works.
```
lbrynet claim list --name=helplbrysavecrypto
```

This doesn't return the correct result.
```
lbrynet claim list --name=HelpLBRYsavecryptO
```

What other methods work case insensitive?

This seems to work with 268 items found.
```
lbrynet claim search --name=helplbrysavecrypto
```

This also works, even if it contains some capital letters.
```
lbrynet claim search --name=HelpLBRYsavecryptO
```

----

The second command, that is case insensitive, apparently uses `elasticsearch` to perform the search. This is explained in #3365. The parameters of the search, such as `name`, are contained in the `kwargs` dictionary.

https://github.com/lbryio/lbry-sdk/blob/8304102136526e42000246a05255ae942dda5083/lbry/wallet/server/db/elasticsearch/search.py#L260-L275

The `kwargs` dictionary is resolved by `expand_query`. If `name` is included, it is passed to `lbry.schema.url.normalize_name`.

https://github.com/lbryio/lbry-sdk/blob/8304102136526e42000246a05255ae942dda5083/lbry/wallet/server/db/elasticsearch/search.py#L454-L460

This function uses `unicodedata.normalize` with normal form D (NFD) to normalize the unicode name. Then it uses `str.casefold()` to return a caseless string. So the `name` is always searched in this normalized, caseless form.

https://github.com/lbryio/lbry-sdk/blob/8304102136526e42000246a05255ae942dda5083/lbry/schema/url.py#L45-L46

----

The method `jsonrpc_claim_list` simply passes the arguments to `lbry.extras.daemon.daemon.Daemon.jsonrpc_txo_list`

https://github.com/lbryio/lbry-sdk/blob/8304102136526e42000246a05255ae942dda5083/lbry/extras/daemon/daemon.py#L2259

https://github.com/lbryio/lbry-sdk/blob/8304102136526e42000246a05255ae942dda5083/lbry/extras/daemon/daemon.py#L2289-L2292

The `jsonrpc_txo_list` method passes the arguments to the static method `lbry.extras.daemon.daemon.Daemon._constrain_txo_from_kwargs`.

[...]

PR #3390, opened 2021-08-09
daemon: `_constrain_txo_from_kwargs` normalize the `name`

belikor:
This addresses issue #3389.

Claim list is normally case sensitive, but with this fix it becomes case insensitive, just like `claim_search`.

To test this, the claim `helplbrysavecrypto` must be reposted first.

Then these two should work.
```
lbrynet claim list --name=helplbrysavecrypto
lbrynet claim list --name=HELPLBRYSAVECRYPTO
```

Just as these two work.
```
lbrynet claim search --name=helplbrysavecrypto
lbrynet claim search --name=HELPLBRYSAVECRYPTO
```
----

If the argument `--name` is passed to `jsonrpc_claim_list`, it will be passed to `jsonrpc_txo_list` which passes the argument to `_constrain_txo_from_kwargs`.

The `name` value is passed as is to `lbry.wallet.database.constrain_single_or_list` meaning that it will be case sensitive.

This is different from `jsonrpc_claim_search` which sends the query to `elasticsearch.search.expand_query`. The `name` argument is passed to `lbry.schema.url.normalize_name` to obtain a normalized version of the name, meaning some substitutions will be made, and it will be lowercase.

Now the `name` argument in `_constrain_txo_from_kwargs` is also passed to `lbry.schema.url.normalize_name`, so now `claim_list` will behave like `claim_search` when using `--name`, and will accept mixed case names.

The functions affected by the normalization are `jsonrpc_txo_list` (which affects `jsonrpc_claim_list`), `jsonrpc_txo_spend`, `jsonrpc_txo_sum`, and `jsonrpc_txo_plot`.

The method `jsonrpc_txo_list` will also affect `jsonrpc_channel_list`, `jsonrpc_stream_list`, `jsonrpc_support_list`, and `jsonrpc_utxo_list` only when they use the `--name` argument.

----

If we don't want the `--name` to be case-insensitive in all these functions, then `_constrain_txo_from_kwargs` needs to be modified to be case sensitive some times, and insensitive other times.

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -51,7 +51,7 @@ from lbry.extras.daemon.undecorated import undecorated
 from lbry.extras.daemon.security import ensure_request_allowed
 from lbry.file_analysis import VideoFileAnalyzer
 from lbry.schema.claim import Claim
-from lbry.schema.url import URL
+from lbry.schema.url import URL, normalize_name
 
 if typing.TYPE_CHECKING:
     from lbry.blob.blob_manager import BlobManager
@@ -2270,7 +2270,7 @@ class Daemon(metaclass=JSONRPCServerType):
             --claim_type=<claim_type>  : (str or list) claim type: channel, stream, repost, collection
             --claim_id=<claim_id>      : (str or list) claim id
             --channel_id=<channel_id>  : (str or list) streams in this channel
-            --name=<name>              : (str or list) claim name
+            --name=<name>              : (str or list) claim name (normalized)
             --is_spent                 : (bool) shows previous claim updates and abandons
             --account_id=<account_id>  : (str) id of the account to query
             --wallet_id=<wallet_id>    : (str) restrict results to specific wallet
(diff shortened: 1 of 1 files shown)


eukreign (changes requested):
@belikor thanks for the PR, two comments:

1. the database `claim_name` column is not a normalized version of the name, therefore this change would actually prevent from finding your claims because it would be matching the non-normalized name to the now normalized name coming via API call.
2. the doc strings shouldn't be updated because the user is not being asked to provide only normalized names, the normalization happens for them and they can pass arguments in any form.

belikor:
> 1. the database `claim_name` column is not a normalized version of the name, therefore this change would actually prevent from finding your claims because it would be matching the non-normalized name to the now normalized name coming via API call.

I think I understand what you mean. So if it is saved like `ALLCAPS`, the search will be performed with `allcaps`, and it won't be found. Is this what you mean?

In that case, what would be the solution? Can the names in the database be normalized before searching? Or can all names be saved both normalized and non-normalized? Does this make sense?

This is a quick fix, of course, I didn't consider all the implications of the code. I just saw the report by Tom, and I tried to fix it quickly.

> 2. the doc strings shouldn't be updated because the user is not being asked to provide only normalized names, the normalization happens for them and they can pass arguments in any form.

I changed the docstrings because `claim_search` also uses `--name`, and in that case it says normalized. I presume this means, it will be normalized, regardless of how the user enters it.

eukreign:
I think to do this properly we'd need to add a new column to store the normalized claim name:
1. for search/list purposes you'd use the normalized version to filter on
2. for returning results or display the name you would show the original name (preserving user supplied capitalization, etc)

PR #3392, opened 2021-08-10
suppress stack traces by default and add `--show-stack-trace` to show them

### LINKED ISSUE #3385: Output one-line error messages
SDK needs to output one-line error messages instead of full traces by default.

shyba:
closes #3385


belikor:
I like this patch. However, it seems there are various exceptions raised which are currently not handled by the code, so they all appear as Errors is the log. Various exceptions could be trapped so that they are only displayed as Warnings.

lyoshenka:
turns out jack fixed the underlying issue in #3399

PR #3364, opened 2021-07-14
file_manager: trap the exception if the claim is not a stream

belikor:
If `claim` is not a stream, the line `claim.stream.source.bt_infohash` will raise a `ValueError` exception because `claim.stream` does not contain a valid object.

This will happen if we try to download a claim that is in fact a `collection` (a playlist).
```
lbrynet get lbry://@something#e/My-New-List#f
```

Since at the moment we don't trap this exception, it prints the entire trace of the exception.
```
Traceback (most recent call last):
  File "/opt/git/lbry-sdk/lbry/file/file_manager.py", line 117, in download_from_uri
    if claim.stream.source.bt_infohash:
  File "/opt/git/lbry-sdk/lbry/schema/claim.py", line 55, in stream
    return Stream(self)
  File "/opt/git/lbry-sdk/lbry/schema/claim.py", line 108, in __init__
    self.message = self.claim.get_message(self.claim_type)
  File "/opt/git/lbry-sdk/lbry/schema/claim.py", line 46, in get_message
    raise ValueError(f'Claim is not a {type_name}.')
```

Therefore we trap this `ValueError` to avoid showing the entire trace.
We don't do anything with the exception, we just raise it again so that it is handled by the calling code `daemon.jsonrpc_get`.

Then it will show an appropriate message.
```
WARNING  lbry.extras.daemon.daemon:1106: Error downloading lbry://@something#e/My-New-List#f: Claim is not a stream.
```

PR #3396, opened 2021-08-16
schema/url: raise `ResolveError` if the URL is invalid

belikor:
This is similar to #3364.

When using `lbrynet get URL`, if the URL is not a valid URL the function `url.URL.parse` will raise a `ValueError` exception which will produce a whole backtrace.

For example, this is the case if we provide a channel name with a forward slash but without a stream name.
```
lbrynet get @Non-existing/
```

```
  File "/opt/git/lbry-sdk/lbry/file/file_manager.py", line 84, in download_from_uri
    if not URL.parse(uri).has_stream:
  File "/opt/git/lbry-sdk/lbry/schema/url.py", line 114, in parse
    raise ValueError('Invalid LBRY URL')
ValueError: Invalid LBRY URL
WARNING  lbry.extras.daemon.daemon:1106: Error downloading @Non-existing/: Invalid LBRY URL
```

Now we raise `ResolveError` which can be trapped in the upper functions that use `url.URL.parse` such as `FileManager.download_from_uri`. The traceback won't be shown.

```
WARNING  lbry.file.file_manager:249: Failed to download @Non-existing/: Failed to resolve 'Invalid LBRY URL'.
WARNING  lbry.extras.daemon.daemon:1106: Error downloading @Non-existing/: Failed to resolve 'Invalid LBRY URL'.
```

----

The `FileManager` class raises `ResolveError` in many situations but I don't know if this is the best exception to throw. Maybe a new exception should be defined like `NotDownloadableError` derived from `WalletError` in `lbry/error`.

### DIFF
--- a/lbry/file/file_manager.py
+++ b/lbry/file/file_manager.py
@@ -244,7 +244,7 @@ class FileManager:
             raise error
         except Exception as err:  # forgive data timeout, don't delete stream
             expected = (DownloadSDTimeoutError, DownloadDataTimeoutError, InsufficientFundsError,
-                        KeyFeeAboveMaxAllowedError)
+                        KeyFeeAboveMaxAllowedError, ResolveError)
             if isinstance(err, expected):
                 log.warning("Failed to download %s: %s", uri, str(err))
             elif isinstance(err, asyncio.CancelledError):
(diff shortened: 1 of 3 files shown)


eukreign:
`ResolveError` would be misleading to use for this case. Parsing a URL is a `ValueError`, meaning that the URL passed in is not an actual URL (an error of value), whereas `ResolveError` should be used when resolution fails (attempting to resolve a valid URL fails for some reason).

PR #3403, opened 2021-08-19
file_manager: raise new `InvalidURLError` if the URL is invalid

belikor:
This is a second attempt on #3396.

When using `lbrynet get URL`, if the URL is not a valid URL the function `url.URL.parse` will raise a `ValueError` exception which will produce a long backtrace.
```
lbrynet get @Non-existing/
```

Since we don't want to see the entire backtrace, we need a way of trapping the exception.

----

In #3364 it is said that
> ValueError is too generic of an exception to catch here, this will make it hard to find actual bugs later.

Therefore, we trap `ValueError`, and then raise a new exception, `InvalidURLError`, which is then trapped later in the code of `FileManager.download_from_uri`.


lbry/file/file_manager.py
```diff
@@ -81,8 +81,11 @@ async def download_from_uri(self, uri, exchange_rate_manager: 'ExchangeRateManag
         payment = None
         try:
             # resolve the claim
-            if not URL.parse(uri).has_stream:
-                raise ResolveError("cannot download a channel claim, specify a /path")
+            try:
+                if not URL.parse(uri).has_stream:
+                    raise ResolveError("cannot download a channel claim, specify a /path")
+            except ValueError:
+                raise InvalidURLError(uri)
```
eukreign, line 88: The other raised `ResolveError` should also probably be `InvalidURLError`.

eukreign (changes requested):

lbry/file/file_manager.py
belikor: Well, that was already there, so I didn't touch it.

I can change it to `InvalidURLError`, but if it was me, I would create another exception like `NotDownloadableError` as a channel is a valid URL but cannot be downloaded, so using a more precise exception would be better.

eukreign (changes requested):
`NotDownloadableChannel` makes it sound like this particular channel is not downloadable (but some channels are downloadable?)

I think instead of both `InvalidURL` and `NotDownloadableChannel`, it would be simpler and clearer to just have one `InvalidClaimURL` or `InvalidStreamURL` which can be used in-place of both of the new exceptions in this PR.

belikor:
New `InvalidStreamURLError` raised.

It should handle these cases where the URL only contains the channel part but not the actual claim part.
```
lbrynet get @Non-existing/
lbrynet get Non-existing/
lbrynet get @Non-existing
```

eukreign:
the code looks good now but this PR needs an integration test to show this new exception working, you can see example of a test which calls `get` and checks for error here:

https://github.com/lbryio/lbry-sdk/blob/d09bfdc4ffc7adf27f77bef1477904f779084167/tests/integration/datanetwork/test_file_commands.py#L226-L238

jackrobison:
merged here https://github.com/lbryio/lbry-sdk/commit/a841d49483bc5fe0895b176b4b2d105f7579f140

PR #3411, opened 2021-08-26
Daemon.jsonrpc_get: add `claim_id` and `download_repost` options

belikor:
The `get` method normally expects a canonical url or permanent url.
```
lbrynet get lbry://@channel/something
```

However, since this URL can have Unicode characters (Romance language accents, Cyrilic, Chinese ideograms, etc.) it may not be easy to type with all keyboard layouts.

A more reliable way to share "addresses" and download content is using the `'claim_id'` of the particular claim. This is a 40-character alphanumeric string so it only has the basic Latin alphabet and numbers.

It is easy to type with any keyboard layout.
```
lbrynet get 70dfefa510ca6eee7023a2a927e34d385b5a18bd --claim_id
```

----

Second commit.

Normally, if a particular claim is in fact a repost, it will not be downloaded, as it is not a stream.
```
lbrynet get reposted-content
```

Will result in
```
{
  "error": "Claim is not a stream."
}
```

To resolve the repost and download the actual claim we can use the option `--download_repost`.
```
lbrynet get reposted-content --download_repost
```

This can also be combined with the `--claim_id` option.
```
lbrynet get add2c47b68849b1ebdf153cae13b5cce7d60d710 --claim_id --download_repost
```

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -1125,14 +1125,14 @@ class Daemon(metaclass=JSONRPCServerType):
               FILE_MANAGER_COMPONENT)
     async def jsonrpc_get(
             self, uri, file_name=None, download_directory=None, timeout=None, save_file=None, wallet_id=None,
-            claim_id=False):
+            claim_id=False, download_repost=False):
         """
         Download stream from a LBRY name.
 
         Usage:
             get <uri> [<file_name> | --file_name=<file_name>]
              [<download_directory> | --download_directory=<download_directory>] [<timeout> | --timeout=<timeout>]
-             [--save_file=<save_file>] [--wallet_id=<wallet_id>] [--claim_id]
+             [--save_file=<save_file>] [--wallet_id=<wallet_id>] [--claim_id] [--download_repost]
 
 
         Options:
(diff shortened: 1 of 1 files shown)


eukreign:
@lyoshenka this is modifying the API, please review

lyoshenka:
- agreed we need a way to download by claim_id because some chars are not easy to input
- i think the syntax we want is `brynet get --claim_id=70dfefa510ca6eee7023a2a927e34d385b5a18bd`. @eukreign correct me if im wrong
- i don't think we need `download_repost`. if you call get on a repost, it should just download the underlying post. @eukreign do you agree, or is there complexity here i'm missing?

eukreign:
- downloading by claim_id is definitely desirable
- i've always wanted support for just the claim_id in the `lbry://` url scheme, so that would be my preference (`lbry://#claimid` without needing to put the claim name)
- downloading the content the repost points to sounds intuitive to me, a repost is like a file system symlink

belikor:
> * i think the syntax we want is `brynet get --claim_id=70dfefa510ca6eee7023a2a927e34d385b5a18bd`. @eukreign correct me if im wrong

I thought about this as well, however, at the moment, the first argument to `jsonrpc_get` is a mandatory `uri`.

https://github.com/lbryio/lbry-sdk/blob/d09bfdc4ffc7adf27f77bef1477904f779084167/lbry/extras/daemon/daemon.py#L1126-L1127

If we want to pass a value to `--claim_id`, then we need to make `uri` optional, and handle the case when both `--uri` and `--claim_id` are used simultaneously. Should `claim_id` take precedence?
```py
async def jsonrpc_get(
            self, uri=None, file_name=None, download_directory=None, timeout=None, save_file=None, wallet_id=None,
            claim_id=None):
```
```
lbrynet get --uri=something --claim_id=asdf2345
```

This is the reason I think the easiest solution is simply treating `uri` as a `claim id`, finding the real `uri`, and leaving most code intact.

The download is actually performed by `uri` (`FileManager.download_from_uri`).

----

> * i don't think we need `download_repost`. if you call get on a repost, it should just download the underlying post. @eukreign do you agree, or is there complexity here i'm missing?

In my own scripts, I did it precisely in the way you mention; I always check if a claim is a repost before downloading it. However, for the sdk, I thought about this `--download_repost` option to avoid breaking compatibility with older scripts that somebody may have written. In this case, the new functionality is only activated with an explicit `--download_repost`, and older scripts that don't use it are unaffected in their behavior (reposts won't be downloaded).

> downloading the content the repost points to sounds intuitive to me, a repost is like a file system symlink

I think to transparently handle the repost, this should be done in the `lbry.file.file_manager.FileManager` class.

https://github.com/lbryio/lbry-sdk/blob/d09bfdc4ffc7adf27f77bef1477904f779084167/lbry/file/file_manager.py#L58-L62

For example, with https://github.com/lbryio/lbry-sdk/pull/3364#issuecomment-912877059 we just raise an error if the claim is not a stream and thus cannot be downloaded, but we could also internally handle the case where the `uri` is in fact a repost.

I haven't tried, but I think it's technically possible to have a repost to a repost, so in that case we would need to recursively run some function to find the actual claim.

lyoshenka:
> If we want to pass a value to --claim_id, then we need to make uri optional, and handle the case when both --uri and --claim_id are used simultaneously. Should claim_id take precedence?

If both are passed, I'd raise an exception. In general it's better to be strict than to make guesses about what we think the user intended (unless the API is explicitly intended to make guesses)

> the easiest solution is simply treating uri as a claim id

Same problem with guessing here. `add2c47b68849b1ebdf153cae13b5cce7d60d710` may be a claimID, or it may be a valid uri. We could maybe say that all uris must start with `lbry://` and then if it doesnt its a claim ID, but that would be an even bigger change since that wasn't required before.

> I thought about this --download_repost option to avoid breaking compatibility with older scripts that somebody may have written. 

Thanks for thinking about BC here. I would be ok with this BC break if we explicitly call it out in the notes. This is a case where I think it is OK to guess what was intended because the request can only have one meaning. It's not possible to get a repost because a repost has no content. If you call `get` on a repost today, you get an error. I guess its possible that some apps rely on that, but most likely they would prefer if the get just got the reposted content.  So I'm ok with not introducing an extra flag here and downloading the reposted content.

PR #3413, opened 2021-08-28
Daemon.jsonrpc_getch: download the latest claims from a channel

belikor:
This pull request should be merged after #3411, as it uses the new option `--download_repost` from `jsonrpc_get`.

---

This pull request implements a new method `Daemon.jsonrpc_getch` to download the latest claims from a particular channel. It will only pick those claims that have a `'source'`, which means it will ignore livestreams.

Download the 10 newest claims, not including livestreams nor reposts.
```
lbrynet getch @MyChannel --number=10
```

If a particular claim is in fact a repost, it won't be downloaded.

To resolve the reposts, and download the original claims, we can use the option `--download_repost`.

Download the 20 newest items, including any reposts.
```
lbrynet getch @MyChannel --number=20 --download_repost
```

Other options are those of the `get` method, such as `--download_directory`, `--timeout`, `--save_file`, and `--wallet_id`.

### DIFF
--- a/lbry/extras/cli.py
+++ b/lbry/extras/cli.py
@@ -208,7 +208,7 @@ def get_argument_parser():
         group_parser = sub.add_parser(group_name, group_name=group_name, help=api['groups'][group_name])
         groups[group_name] = group_parser.add_subparsers(metavar='COMMAND')
 
-    nicer_order = ['stop', 'get', 'publish', 'resolve']
+    nicer_order = ['stop', 'get', 'getch', 'publish', 'resolve']
     for command_name in sorted(api['commands']):
         if command_name not in nicer_order:
             nicer_order.append(command_name)
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -1125,14 +1125,14 @@ class Daemon(metaclass=JSONRPCServerType):
               FILE_MANAGER_COMPONENT)
     async def jsonrpc_get(
             self, uri, file_name=None, download_directory=None, timeout=None, save_file=None, wallet_id=None,
-            claim_id=False):
+            claim_id=False, download_repost=False):
         """
         Download stream from a LBRY name.
 
         Usage:
             get <uri> [<file_name> | --file_name=<file_name>]
              [<download_directory> | --download_directory=<download_directory>] [<timeout> | --timeout=<timeout>]
-             [--save_file=<save_file>] [--wallet_id=<wallet_id>] [--claim_id]
+             [--save_file=<save_file>] [--wallet_id=<wallet_id>] [--claim_id] [--download_repost]
 
 
         Options:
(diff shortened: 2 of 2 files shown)


eukreign:
@lyoshenka @kauffj this PR is adding a new API end point, thoughts?

I think the plain `get` command accepting a channel URL would be a better way to go rather than adding a `getch` command (if we want to implement this kind of feature).

belikor:
> I think the plain get command accepting a channel URL

I like that `get` handles a single "thing", so a single claim. But allowing it to handle channel claims means that we have to decide what to do with multiple things.

Should `get` be able to download the claims inside a collection?
```
lbrynet get my-collection --download_collection
```

Get could resolve the collection, find the individual claims, and download each of these. Allowing `get` to act on channels could work in the same way; it would resolve the channel, find the newest claims, and download these.
```
lbrynet get @MyChannel --download_channel --number=20 --download_repost
```

I think I would be in favor of `get` accepting collections, but not on accepting channels. I feel that accepting channels would be a bit more involved, as a single channel can contain many things, streams, but also reposts, livestreams, and collections. Should downloading a channel that has a collection further resolve and download all items in that collection?
```
lbrynet get @MyChannel --download_channel --number=20 --download_repost --download_collection
```

This is why I think it's simpler to use a separate `getch` to deal with channel claims, which internally calls `get` on each of the found items.

> (if we want to implement this kind of feature).

This is a feature often requested: being able to download the latest content (or all content) from a single channel. The idea is that this is available in the SDK so that we can call it through some interface from the LBRY Desktop application and any other application that uses the SDK.

lyoshenka:
I haven't thought through all the details yet, but here are some quick thoughts:

- thanks for creating this. i can see why people would want an easy way to download content from a channel
- no matter what we decide about new api vs using `get`, the name `getch` is not great. its unclear and doesnt match our existing naming style.
- not every use SDK use case needs to be a single RPC call. the SDK is mostly meant for other apps to call, not for direct cli use. i'd prefer a design with fewer calls and flags over a swiss-army-knife approach. i think this use case can be accomplished by doing a claim search and then getting each result, and that would keep our api simpler

belikor:
> * thanks for creating this. i can see why people would want an easy way to download content from a channel

I think this is one of the most requested features that I've seen in the comments section of Odysee, on Discord, etc. Power users, that understand the peer to peer nature of LBRY, want to be able to download and seed the channels they like; they want to set up seedboxes. This is the reason projects like [lbry-seedit](https://github.com/tuxfoo/lbry-seedit) and [lbrytools](https://github.com/belikor/lbrytools) exist.

> the SDK is mostly meant for other apps to call, not for direct cli use.

I don't disagree, but downloading content from channels from the command line doesn't introduce at lot of complexity, it's as simple as using the regular `get`.
```
lbrynet getch @MyFavoriteChannel
```

> i think this use case can be accomplished by doing a claim search and then getting each result, 

Yes, but this would have to be redone in every application currently existing and that may be developed in the future (LBRY Desktop, Android, iOS, terminal interface, flutter interface, HTML5 interface, etc.). Integrating this functionality into the SDK itself is better to avoid duplicating that work.

jring-o:
The ability to proactively choose what content and creators to support by reseeding their content is the #1 requested feature that I see from users. The potential, from engagement to metrics to business-models, is insane.

Any progress?

lyoshenka:
Sorry it's taking me a while on this. It's hard to evaluate a solution from a PR. Much easier when there's a description of the problem and some discussion about the best way to solve it (if this happened in Discord and I missed it, plz post a link). I also don't have a clear picture of how this works from an app user's perspective.

Here's how I see it: the same way that podcast apps let people subscribe to a show and auto-download every episode (or up to 5 episodes, or the 3 most recent episodes, etc), people want to subscribe to LBRY channels to auto-download and seed content. In the app, this could be a selector somewhere near the "Following" button that lets you enable auto-downloading and choose how you want it to work.

If this is a good description of what people want, then we can move on to solutions. A one-time call to download some number of videos might not be the best choice. Perhaps the call we really need is `channel_subscribe`, which tells the SDK how to subscribe to and auto-download from a channel. Or maybe it's something else.

@belikor: I appreciate all the PRs you've been opening and I'm sorry we've been slow to merge them. When we see your work, we're also thinking about how it fits in with the overall design of LBRY, and knowing what's going on from the user's perspective (especially the app user, not just the power user) is very important. Please include that in your PR descriptions, or @ me in Discord next time you're planning a feature. Alternatively, create an issue first that describes the problem, and then make a PR that fixes the issue.

lyoshenka:
@jring-o does that :point_up: describe the problem well? What did I leave out?

kauffj:
@belikor if you want to work on this, please create an API for subscribing to a channel, which could then have the settings this call adds.

PR #3414, opened 2021-08-31
removed `comment` API endoints

cristi-zz:
Removed the comment API
Removed tests for the comment API
Removed the documentation section
Removed the comment server configuration


shyba (approved):

kodxana:
Will be there any replacement for this?

lyoshenka:
we will keep using Odysee comment system in the app. no login required - comments are signed using channel keys. this is what the sdk was already doing underneath, so removing these APIs is just about clarifying that the comments are not part of the lbry protocol.

we would like to do federated comments as part of the protocol down the road

PR #3415, opened 2021-09-04
Daemon.jsonrpc_get: add `download_collection` option for handling collections

belikor:
This can be combined with #3411, but in this pull request we only show the code to handle collections.

Normally, if a particular claim is in fact a collection (playlist), it will not be downloaded, as it is not a stream.
```
lbrynet get collection-claim
```

Will result in
```
{
  "error": "Claim is not a stream."
}
```

To resolve the collection, and download the claims that it contains we can use a new option `--download_collection`.
```
lbrynet get collection-claim --download_collection
```

A collection can have an arbitrary number of items, so to limit the number of claims to download we can use `--max_claims`.
```
lbrynet get collection-claim --download_collection --max_claims=4
```

To reverse the order of the claims in the collection use `--reverse_collection`. This can be used to get the newest items in the collection.
```
lbrynet get collection-claim --download_collection --max_claims=10 --reverse_collection
```

----

The inner code to download a URL is put in its own method `_get`, then the collection is parsed, and `_get` is used on each of the internal claims.

----

After merging #3411, we can refactor this pull request to be able to use a claim ID.
```
lbrynet get collection-music
lbrynet get collection-music --download_collection
lbrynet get collection-music --download_collection --max_claims=1 --reverse_collection

lbrynet get 87560bfa01c2fb92d39394e4d508cf7a453eef77 --claim_id
lbrynet get 87560bfa01c2fb92d39394e4d508cf7a453eef77 --claim_id --download_collection
lbrynet get 87560bfa01c2fb92d39394e4d508cf7a453eef77 --claim_id --download_collection --max_claims=1 --reverse_collection
```


eukreign:
@lyoshenka please review the API changes

lyoshenka (changes requested):
There's no need for a `--download_collection` flag. If you call `get` on a collection, then you want to download the collection.

I'd rename `--max_claims` to `--collection_limit` so it's clear that flag applies to collections.

Otherwise the API looks good to me. @eukreign please review the code itself once @belikor makes these changes.

kauffj:
@belikor planning to come back to this?

PR #3417, opened 2021-09-08
use database to track blob disk space use and preserve own blobs

eukreign:


shyba (approved):
Looks great, except for those tests failing. I can do some exploratory testing and review again when they're green.

PR #3430, opened 2021-09-17
Daemon.jsonrpc_file_delete: new `--what` parameter to control what to delete

belikor:
The `--what` parameter can be used to specify what to delete: the media file only, the blobs only, or both, deleting completely the downloaded stream.
```
lbrynet file delete --claim_name=my-claim --what=file
lbrynet file delete --claim_name=my-claim --what=blobs
lbrynet file delete --claim_name=my-claim --what=both
```

If no argument is used, this is equivalent to `--what=blobs`, which is the original behavior.
```
lbrynet file delete --claim_name=my-claim
```

The old argument, `--delete_from_download_dir`, works as `--what=both`.
```
lbrynet file delete --claim_name=my-claim --delete_from_download_dir
lbrynet file delete --claim_name=my-claim --what=both
```

Personally, I think we could deprecate completely this `--delete_from_download_dir` parameter.

----

In general, we extend the `delete` method of `FileManager` with the parameter `delete_blobs` to make the deletion of the blobs optional. This requires adding the argument `delete_source` to `StreamManager.delete`.

However, the `pylint` configuration complains and tells us that there is a mismatch in the arguments of the parent and children classes, so we also need to add the corresponding `delete_source` argument to `SourceManager.delete` and `TorrentManager.delete`.

### DIFF
--- a/lbry/extras/daemon/daemon.py
+++ b/lbry/extras/daemon/daemon.py
@@ -2120,24 +2120,24 @@ class Daemon(metaclass=JSONRPCServerType):
         return msg
 
     @requires(FILE_MANAGER_COMPONENT)
-    async def jsonrpc_file_delete(self, delete_from_download_dir=False, delete_all=False, **kwargs):
+    async def jsonrpc_file_delete(self, delete_from_download_dir=False, delete_all=False, what=None, **kwargs):
         """
-        Delete a LBRY file
+        Delete the blobs and file from a downloaded LBRY claim
 
         Usage:
             file_delete [--delete_from_download_dir] [--delete_all] [--sd_hash=<sd_hash>] [--file_name=<file_name>]
                         [--stream_hash=<stream_hash>] [--rowid=<rowid>] [--claim_id=<claim_id>] [--txid=<txid>]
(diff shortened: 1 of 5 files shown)


eukreign:

belikor:
> does this add anything we don't have today, or is it just a rename of the flags that already exist?

With this new option, we make deleting blobs optional (True by default).

Currently `file_delete` always deletes the blobs. There are instances where the user would like to delete the media files (mp4, mkv, mp3, etc.) to free space on disk but keep the blobs to continue seeding the content. This is desired for a seedbox, for example.

```
# Current and new; delete only blobs
lbrynet file delete --claim_name=<name>
lbrynet file delete --claim_name=<name> --what=blobs  # Also without this option

# Current and new; delete blobs and file
lbrynet file delete --claim_name=<name> --delete_from_download_dir
lbrynet file delete --claim_name=<name> --what=both

# New; delete only the file
lbrynet file delete --claim_name=<name> --what=file
```

lyoshenka:
The SDK should not be responsible for deleting files that users download. They can be deleted directly using normal commands such as `rm`.

I'm going to close this. We can continue to discuss if you think I missed something.

PR #3459, opened 2021-10-22
Improve P2P issues around banning peers and duplicated DHT responses

shyba:


jackrobison:
can this be reproduced in an integration test? I don't know a scenario for this to happen other than downloading from a peer who doesn't yet have the fuil stream, where you download from them faster than they download from anyone else.

shyba:
This happens frequently when downloading from reflector. When that happens and you only have one peer, the download gets stuck and it doesn't retry.

The reason for failures isn't only not having the blob. You can have a timeout on a first moment, then retry and succeed, which is the case I am seeing.

I will try to make it into an integration test.

shyba:
@jackrobison 
`test_incomplete_downloads_retry` was broken. It wasn't deleting the blobs for testing. I fixed and added a check so it ensures there are things on the list to delete.

Then, if you check the last commit you will see that this fix makes the test pass without restarting the stream.

Note that, as explained in the last comment, downloading from incomplete stream is one of the cases it covers but the main one I want to fix is about normal failures where a retry is enough. As retrying an incomplete stream covers more than that, this test covers both cases (what will cause it to break is irrelevant as long as it recovers automatically).

jackrobison (approved):

PR #3484, opened 2021-11-15
Use rocksdb instead of leveldb

jackrobison:
This switches the key/value database used by hubs from `plyvel` (leveldb) to `lbry-rocksdb`, which is an up-to-date binding for rocksdb.

### DIFF
--- a/docker/Dockerfile.wallet_server
+++ b/docker/Dockerfile.wallet_server
@@ -15,7 +15,6 @@ RUN apt-get update && \
       build-essential \
       automake libtool \
       pkg-config \
-      libleveldb-dev \
       python3.7 \
       python3-dev \
       python3-pip \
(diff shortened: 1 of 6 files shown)


lyoshenka:
@jackrobison who should review this? plz assign reviewers

eukreign:
@jackrobison let me know when this is ready for review

ofek:
@jackrobison Hello there, great job on https://github.com/lbryio/lbry-rocksdb! We're thinking of using that so I'm testing it now. I have a question: Can secondary instances read statistics of the primary? For example, getting the `rocksdb.stats` property https://github.com/facebook/rocksdb/blob/v6.27.3/include/rocksdb/db.h#L770-L772

jackrobison:
@ofek Secondary dbs can read many of the properties, but I'm not sure about `rocksdb.stats`. Here's a test that checks some of the other properties: https://github.com/lbryio/lbry-rocksdb/blob/master/tests/test_db.py#L720

ofek:
I found out it was impossible https://github.com/facebook/rocksdb/issues/908#issuecomment-1007048041

PR #3486, opened 2021-11-17
Fix claims without source appearing on `stream_types` searches

### LINKED ISSUE #3472: claim_search: finds livestream claims when `stream_types=['binary']`
This will return livestreams claims and only a few actual binary files (zip files for example).
```py
msg = {'method': 'claim_search',
       'params': {'page': 1,
                  'page_size': 50,
                  'no_totals': True,
                  'order_by': 'trending_mixed',
                  'claim_type': "stream",
                  'stream_types': ['binary']}}
items = requests.post("http://localhost:5279", json=msg).json()["result"]["items"]

for i in items:
    print(i["value_type"], i["value"].get("stream_type", None), i["name"])
```

----

In the next example, we specifically search for stream claims with no source, that is, these should be only livestreams.

```py
msg2 = {'method': 'claim_search',
        'params': {'page': 1,
                   'page_size': 50,
                   'no_totals': True,
                   'order_by': 'trending_mixed',
                   'claim_type': "stream",
                   'has_no_source': True}}
items2 = requests.post("http://localhost:5279", json=msg2).json()["result"]["items"]

for i in items2:
    print(i["value_type"], i["value"].get("stream_type", None), i["name"])
```

----

So it seems to me that in the first case, it should return only the binary files, and omit livestream claims that don't have a source.

I found a reference in `lbry.schema.mime_types` that it returns a `"binary"` if it cannot guess the stream type. So maybe this is the reason the livestreams are reported as `"binary"`. Probably it can be a different default value.

https://github.com/lbryio/lbry-sdk/blob/ddbbb6f1dd9e5117eb503841ef43cb0f2b2fb37a/lbry/schema/mime_types.py#L180-L184

shyba:
closes #3472

### DIFF
--- a/lbry/wallet/server/leveldb.py
+++ b/lbry/wallet/server/leveldb.py
@@ -565,7 +565,7 @@ class LevelDB:
             reposted_has_source = False if not reposted_metadata.is_stream else reposted_metadata.stream.has_source
             reposted_claim_type = CLAIM_TYPES[reposted_metadata.claim_type]
             reposted_stream_type = STREAM_TYPES[guess_stream_type(reposted_metadata.stream.source.media_type)] \
-                if reposted_metadata.is_stream else 0
+                if reposted_has_source else 0
             reposted_media_type = reposted_metadata.stream.source.media_type if reposted_metadata.is_stream else 0
             if not reposted_metadata.is_stream or not reposted_metadata.stream.has_fee:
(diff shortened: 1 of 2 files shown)


lyoshenka:
@jackrobison please review

jackrobison (approved):

Issue #3482, opened 2021-11-15
Add timeouts to test setup and teardown


kauffj:

FemtosecondLaser:
Which test setup/teardown the issue is about?
What is the problem that requires a timeout mechanism?

FemtosecondLaser:
@eukreign if this is the problem of the 6h pipeline runs - please assign me. I am working on it.

eukreign:
@FemtosecondLaser yes, this is related to tests timing out on GitHub CI

PR #3490, opened 2021-11-20
added timeout of async operations to integration test setup/teardown

FemtosecondLaser:
pull request addressing the issue #3482
Had a look at the causes of the 6h pipeline runs (such as [this](https://github.com/lbryio/lbry-sdk/runs/4194314255?check_suite_focus=true)) - appeared to be async operations in the test cleanup procedure. This change adds timeout to the setup and cleanup of the integration tests.

Is it known why the operations were getting stuck? If not - could more logging be added to look into this problem further?

### DIFF
--- a/lbry/testcase.py
+++ b/lbry/testcase.py
@@ -132,17 +132,18 @@ class AsyncioTestCase(unittest.TestCase):
 
             with outcome.testPartExecutor(self):
                 self.setUp()
+                self.addTimeout()
                 self.loop.run_until_complete(self.asyncSetUp())
             if outcome.success:
                 outcome.expecting_failure = expecting_failure
                 with outcome.testPartExecutor(self, isTest=True):
                     maybe_coroutine = testMethod()
                     if asyncio.iscoroutine(maybe_coroutine):
-                        if self.TIMEOUT:
-                            self.loop.call_later(self.TIMEOUT, self.cancel)
+                        self.addTimeout()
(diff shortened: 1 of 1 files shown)


eukreign:
@FemtosecondLaser thanks for this PR!

the tests don't seem to be passing... also, where does `TIMEOUT` get set?

i think we always want all tests to have a timeout, there should be some default timeout set (how about 30 seconds)?

FemtosecondLaser:
> `TIMEOUT` should be set to a default value somewhere and we always want the timeout enabled, unless you can think of a reason a test might not want to have a timeout?

TIMEOUT is set to 120 sec [here](https://github.com/FemtosecondLaser/lbry-sdk/blob/d69486fb6e1a95538515350b6d4fd80198e84c9c/lbry/testcase.py#L87). It has been there for timing out async operations in the main body of the tests. I extended the usage for startup and cleanup parts of the tests.
Timeout gets disabled for some tests [here](https://github.com/FemtosecondLaser/lbry-sdk/blob/d69486fb6e1a95538515350b6d4fd80198e84c9c/tests/unit/dht/test_node.py#L15). The removal of conditional set up of timeout makes those tests unhappy.

FemtosecondLaser:
@eukreign 
I suspect the tests are failing for reasons unrelated to this change.
The latest run failed while building for mac. All the tests passed.
I don't have an option to re-run the pipe, could you do that please.

Issue #3463, opened 2021-10-25
Understand and/or design system on how to measure disk space available


kauffj:
And/or content availability overall

shyba:
Some articles I found related to estimating the network size:
https://www.irjet.net/archives/V5/i12/IRJET-V5I1208.pdf
https://eli.sohl.com/2020/06/05/dht-size-estimation.html
https://www.cl.cam.ac.uk/~lw525/MLDHT/

First phase will be adding more metrics options to the DHT seed node, so an operator can use it to see data about the network partition it belongs. Then this data can be used to estimate the other partitions.

Some idea: fetch from blockchain data (hubs?) how many SD blob hashes there are for a partition and compare to how many we have seen. The ratio of existing/announced should tell how much of it is hosted initially.

PR #3500, opened 2021-12-03
Add Prometheus metrics for DHT internals

shyba:
The SDK already supports exporting metrics over Prometheus if the user wishes to do so. This PR adds the following metrics under the `dht_node` namespace:

#### data store
- `stored_blobs_x_bytes_colliding`: Number of blobs with at least X bytes colliding with this node id prefix
    - This is used to measure the ratio of content on network vs on blockchain. Check #3521 and [the dashboard repo](https://github.com/lbryio/dashboard), but it also needs #3511 merged.
- `stored_blobs`: Number of blobs announced by other peers
- `storing_peers`: Number of peers storing blobs announced to this node
#### protocol
- `peer_manager_keys`: Number of keys tracked by PeerManager dicts (sum)
    - If that grows too large, we can get memory leaks
- `request_sent`: Number of requests send from DHT RPC protocol
- `request_success`: Number of successful requests
- `request_error`: Number of errors returned from request to other peers
- `response_time`: Response times of DHT RPC requests
- `received_request`: Number of received DHT RPC requests
#### routing table
- `peers_in_routing_table`: Number of peers on routing table
- `peer_x_bit_colliding`: Number of peers with at least X bits colliding with this node id
    - This is used for estimating network size. Details on #3463, but the raw calculations will live on [the dashboard repo](https://github.com/lbryio/dashboard). This also tells how close we are from the peers we added on the routing table, which can help dissecating a live node health into how well we are able to find and and add neighbors.
- `buckets_in_routing_table`: Number of buckets on routing table
    - Debugging purposes.

### New endpoints:

When starting the seed node script with `--metrics_port=<some_port>`, a web server will expose the following endpoints:

- `/metrics`: Prometheus endpoint with the above metrics.

If desired, routing table peers and announced blobs can be exported in CSV format using `--enable_csv_export`. However, you should use a firewall and avoid exposing that publicly as the announced blobs list can reach multiple thousands items, easily bringing your server down by malicious users.
- `/peers.csv`: All peers from the routing table in CSV format (ip, port, node_id)
- `/blobs.csv`: All blobs announced to the node in CSV format (hex encoded blob hash)

### DIFF
--- a/lbry/dht/node.py
+++ b/lbry/dht/node.py
@@ -24,6 +24,10 @@ class Node:
         "storing_peers", "Number of peers storing blobs announced to this node", namespace="dht_node",
         labelnames=("scope",),
     )
+    stored_blob_with_x_bytes_colliding = Gauge(
+        "stored_blobs_x_bytes_colliding", "Number of blobs with at least X bytes colliding with this node id prefix",
+        namespace="dht_node", labelnames=("amount",)
+    )
     def __init__(self, loop: asyncio.AbstractEventLoop, peer_manager: 'PeerManager', node_id: bytes, udp_port: int,
                  internal_udp_port: int, peer_port: int, external_ip: str, rpc_timeout: float = constants.RPC_TIMEOUT,
                  split_buckets_under_index: int = constants.SPLIT_BUCKETS_UNDER_INDEX,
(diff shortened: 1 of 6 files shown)


lyoshenka:
- drop `*_flight` counter
- rebase
- ready to merge

PR #3503, opened 2021-12-03
WIP: iterative find cleanup

shyba:
This makes the iterative find algorithm behave mostly like [described in the wiki](https://github.com/lbryio/lbry-sdk/wiki/DHT-Resources#naive-iterativefind-description), which looks in sync with how I found it described everywhere else so far.

Missing and some notes:
- add the stop condition in a way that is not too aggressive (tried as described, but it stops immediately)
  - the stop condition described in the wiki happens when the last element, sorted by distance, in list of nodes we have to return is closer to the key we are searching than the nodes we have available to query. The current issues trying to implement that are:
      - it is concurrent, so this can happen intermittently at many points in time, which makes it only feasible to do it on the `_search_round` which kind of has the whole context
      - stops too early when done there... which is weird (stops improving for real?)
      - when that happens, sometimes lowering alpha helps and increasing worsens it
- some kind of metric to see if we are improving or not, which I suppose could fit mainly 3 categories:
  - quality (average distance of returned nodes to key)
  - speed (time to find first value?)
  - traffic (crashing routers is no fun)
- feedback (later stage)

### DIFF
--- a/lbry/dht/protocol/iterative_find.py
+++ b/lbry/dht/protocol/iterative_find.py
@@ -85,22 +85,19 @@ class IterativeFinder:
         self.protocol = protocol
 
         self.key = key
-        self.bottom_out_limit = bottom_out_limit
-        self.max_results = max_results
+        self.max_results = max(constants.K, max_results)
         self.exclude = exclude or []
 
         self.active: typing.Set['KademliaPeer'] = set()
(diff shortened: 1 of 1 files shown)


lyoshenka:
- after #3500 is merged, run master and this branch and compare which does better
- if its better, merge it

PR #3534, opened 2022-01-01
drop dependency on cryptography library in wallet module

### LINKED ISSUE #3532: investigate: ecdsa created signatures do not validate with coincurve
Investigate why old signatures created using `ecdsa` do not validate with `coincurve`:

Old signing code (`private_key` in this case is instance of `ecdsa.SigningKey`):
https://github.com/lbryio/lbry-sdk/blob/26d0e87f463cdede9a31468ddcf397351a1aa397/lbry/wallet/transaction.py#L459

New signing code (calls the `libsecp256k1` code wrapped by `coincurve`):
https://github.com/lbryio/lbry-sdk/blob/8076000c2761d7ca077049f1e9e1de177553d6f0/lbry/wallet/bip32.py#L238-L258

New validation code includes the old validation as fallback for old signatures:
https://github.com/lbryio/lbry-sdk/blob/8076000c2761d7ca077049f1e9e1de177553d6f0/lbry/wallet/transaction.py#L428-L440

The following tests illustrate the problem, if you step through debugger you will see that it does not validate with `coincurve` and instead falls back to using the old validation with `cryptography`:

Example 1 (older):
https://github.com/lbryio/lbry-sdk/blob/8076000c2761d7ca077049f1e9e1de177553d6f0/tests/unit/wallet/test_schema_signing.py#L74

Example 2 (much more recent):
https://github.com/lbryio/lbry-sdk/blob/8076000c2761d7ca077049f1e9e1de177553d6f0/tests/unit/wallet/test_schema_signing.py#L123

Why can't `coincurve` validate those signatures? Is there anything we can do to get it to validate?

One interesting thing to note is that the new signatures do validate with the old validation scheme using `cryptography` library.

eukreign:
fixes #3532

### DIFF
--- a/lbry/wallet/bip32.py
+++ b/lbry/wallet/bip32.py
@@ -93,6 +93,10 @@ class PublicKey(_KeyBase):
             self.verifying_key = self._verifying_key_from_pubkey(pubkey)
 
     @classmethod
+    def from_compressed(cls, public_key_bytes, ledger=None) -> 'PublicKey':
+        return cls(ledger, public_key_bytes, bytes((0,)*32), 0, 0)
+
+    @classmethod
     def _verifying_key_from_pubkey(cls, pubkey):
         """ Converts a 33-byte compressed pubkey into an coincurve.PublicKey object. """
         if not isinstance(pubkey, (bytes, bytearray)):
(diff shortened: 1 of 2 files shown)


lbry/wallet/bip32.py
```diff
@@ -137,9 +141,32 @@ def extended_key(self):
             self.pubkey_bytes
         )
 
-    def verify(self, signature, data):
-        """ Produce a signature for piece of data by double hashing it and signing the hash. """
-        return self.verifying_key.verify(signature, data, hasher=double_sha256)
+    def verify(self, signature, data) -> bool:
+        """ Verify that a signature is valid for data. """
+
+        if len(signature) != 64:
```
shyba, line 147: can you add `data` too? Just to make sure it isn't None and is a valid input before it hits FFI

shyba (changes requested):

lbry/wallet/bip32.py
eukreign: it currently just throws a `TypeError` if you pass `None`, what should it do instead?

shyba: It was a suggestion as I noticed you added this check for `signature`, which I agree and liked a lot because it protects against mistakes when it reaches the FFI calls.

Usually I would not mind, but since this piece interfaces with C and cryptography, I went to read the C and coincurve parts. Here are some ideas I had to try to clarify this piece:
* Rename ([it looks like this is not a raw free form `data` field, it is a message digest according to the C definition](https://github.com/bitcoin-core/secp256k1/blob/0559fc6e41b65af6e52c32eb9b1286494412a162/src/secp256k1.c#L400))
  * The argument is [named `digest` across callers](https://github.com/lbryio/lbry-sdk/pull/3534/files/42cf4b44a218177781002f9f8b81f1a7ef2e8fb9#diff-6647274cf37bffc93ba6502f9527d836b9165aa8d83f585d7d31d66d669a54c3R424), which sounds like a better name
* The docstring feels incomplete, it verifies the signature for a digest of exactly 32 bytes.
* Check that the length is valid, as done with `signature`, for consistency
* Consider using `ValueError` instead. [At least for consistency with coincurve if that gets upstreamed](https://github.com/ofek/coincurve/blob/23c9f2726484c2cf42fb1890902cc075d06b8e61/coincurve/keys.py#L405)

Those are suggestions for trying to clarify this parameter expectations, but you are free to ignore and merge as no caller is currently misusing it. I just think it helps clarifying and avoiding later misuse or digging into this code to recover that knowledge.

shyba (approved):

PR #3542, opened 2022-01-13
Gh 3481 file type detection

eug3nix:
Hi.
This is my attempt at addressing the issue mentioned in bug https://github.com/lbryio/lbry-sdk/issues/3481

I have used a pure-python library for file type detection instead of libmagic binding in order to not introduce extra complexity to the build process and possible  issues with other platforms (although libmagic is far superior in its robustness and the amount of supported formats).

Here are some notes on the implementation:
1 - The **guess_media_type** function was changed so that it tries to open the file from the provided path
2 - If it is able to open the file and detect its type it checks if the extension parsed from the path (**extA**) is not matching the detected extension (**extB**)
3 - Then it checks if the **extA** is a synonym of **extB** by looking at the synonyms_map, if **extB** is NOT a synonym, then **extA** is replaced.

The rest of the code stays the same.
**Note0:** Extra logic described in step 3 is to handle edge cases when the internals format of the file has an extension which differs from the current file extension but still the current extension is legit. This allows the new code to keep formats such as .cbz (internally a .zip) or cbr (internally a .rar) to be overridden.

**Note1:** This function may potentially fail with relative file paths in case if the CWD of lbrynet process is not the same as the user's working dir (this is my guess, I have not tested this tbh.)

Please let me know what you think, does this solution is potentially worth of merging, should I add tests or do some other improvements?

Cheers,
Eugene


kauffj:
Thanks @eug3nix, review coming shortly!

lbry/schema/mime_types.py
```diff
@@ -166,10 +168,41 @@
...
+
+            # override extension parsed from file...
+            if extension != realext:
+                if extension:
+                    log.warning(f"file extension does not match it's contents {path}, identified as {realext}")
+                else:
+                    log.debug(f"file {path} does not have extension, identified by contents as {realext}")
+
+                # don't do anything if extension is in synonyms
+                if not extension in synonyms_map.get(realext, []):
```
eukreign, line 200: I think this would be clearer to move the `not` to be before the `in`. Like this:
```python
if extension not in synonyms_map.get(realext, []):
```

lbry/schema/mime_types.py
```diff
@@ -166,10 +168,41 @@
...
 
 def guess_media_type(path):
     _, ext = os.path.splitext(path)
     extension = ext.strip().lower()
+
+    # try detecting real file format if path points to a readable file
+    try:
+        kind = filetype.guess(path)
+        if kind:
+            realext = f".{kind.extension}"
```
eukreign, line 190: In Python the standard variable naming convention is to use full words and to put underscores between words.

lbry/schema/mime_types.py
```diff
@@ -166,10 +168,41 @@
...
+    '.rar': ['.cbr'],
+    '.ar': ['.a']
+}
+
+log = logging.getLogger(__name__)
+
 
 def guess_media_type(path):
     _, ext = os.path.splitext(path)
     extension = ext.strip().lower()
```
eukreign, line 184: Since the `extension` is needed without the dot in the old code and in your code, it would make sense to slice it here to remove the dot. This will simplify your code (no need for `f".{kind.extension}"`) and the old code as well.

eukreign (changes requested):
Thanks for this PR.

Two overall comments:
1. this definitely will need tests
2. avoid redundant comments. use comments judiciously, only in cases where the code cannot be written clearer without a noticeable performance penalty. comments are not a replacement for writing good, legible, code. use clearer variable names if necessary.

I've also added some inline feedback.

lbry/schema/mime_types.py
eug3nix: Adjuster per your feedback.

lbry/schema/mime_types.py
eug3nix: Adjusted.

lbry/schema/mime_types.py
eug3nix: the dotted extension is being used in the types_map, thats why I prepending the dot to the detected extension when setting the real_extension var, am I getting something wrong here ?

eug3nix:
I've also added a number of tests to cover all cases I could imagine, please take a look @eukreign 
Thanks!

lbry/schema/mime_types.py
eukreign: does `synonyms_map` need to have the dots?

eukreign (approved):

PR #3559, opened 2022-02-04
DHT: iterative find bugfixes and better logging

shyba:
backwards-incompatible: removed unused `search_bottom_out_limit` from `peer list` API (it was ignored before and got removed entirely from internals)

- closest peer is only considered when already contacted and not known to be bad
- don't add bad peers to active
- reset closest and previous closest if they go bad during search
- fix sorting on search round
- improved and unified some of value/node find debug logging into a common one


lbry/dht/protocol/iterative_find.py
```diff
@@ -138,15 +132,14 @@ def get_initial_result(self) -> typing.List['KademliaPeer']:  #pylint: disable=n
...
+            return
+        if peer in self.contacted:
+            return
         if peer not in self.active and peer.node_id and peer.node_id != self.protocol.node_id:
-            self.active.add(peer)
-            if self._is_closer(peer):
-                self.prev_closest_peer = self.closest_peer
-                self.closest_peer = peer
+            self.active[peer] = self.distance(peer.node_id)
+            self.active = OrderedDict(sorted(self.active.items(), key=lambda item: item[1]))
```
jackrobison, line 142: instead of remaking the ordered dict you could use an always-sorted list that you `insort` into

PR #3562, opened 2022-02-08
DHT: improve iterative find

shyba:
- uses a dict instead
- bottoming out on peer search gives a warning and returns empty (returning earlier could give bad results for announcements, so imo it is preferable to fail entirely)
- probes a lot less (stops probing if it goes outside of the top `max_result` closest nodes) = less packets, same precision, more concurrency expected

depends: #3559

### DIFF
--- a/lbry/dht/protocol/iterative_find.py
+++ b/lbry/dht/protocol/iterative_find.py
@@ -360,8 +360,8 @@ class IterativeNodeFinder(IterativeFinder):
         if self.are_k_closest_peers_ready:
             self.put_result(self.active.keys(), True)
         elif self.bottom_out_count >= self.bottom_out_limit or self.iteration_count >= self.bottom_out_limit:
-            log.debug("peer search bottomed out.")
-            self.put_result(self.active.keys(), True)
+            log.warning("peer search bottomed out.")
+            self.put_result([], True)
 
 
 class IterativeValueFinder(IterativeFinder):
--- a/lbry/dht/protocol/iterative_find.py
+++ b/lbry/dht/protocol/iterative_find.py
@@ -89,7 +89,7 @@ class IterativeFinder:
         self.max_results = max_results
         self.exclude = exclude or []
 
-        self.active: typing.Set['KademliaPeer'] = set()
+        self.active: typing.Dict['KademliaPeer', int] = {}  # peer: distance, sorted
         self.contacted: typing.Set['KademliaPeer'] = set()
         self.distance = Distance(key)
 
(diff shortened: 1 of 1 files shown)


lbry/dht/protocol/iterative_find.py
```diff
@@ -85,46 +85,27 @@ def __init__(self, loop: asyncio.AbstractEventLoop, peer_manager: 'PeerManager',
         self.protocol = protocol
 
         self.key = key
-        self.bottom_out_limit = bottom_out_limit
-        self.max_results = max_results
+        self.max_results = max(constants.K, max_results)
         self.exclude = exclude or []
 
-        self.active: typing.Set['KademliaPeer'] = set()
+        self.active: typing.Dict['KademliaPeer', int] = {}  # peer: distance, sorted
```
jackrobison, line 91: should be an ordered dict to be explicit

lbry/dht/protocol/iterative_find.py
```diff
@@ -242,35 +211,31 @@ def _schedule_probe(self, peer: 'KademliaPeer'):
...
         try:
             if self.running:
-                await self._search_round()
-            if self.running:
-                self.delayed_calls.append(self.loop.call_later(delay, self._search))
+                if self.delayed_call:
+                    self.delayed_call.cancel()  # ensure anything scheduled gets cancelled
+                self._search_round()
+            #if self.running:
+            #    self.delayed_call = self.loop.call_later(delay, self._search)
```
jackrobison, line 228: can `delayed_call` be deleted?

lbry/dht/protocol/iterative_find.py
```diff
@@ -242,35 +211,31 @@ def _schedule_probe(self, peer: 'KademliaPeer'):
...
 
         def callback(_):
-            self.running_probes.difference_update({
-                probe for probe in self.running_probes if probe.done() or probe == t
-            })
-            if not self.running_probes:
-                self.tasks.append(self.loop.create_task(self._search_task(0.0)))
+            for peer in [peer for peer, task in self.running_probes.items() if task.done() or task == t]:
+                del self.running_probes[peer]
+            self._search_task(0.0)
```
jackrobison, line 216: the delay isn't needed as an arg anymore

lbry/dht/protocol/iterative_find.py
```diff
@@ -242,35 +211,31 @@ def _schedule_probe(self, peer: 'KademliaPeer'):
...
-        if self.closest_peer and self.prev_closest_peer:
-            log.debug("[%s] best node id: %s (contacted: %s, good: %s), previous best: %s",
-                      self.key.hex()[:8], self.closest_peer.node_id.hex()[:8], self.closest_peer in self.contacted,
-                      self.peer_manager.peer_is_good(self.closest_peer), self.prev_closest_peer.node_id.hex()[:8])
+        log.debug("[%s] check result: %i active nodes %i contacted",
+                  self.key.hex()[:8], len(self.active), len(self.contacted))
 
     def _search(self):
-        self.tasks.append(self.loop.create_task(self._search_task()))
+        self._search_task()
```
jackrobison, line 238: delete this method?

lbry/extras/daemon/daemon.py
```diff
@@ -4885,20 +4885,16 @@ async def jsonrpc_blob_delete(self, blob_hash):
     """
 
     @requires(DHT_COMPONENT)
-    async def jsonrpc_peer_list(self, blob_hash, search_bottom_out_limit=None, page=None, page_size=None):
```
jackrobison, line 4888: this should be documented as a breaking change

PR #3563, opened 2022-02-09
DHT: make hash announcer use consumers instead of gathering batches

shyba:
Makes the hash announcer consume from a queue instead of just gathering batches (we were pausing for each gather, instead this gather consumers for a queue)


shyba:
since it depends on `dht_bug`, I changed the base there and will merge to unify the PRs

PR #3567, opened 2022-02-09
INSTALL.md: Add more dependencies

orblivion:
I needed these on Debian Buster (on QubesOS). I thought I could speed along the next person here, assuming these are correct in the general case.

In case it matters: I did not use deadsnakes because Buster still has Python 3.7.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -44,7 +44,7 @@ On Ubuntu (we recommend 18.04 or 20.04), install the following:
 ```
 sudo add-apt-repository ppa:deadsnakes/ppa
 sudo apt-get update
-sudo apt-get install build-essential python3.7 python3.7-dev git python3.7-venv libssl-dev python-protobuf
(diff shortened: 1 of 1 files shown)


orblivion:
(This was part of the previous protobuf-related PR I otherwise threw out, but I thought I might as well submit this)

eukreign:
A while ago we decided to only officially support Ubuntu for development (after getting a bunch of PRs supporting different Linux variants) as it would be impractical to support every variant. Also, we make the assumption that basic Python development tools are already installed.

Finally, `leveldb` is being deprecated so we should not need to compile it in the future.

orblivion:
Hmm, okay. I certainly wouldn't want to pollute this with extra dependencies. When I made this I think I was just guessing that the req's would be similar between Debian and Ubuntu. But perhaps I'll try again on my Ubuntu machine at some point and report back.

I'm not sure about the assumption that these tools would already be installed. If my experience on Ubuntu would be the same as what I had on Debian, I'd save time having these things listed explicitly. (Also I'm not sure if you mean that `autoconf`, `autogen` etc are Python dev tools.)

PR #3568, opened 2022-02-09
Comment about dead code in WalletStore.upgrade()

orblivion:
The other thing I wanted to mention from my thrown-out protobuf PR - I'm not sure what `WalletStore.upgrade()` was intended to do, but it looks like there's a bug. I'm not necessarily proposing this specific change, but here I'm just deleting what looks like dead code.

### DIFF
--- a/lbry/wallet/wallet.py
+++ b/lbry/wallet/wallet.py
@@ -268,6 +268,10 @@ class WalletStorage:
             return self._default.copy()
 
     def upgrade(self, json_dict):
+        # NOTE - The purpose of this method is to act as a scaffolding for
+        # someone to implement actual upgrade when it becomes necessary. As
+        # currently written, most of the code has no effect.
+
         json_dict = json_dict.copy()
         version = json_dict.pop('version', -1)
         if version == -1:


lbry/wallet/wallet.py
```diff
@@ -269,11 +269,7 @@ def read(self):
 
     def upgrade(self, json_dict):
         json_dict = json_dict.copy()
-        version = json_dict.pop('version', -1)
-        if version == -1:
-            pass
```
orblivion, line 274: The value of `version` has no consequence.

lbry/wallet/wallet.py
```diff
@@ -269,11 +269,7 @@ def read(self):
...
-        upgraded = self._default.copy()
-        upgraded.update(json_dict)
```
orblivion, line 276: This changes `upgraded` and `upgraded` is discarded. Perhaps you meant to return it instead of `json_dict`.

kauffj:
FYI @eukreign is off this week (and last week), so this will probably sit for a bit

eukreign:
@orblivion The purpose of the code in the upgrade function is to act as scaffolding for someone to implement actual upgrade when it becomes necessary. Generally I would agree with you that having dead code isn't great but I would like to make an exception in this particular case because it's just a few lines and makes it obvious where the upgrade code should go.

Does that explanation make sense? Can I close this PR?

orblivion:
Oh, okay. What if I just add a comment about this?

PR #3571, opened 2022-02-11
Seed mode: check that the candidate blob is at least 1 prefix byte close to peer id

shyba:


lbry/extras/daemon/components.py
```diff
@@ -413,7 +413,8 @@ async def download_blobs_in_background(self):
             self.space_available = await self.space_manager.get_free_space_mb(True)
             if not self.is_busy and self.space_available > 10:
                 blob_hash = next((key.hex() for key in self.dht_node.stored_blob_hashes if
-                                 key.hex() not in self.blob_manager.completed_blob_hashes), None)
+                                 key[0] == self.dht_node.protocol.node_id[0]
```
jackrobison, line 416: I think there should be a constant to determine how many bits are matching

shyba: extracted to `MIN_PREFIX_COLLIDING_BITS`

PR #3572, opened 2022-02-16
Add wallet json-schema, validate in one test.

orblivion:
A replacement for the would-be protobuf pull request. Depends on this PR: https://github.com/lbryio/types/pull/48/files

The idea is to add the wallet file format to the list of our known formats. This PR will confirm that `Wallet.to_dict` generates a wallet that conforms to the spec. This way we'll at least have some assurance that the spec is up to date.

I sort of made a guess as to what directory wallet.json would be best to go into.

BTW I added this to `Makefile` to try to match the `protoc` stuff, but running `make` doesn't seem to generate all of the python files that are there. There ends up being files removed from the repository.

### DIFF
--- a/lbry/schema/Makefile
+++ b/lbry/schema/Makefile
@@ -2,4 +2,5 @@ build:
 	rm types/v2/* -rf
 	touch types/v2/__init__.py
 	cd types/v2/ && protoc --python_out=. -I ../../../../../types/v2/proto/ ../../../../../types/v2/proto/*.proto
+	cd types/v2/ && cp ../../../../../types/jsonschema/* ./
 	sed -e 's/^import\ \(.*\)_pb2\ /from . import\ \1_pb2\ /g' -i types/v2/*.py
--- a/tests/unit/wallet/test_wallet.py
+++ b/tests/unit/wallet/test_wallet.py
@@ -1,10 +1,10 @@
 import json
 import jsonschema
 import os
-import pathlib
 import tempfile
 from binascii import hexlify
 
+import lbry.schema.types.v2 as schema_v2
 from unittest import TestCase, mock
 from lbry.testcase import AsyncioTestCase
 from lbry.wallet import (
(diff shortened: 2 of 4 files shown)


tests/unit/wallet/test_wallet.py
```diff
@@ -17,6 +20,11 @@ async def asyncSetUp(self):
         self.main_ledger = self.manager.get_or_create_ledger(Ledger.get_id(), config)
         self.test_ledger = self.manager.get_or_create_ledger(RegTestLedger.get_id(), config)
 
+        base_dir = os.path.join(os.path.dirname(__file__), '../../../')
+        types_dir = os.path.join(base_dir, 'lbry/schema/types/v2')
```
orblivion, line 24: I know this is ugly but I couldn't find an example of doing anything similar in the codebase.

tests/unit/wallet/test_wallet.py
```diff
@@ -74,6 +82,62 @@ def test_load_and_save_wallet(self):
...
+                        'name': 'single-address',
+                    }
+                },
+            ]
+        }
+
+        storage = WalletStorage(default=wallet_dict)
+        wallet = Wallet.from_storage(storage, self.manager)
+        self.assertDictEqual(wallet_dict, wallet.to_dict())
+        jsonschema.validate(schema=self.wallet_schema, instance=wallet.to_dict())
```
eukreign, line 139: Could this just be added to one of the other tests which save and load the wallet, seems like it fits in well with the concept of testing saving/loading. I think the test right above this one would work?

If I missed something and this is doing something extra than what's in the existing saving/loading tests let me know.

orblivion: Yeah I actually started off adding this to the save/load test. I think the only thing that this is adding is testing both single and deterministic addresses (that's why there are two accounts). I figured it'd be simpler to have it be a separate test case rather than maintain a single test case with both purposes, but it's not a strong preference. Could I expand the test data for save/load test to include a second account?

lbry/schema/Makefile
```diff
@@ -2,4 +2,5 @@ build:
 	rm types/v2/* -rf
 	touch types/v2/__init__.py
 	cd types/v2/ && protoc --python_out=. -I ../../../../../types/v2/proto/ ../../../../../types/v2/proto/*.proto
+	cd types/v2/ && cp ../../../../../types/jsonschema/* ./
```
orblivion, line 5: Updated per change in types repo.

eukreign (approved):

PR #3586, opened 2022-03-22
fix #3530 added error log when tcp port is already in use

### LINKED ISSUE #3530: SDK fails silently if P2P TCP port is in use
SDK silently fails to start the peer protocol server, making the whole startup fail without a clear reason.

This issue is about making that clear somehow in the logs and status.

We should consider picking a random port if no port was set by the user and 4444 is in use by something else (maybe another issue, may need discussion and input @jackrobison @lyoshenka ).

AlessandroSpallina:
This commit adds a log error when the TCP port 4444 is already in use in the system.
Refs to #3530

When the port is in use the log would be something like the following

```
2022-03-22 17:15:43,376 ERROR    lbry.blob_exchange.server:176: Failed to bind TCP 0.0.0.0:4444
```


shyba (approved):

shyba:
Thank you!

PR #3599, opened 2022-04-12
Tighten up IterativeFinder async close behavior (DHT iterator continues after consumer breaks out of it)

moodyjon:
See: https://github.com/lbryio/lbry-sdk/issues/3504

Script:
[scratch.py.txt](https://github.com/lbryio/lbry-sdk/files/8476728/scratch.py.txt)

Script output:
[scratch_1.txt](https://github.com/lbryio/lbry-sdk/files/8476723/scratch_1.txt)

Initially there is a lot of background activity from IterativeNodeFinder during what I presume is startup.
In steady state there is little activity after each "PAUSE---". Usually just stuff related to cancellation, if anything.

Eventually, I hope the "async with aclosing(...)" feature can be used:
https://docs.python.org/3.10/library/contextlib.html?highlight=aclosing#contextlib.aclosing

### DIFF
--- a/lbry/dht/protocol/iterative_find.py
+++ b/lbry/dht/protocol/iterative_find.py
@@ -276,8 +276,8 @@ class IterativeFinder(AsyncGenerator):
     async def __anext__(self) -> typing.List['KademliaPeer']:
         return await super().__anext__()
 
-    async def asend(self, val):
-        return await self.generator.asend(val)
+    async def asend(self, value):
+        return await self.generator.asend(value)
 
     async def athrow(self, typ, val=None, tb=None):
         return await self.generator.athrow(typ, val, tb)
--- a/lbry/dht/protocol/iterative_find.py
+++ b/lbry/dht/protocol/iterative_find.py
@@ -1,6 +1,7 @@
 import asyncio
 from itertools import chain
 from collections import defaultdict, OrderedDict
+from collections.abc import AsyncGenerator
 import typing
 import logging
 from typing import TYPE_CHECKING
@@ -71,7 +72,7 @@ def get_shortlist(routing_table: 'TreeRoutingTable', key: bytes,
     return shortlist or routing_table.find_close_peers(key)
 
 
-class IterativeFinder:
+class IterativeFinder(AsyncGenerator):
     def __init__(self, loop: asyncio.AbstractEventLoop, peer_manager: 'PeerManager',
                  routing_table: 'TreeRoutingTable', protocol: 'KademliaProtocol', key: bytes,
                  max_results: typing.Optional[int] = constants.K,
@@ -98,6 +99,8 @@ class IterativeFinder:
         self.iteration_count = 0
         self.running = False
         self.tasks: typing.List[asyncio.Task] = []
+        self.generator = None
+
         for peer in get_shortlist(routing_table, key, shortlist):
             if peer.node_id:
                 self._add_active(peer, force=True)
@@ -163,12 +166,16 @@ class IterativeFinder:
         except asyncio.TimeoutError:
             self._reset_closest(peer)
             return
+        except asyncio.CancelledError:
+            log.debug("%s[%x] cancelled probe",
+                      type(self).__name__, id(self))
+            return
         except ValueError as err:
             log.warning(str(err))
             self._reset_closest(peer)
             return
         except TransportNotConnected:
-            return self.aclose()
+            return self._aclose()
         except RemoteException:
             self._reset_closest(peer)
             return
@@ -182,12 +189,16 @@ class IterativeFinder:
         added = 0
         for index, peer in enumerate(self.active.keys()):
             if index == 0:
-                log.debug("closest to probe: %s", peer.node_id.hex()[:8])
+                log.debug("%s[%x] closest to probe: %s",
+                          type(self).__name__, id(self),
+                          peer.node_id.hex()[:8])
             if peer in self.contacted:
                 continue
             if len(self.running_probes) >= constants.ALPHA:
                 break
             if index > (constants.K + len(self.running_probes)):
+                break
+            if self.iteration_count + self.iteration_queue.qsize() >= self.max_results:
                 break
             origin_address = (peer.address, peer.udp_port)
             if origin_address in self.exclude:
@@ -198,9 +209,13 @@ class IterativeFinder:
                 continue
             self._schedule_probe(peer)
             added += 1
-        log.debug("running %d probes for key %s", len(self.running_probes), self.key.hex()[:8])
+        log.debug("%s[%x] running %d probes for key %s",
+                  type(self).__name__, id(self),
+                  len(self.running_probes), self.key.hex()[:8])
         if not added and not self.running_probes:
-            log.debug("search for %s exhausted", self.key.hex()[:8])
+            log.debug("%s[%x] search for %s exhausted",
+                      type(self).__name__, id(self),
+                      self.key.hex()[:8])
             self.search_exhausted()
 
     def _schedule_probe(self, peer: 'KademliaPeer'):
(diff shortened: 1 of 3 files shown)


shyba:
Hello, thank you for looking into it!

Since [aclosing](https://docs.python.org/3.10/library/contextlib.html?highlight=aclosing#contextlib.aclosing) is introduced in python 3.10, I think it would be better to add the snippet that they show as equivalent to [utils.py](https://github.com/lbryio/lbry-sdk/blob/master/lbry/utils.py):
```
from contextlib import asynccontextmanager

@asynccontextmanager
async def aclosing(thing):
    try:
        yield thing
    finally:
        await thing.aclose()
```

Then, we can use it from utils until we support 3.10, changing the import to the built-in one.

I modified the snippet to test that:
```
import asyncio
import inspect
from contextlib import asynccontextmanager

@asynccontextmanager
async def aclosing(thing):
    try:
        yield thing
    finally:
        close = thing.aclose()
        if inspect.isawaitable(close):
            await close

from lbry.dht.constants import generate_id
from lbry.dht.node import Node
from lbry.dht.peer import PeerManager
from lbry.extras.daemon.storage import SQLiteStorage
from lbry.conf import Config
import logging
logging.basicConfig(level=logging.DEBUG, format="%(asctime)s %(levelname)-4s %(name)s:%(lineno)d: %(message)s")


async def do_it():
    loop = asyncio.get_event_loop()
    #loop.set_debug(True)
    conf = Config()
    print(str(conf))
    storage = SQLiteStorage(conf, ":memory:", loop, loop.time)
    await storage.open()
    node = Node(
        loop, PeerManager(loop), generate_id(), 4444, 4444, 4444, None,
        storage=storage
    )
    node.start(interface='0.0.0.0', known_node_urls=conf.known_dht_nodes)
    await asyncio.sleep(10)
    key = bytes.fromhex('2ecce30223bc2808bd10cf8e18f1cccecedf9fd35fcc370b9d15ad7473a59d3dc16e85a8f08a1891e668608d1660b8c7')
    while True:
        async with aclosing(node.get_iterative_value_finder(key=key, max_results=1)) as values:
            i = 0
            print('SEARCH---')
            async for _ in values:
                print(f'{i}: {_}')
                i += 1
                break
        print('PAUSE---')
        await asyncio.sleep(10)
    await asyncio.wait(asyncio.all_tasks())

if __name__ == '__main__':
    asyncio.run(do_it())
```

Doing it this way, the pause and search messages clean, like
```
PAUSE---
SEARCH---
```

This looks cleaner than implementing the finally->aclose and other changes on every call and we would be able to use the built in version when it is out.

moodyjon:
Implemented an aclosing() in lbry/utils/py as suggested.

lbry/dht/protocol/iterative_find.py
```diff
@@ -163,12 +166,17 @@ async def _send_probe(self, peer: 'KademliaPeer'):
         except asyncio.TimeoutError:
             self._reset_closest(peer)
             return
+        except asyncio.CancelledError:
+            log.debug("%s[%x] cancelled probe",
+                      type(self).__name__, id(self))
+            return
```
shyba, line 172: Can you please re-raise? Catching CancelledError can prevent it from propagating and properly cancelling other parts. [In almost all situations it must be re-raised.](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError).

lbry/dht/protocol/iterative_find.py
```diff
@@ -182,13 +190,17 @@ def _search_round(self):
...
+                log.debug("%s[%x] closest to probe: %s",
+                          type(self).__name__, id(self),
+                          peer.node_id.hex()[:8])
             if peer in self.contacted:
                 continue
             if len(self.running_probes) >= constants.ALPHA:
                 break
             if index > (constants.K + len(self.running_probes)):
                 break
+            if self.iteration_count + self.iteration_queue.qsize() >= self.max_results:
```
shyba, line 202: Is this related to the current bug? If not, it would be cool to separate in another PR

lbry/dht/protocol/iterative_find.py
```diff
@@ -217,38 +233,74 @@ def callback(_):
...
 
     def _log_state(self):
-        log.debug("[%s] check result: %i active nodes %i contacted",
-                  self.key.hex()[:8], len(self.active), len(self.contacted))
+        log.debug("%s[%x] [%s] check result: %i active nodes %i contacted %i produced %i queued",
+                  type(self).__name__, id(self), self.key.hex()[:8],
+                  len(self.active), len(self.contacted),
+                  self.iteration_count, self.iteration_queue.qsize())
+
+    async def _generator_func(self):
```
shyba, line 241: Is it necessary to separate the generator when using the `aclosing` for fixing this bug? I tried on master and it worked without doing it.
I see there are some nice changes in behavior which would be cool to have in a new PR with the others. This helps testing and tidying the changelog.

lbry/dht/protocol/iterative_find.py
moodyjon: Will do.

lbry/dht/protocol/iterative_find.py
moodyjon: When I started looking at this, it seemed questionable that max_results was not considered before doing more work. These might be extra results which would not be consumed. But I see now the recent changes to iterative_find.py in Mar and Dec which removed max_results from consideration in some cases. And it's unclear when/where a caller might utilize max_results=N -- existing code defaults to K or K*2.

For this particular key (2ecce30223bc2808bd10...) it seems like search-exhausted comes from this bit of code even with large max_results.

`
         if index > (constants.K + len(self.running_probes)):
             break
`

Small max_results (<K) is ignored too (`self.max_results = max(constants.K, max_results)`)

Since I have little understanding of the algorithm and purpose of max_results, I will drop code limiting output.

moodyjon:
I think I have addressed the 3 comments. The IterativeFinder is mostly back to the original logic with no generator function stuff or output limiting based on max_results. Let me know if there are more comments or a request to squash changes.

shyba (approved):

moodyjon:
I've seen this failure in my runs as well. Seems like the upper bound needs a bump analogous to:

https://github.com/lbryio/lbry-sdk/commit/fd73412f12cc5c992c3f161e1fce721d856b4cf6

  File "/home/runner/work/lbry-sdk/lbry-sdk/tests/integration/other/test_exchange_rate_manager.py", line 30, in test_exchange_rate_manager
[123](https://github.com/lbryio/lbry-sdk/runs/6455212276?check_suite_focus=true#step:11:123)
    self.assertLessEqual(lbc, 20_000)
[124](https://github.com/lbryio/lbry-sdk/runs/6455212276?check_suite_focus=true#step:11:124)
AssertionError: Decimal('20004.00080016') not less than or equal to 20000
[125](https://github.com/lbryio/lbry-sdk/runs/6455212276?check_suite_focus=true#step:11:125)

moodyjon:
Argggggh :-/

test_exchange_rate_manager.py still failing, but due to server being down:

```
Hotbit is under maintenance...
...
Dear Respected Hotbit Users, Hotbit will be under maintenance from 1:30 AM to 1:30 PM UTC on 17th May to optimize the servers' performance. During the maintenance, the service will be unavailable. We apologize for any inconvenience caused!
````

eukreign:
@moodyjon @shyba what's the status of this PR?

shyba:
@moodyjon Can you please rebase on  master instead of merging? Using rebase you can also remove the merge commits, making it a cleaner history

moodyjon:
> @moodyjon @shyba what's the status of this PR? (it needs a rebase to get some test run fixes from master)

I had been observing/merging the test changes. Added one more test change yesterday: https://github.com/lbryio/lbry-sdk/pull/3599/commits/f91bf8f3e567ddcd029f6c9951c7a604dbf48ff7

> @moodyjon Can you please rebase on master instead of merging? Using rebase you can also remove the merge commits, making it a cleaner history

OK, I think I have rebased correctly.

moodyjon:
Removed exchange rate test bump of upper limit. It was changed in https://github.com/lbryio/lbry-sdk/commit/5c708e1c6f80e7f60d3e9065f26da2e6d330517b

shyba:
@moodyjon thank you very much!

PR #3637, opened 2022-07-22
new `wallet` commands for `import`ing and `export`ing wallets

jessopb:
@eukreign Recommended a wallet import / output method rather than altering the behavior of sync apply. This does not force reset the wallet encrypt password with the import/export data encrypt password.

I mostly duplilcated what I thought was the relevant sync_apply test. Let me know if you think any of the test is too redundant or missing.


lbry/extras/daemon/daemon.py
```diff
@@ -1382,6 +1382,69 @@ async def jsonrpc_wallet_create(
...
+        """
+        Export wallet data
+
+        Wallet must be unlocked to perform this operation.
+
+        Usage:
+            wallet_export [--password=<password>] [--wallet_id=<wallet_id>]
+
+        Options:
+            --password=<password>         : (str) password to decrypt incoming and encrypt outgoing data
```
eukreign, line 1396: probably removing the "incoming" part in the help string

eukreign (changes requested):
is this PR ready for one final review?

jessopb:
> Review?

Yes please

eukreign (approved):

PR #3646, opened 2022-08-08
--exit-on-disconnect flag to start lbrynet

eukreign:


moodyjon:
Can you please rebase or use some other technique to simplify the change? Things like https://github.com/lbryio/lbry-sdk/pull/3646/commits/54da1e3eaa7057433e75b51b6512a9a59719b396 and https://github.com/lbryio/lbry-sdk/pull/3646/commits/4f9c98e6b888921add54a0bbb8a45126d0977791 have already been submitted.

lbry/wallet/network.py
```diff
@@ -387,9 +398,16 @@ async def retriable_call(self, function, *args, **kwargs):
...
+                if self.exit_on_disconnect:
+                    log.error("Wallet server call timed out, exiting on server disconnect.")
+                    sys.exit(1)
+                else:
+                    log.warning("Wallet server call timed out, retrying.")
             except ConnectionError:
                 log.warning("connection error")
+                if self.exit_on_disconnect:
+                    log.error("exiting on server disconnect")
+                    sys.exit(1)
```
moodyjon, line 410: You should consider adding SystemExit exception handling here: 
https://github.com/lbryio/lbry-sdk/blob/a334a93757d8e5c8310f7ab0bf74562a6d361993/lbry/extras/cli.py#L281

I guess the risk is that it would hang while shutting down components, but calling daemon.stop() is done in other scenarios like SIGINT, SIGTERM.

moodyjon (commented):
General comment about `sleep_delay` and `_urgent_need_reconnect`:

```
    async def network_loop(self):
        sleep_delay = 30
        while self.running:
            await asyncio.wait(
                [asyncio.sleep(30), self._urgent_need_reconnect.wait()], return_when=asyncio.FIRST_COMPLETED   #### It seems like this should be sleep(sleep_delay), or min(sleep_delay, X) not a fixed 30s.
            )
            if self._urgent_need_reconnect.is_set():
                sleep_delay = 30
            self._urgent_need_reconnect.clear()   #### Should this be moved after "if not client:" so that the urgent reconnect will still be needed? When cleared here, the urgency goes away until another RPC is tried and finds the connection is not available.
            if not self.is_connected:
                client = await self.connect_to_fastest()
                if not client:
                    log.warning("failed to connect to any spv servers, retrying later")
                    sleep_delay *= 2
                    sleep_delay = min(sleep_delay, 300)
                    continue
                log.debug("get spv server features %s:%i", *client.server)
```

PR #3647, opened 2022-08-08
removed go hub dependency

eukreign:


moodyjon:
Fixes https://github.com/lbryio/lbry-sdk/issues/3643

Problematic dependency was deleted in these changes.

PR #3677, opened 2022-10-03
`claim_list` shows metadata of blocked claims

### LINKED ISSUE #3670: Better Support blocklist reposting
We propose the general principle that
`claim_list` should always be able to show you your claims and their references in entirety.
`resolve` if blocked returns the error, respecting the block list.
`claim_search` filters out blocked claims.

Currently, it seems `claim_list` with `--resolve` does not resolve reposted claims.
Currently, we cannot look up a particular blocked item or channel with a blocked item in the event of a dispute/unblock.

- [ ] `claim_list --resolve` resolves reposted claims (even if blocked) and their signing channels
- [ ] `claim_list --reposted_claim_id --resolve (implied?)` returns a repost of the claim_id
- [ ] `claim_list --reposted_channel_id` returns all reposts for claims in the channel.
- [ ] `claim_list` output for reposts has claim_id rather than claim_hash bytes.

eukreign:
fixes #3670

### DIFF
--- a/lbry/wallet/ledger.py
+++ b/lbry/wallet/ledger.py
@@ -993,7 +993,13 @@ class Ledger(metaclass=LedgerRegistry):
                 results.append(resolved)
             else:
                 if isinstance(resolved, dict) and 'error' in resolved:
-                    txo.meta['error'] = resolved['error']
+                    if resolved['error'].get('name') == 'BLOCKED' and txo.is_claim and txo.claim.is_repost:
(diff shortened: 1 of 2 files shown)


moodyjon:
Appears the plan is to make `claimtrie.getclaimbyid` enforce block lists (https://github.com/lbryio/hub/issues/108). If that's implemented, this will no longer work.

PR #3685, opened 2022-10-12
added `claim_list` filtering by `--reposted_claim_id` and fix for `claim_id` of reposted claim in JSON output

eukreign:


jessopb:
This works great.

PR #3687, opened 2022-10-13
Repost with deletions/edits v1

moodyjon:
Context: https://github.com/lbryio/types/issues/49

Definitely very rough, with only a few attributes tested. Just putting this out there as an object for discussion.

The issue I'm currently wrestling with is how to treat common fields like tags, languages, title, description. When an update() request to set (or clear) such a field comes from `jsonrpc_stream_update()`, do I put it in `BaseClaim`, or `ClaimReference`? I am trying to set these first in the `Repost(BaseClaim)`, then for the ones that fail, I try to put them in `ClaimReference` (either deletions or edits).

Maybe there should be separate args or a mode flag at the level of `jsonrpc_stream_update` that selects where to apply the modifications? Regular mode would modify the Repost. Embedded mode would modify the ClaimReference (deletions/edits).

I could also do something like split the updates into deletions and edits first. For deletions, try the BaseClaim (of the Repost) first, then fallback to ClaimReference. For edits, try the ClaimReference first, then fallback to BaseClaim (of the Repost).

Protobuf changes:
```
diff --git a/v2/proto/claim.proto b/v2/proto/claim.proto
index cccb960..7008578 100644
--- a/v2/proto/claim.proto
+++ b/v2/proto/claim.proto
@@ -42,6 +42,8 @@ message Channel {
 
 message ClaimReference {
     bytes claim_hash = 1;
+    Claim deletions = 2;
+    Claim edits = 3;
 }
```


Makefile
```diff
@@ -23,4 +23,4 @@ idea:
 	cp -r scripts/idea/* .idea
 
 elastic-docker:
-	docker run -d -v lbryhub:/usr/share/elasticsearch/data -p 9200:9200 -p 9300:9300 -e"ES_JAVA_OPTS=-Xms512m -Xmx512m" -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.12.1
+	docker run -d --env network.publish_host=127.0.0.1 -v lbryhub:/usr/share/elasticsearch/data -p 9200:9200 -p 9300:9300 -e"ES_JAVA_OPTS=-Xms512m -Xmx512m" -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.12.1
```
moodyjon, line 26: I seem to have to use this to get "make elastic-docker" to work. Is it just me? (MacOS aarch64)

lbry/schema/claim.py
```diff
@@ -398,6 +420,18 @@ class Repost(BaseClaim):
...
     claim_type = Claim.REPOST
 
+    def update(self, **kwargs) -> dict:
+        claim_type = kwargs.pop('claim_type', None)
+        # Update common fields within BaseClaim.
+        kwargs = super().update(strict_update=False, **kwargs)
+        if claim_type:
+            # Remaining updates go into deletes/edits of ClaimReference.
+            kwargs = self.reference.update(claim_type, **kwargs)
+        return kwargs
```
moodyjon, line 430: See PR description. This is where I have a choice of applying the update to BaseClaim or stuff it inside ClaimReference.

lyoshenka:
gonna hold off on reviewing this for now. we discussed how edit/deletions are a bigger project that we need to tackle now, and we'll think about them again once the extensions work is done and we see how it plays in the wild

moodyjon:
Closed in favor of #3706

PR #3697, opened 2022-10-24
Reposts with deletions/edits V2

moodyjon:
This based on a different claim.proto schema that limits modifications to only reposts of streams and only within a specific set of "Stream.Extension" fields. This is enforced at the level of the protobuf.

The Stream.Extension message contains a protobuf.Any. Right now I check the type name to make sure it's a type that 1) is known when the SDK is built, and 2) comes from a limited list of messages that are intended to be used as extensions. But I think it would be possible to put a claim ID as the "type_url_prefix" of protobuf.Any and download the "descriptor" needed to understand the type. For example, if the "@type" field of the Any message contains "<claim_id>/pb.Cad", the SDK can try to download the contents pointed to by <claim_id>.

To see the claim.proto changes go here:
https://github.com/moodyjon/types/commits/repost_modify

Hub changes that would index the extension fields are TODO.


lyoshenka:
leaving a note since im assigned as reviewer: we decided to go with a simple string->[]string mapping instead of protobuf.Any

lbry/extras/daemon/daemon.py
```diff
@@ -3352,8 +3355,29 @@ async def jsonrpc_stream_repost(
...
+                    obj = StreamExtension(None, AnyMessage())
+                    if isinstance(ext, StreamExtension):
+                        obj.from_value(ext)
+                    else:
+                        obj.from_value({schema: ext})
+                except StreamExtensionTypeUnresolved as e:
+                    descriptor_txo = await self.ledger.get_claim_by_claim_id(e.url_prefix, include_is_my_output=True)
+                    ext = descriptor_txo.claim.stream.extensions['descriptor']
+                    pb_serialized = ext.unpacked['descriptor']
+                    StreamExtension.DESCRIPTOR_POOL.AddSerializedFile(pb_serialized)
```
moodyjon, line 3372: Illustrative of how dynamic extension types would be handled. To be removed.

lbry/extras/daemon/daemon.py
```diff
@@ -3506,6 +3530,22 @@ async def jsonrpc_stream_create(
...
```
moodyjon, line 3547: Illustrative of how dynamic extension types would be handled. To be removed.

tests/integration/claims/test_claim_commands.py
```diff
@@ -1526,6 +1531,124 @@ async def test_repost(self):
...
+        ext0.Pack(m0, type_url_prefix='')
+        ext0 = StreamExtension('descriptor', ext0)
+        self.assertEqual("ext.Descriptor", ext0.message.TypeName())
+
+        # create descriptor claim
+        tx = await self.stream_create(
+            f'{StringMapMessage.DESCRIPTOR.name}', '1.0', channel_name='@goodies',
+            extensions={ext0.schema: ext0},
+        )
+        string_map_claim = self.get_claim_id(tx)
```
moodyjon, line 1556: Illustrative of submitting a Descriptor as claim to the blockchain. To be removed.

lbry/schema/attrs.py
```diff
@@ -569,3 +666,305 @@ def append(self, tag: str):
...
+    def __delitem__(self, key):
+        self.message.ClearField(key)
+
+    def __iter__(self):
+        present = filter(lambda f: self.message.HasField(f.name), self.message.ListFields())
+        return iter(map(lambda f: f.name, present))
+
+    def __len__(self):
+        present = filter(lambda f: self.message.HasField(f.name), self.message.ListFields())
+        return len(present)
```
moodyjon, line 913: First attempt at a generic extension wrapper to be used when extension type is dynamic, not known at compile time. Most likely broken, though I did see __getitem__() work. To be removed.

moodyjon:
State of this is I need to remove the dynamic extension type stuff, and add some code to catch `StreamExtensionTypeUnresolved` when an unknown extension is encountered. For example, a stream claim has some extensions only some of which are of known type. We would want to ignore the unknown ones.

moodyjon:
To see the claim.proto changes go here:
https://github.com/moodyjon/types/commits/repost_modify

The latest is I am storing extensions in a map:
```
    // Value should contain one of the extension types (e.g. StringMap).
    // Key provides a hint indicating what keys we should expect in the map.
    map<string, google.protobuf.Any> extensions = 14;
```

Also, the values in StringMap are lists of strings.
```
// StringMap is a generic kind of Stream.Extension (see claim.proto) which
// is useful for testing, but is not space-efficient and does not enforce a
// limited set of fields.
message StringMap {
    map<string, StringList> s = 2;
}

message StringList {
    repeated string vs = 1;
}
```

moodyjon:
Planning some further simplification after talking to @lyoshenka.

moodyjon:
Closed in favor of #3706

